PR #6, opened 2020-10-18
Initial work on PluginAPI

Punikekk:


src/main/java/eu/darkbot/api/entities/Entity.java
```diff
@@ -0,0 +1,58 @@
...
+     * @return {@link LocationInfo}
+     */
+    LocationInfo getLocationInfo();
+
+    /**
+     * Checks that entity have effect
+     * @param effect to check
+     * @return true if current entity have indicated effect
+     */
+    boolean hasEffect(EntityEffect effect);
```
Pablete1234, line 33:
```suggestion
    default boolean hasEffect(EntityEffect effect) {
        return hasEffect(effect,getId());
    }
```
it'd be simpler if it's defined here as default, can apply to a few things

Punikekk:
<:PeepoSlippy:642073706985816087>

src/main/java/eu/darkbot/api/managers/HeroManager.java
```diff
@@ -1,9 +1,76 @@
...
+    default boolean setMode(ShipConfig shipConfig) {
+        return setMode(shipConfig.CONFIG, shipConfig.FORMATION);
+    }
+
+    boolean setAttackMode(Npc target);
+    boolean setAttackMode();
+    boolean setRoamMode();
+    boolean setRunMode();
+
+    //should be added into heroManager directly?
```
Pablete1234, line 51: I'd move it away into a separate interface

src/main/java/eu/darkbot/api/managers/HeroManager.java
```diff
@@ -15,20 +15,24 @@
...
+        return timeTo(getLocationInfo().distance(destination));
+    }
+
+    default long timeTo(LocationInfo destination) {
+        return timeTo(getLocationInfo().distance(destination));
+    }
+
+    default long timeTo(Entity destination) {
+        return timeTo(getLocationInfo().distance(destination));
     }
```
Pablete1234: Consider adding a generic Locatable interface with a "getLocation" method, Entity, LocationInfo and Location all implement it, and you stop having the need for 3 diff methods for this

Punikekk: <:thumbsup~1:670684720639770635>

src/main/java/eu/darkbot/api/core/Memory.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    /**
+     * Reads String from memory.
+     *
+     * @param address to read
+     */
+    String readString(long address);
+
+    default String readString(long address, int offsets) {
```
Pablete1234, line 81:
```suggestion
    default String readString(long address, int... offsets) {
```
That's supposed to be an array right?

src/main/java/eu/darkbot/api/entities/Entity.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    boolean hasEffect(int effect);
+
+    default boolean hasEffect(EntityEffect effect) {
+        return hasEffect(effect.getId());
+    }
+
+    /**
+     * Sets metadata key with given value and stores it only for current entity.
+     * Can be used for custom timers, checks, predicates etc.
```
Pablete1234, line 43:
```suggestion
     * Can be used for custom timers, checks, predicates etc. Make sure not to save
     * any custom class that wouldn't be able to be unloaded. Use java types.
```
I'd clarify this

src/main/java/eu/darkbot/api/entities/utils/Area.java
```diff
@@ -0,0 +1,8 @@
+package eu.darkbot.api.entities.utils;
+
+public class Area extends com.github.manolo8.darkbot.core.utils.pathfinder.Area {
```
Pablete1234, line 3: Keep Area as an interface, and `AreaImpl extends manolo8.Area implements Area`. In the future we will probably want circular areas or other kind of implementations that don't necessarily conform to the minxy maxxy rectangle model

src/main/java/eu/darkbot/api/objects/Locatable.java
```diff
@@ -0,0 +1,12 @@
...
+
+import eu.darkbot.utils.Location;
+
+public interface Locatable {
+
+    double getX();
+    double getY();
+
+    //maybe not necessary
+    Location getLocation();
```
Pablete1234, line 11: probs not necessary, and could even be implemented as a default method since we have x and y

Pablete1234 (commented):
Looking good, keep up the good work

src/main/java/eu/darkbot/api/entities/Pet.java
```diff
@@ -1,6 +1,74 @@
 package eu.darkbot.api.entities;
 
+import eu.darkbot.utils.ThrowableItem;
+
 public interface Pet extends Ship {
+
     int getLevel();
     int getOwnerId();
+
+    enum Buff {
```
Pablete1234, line 10:
```suggestion
    enum Cooldown {
```
While in-game they may be considered buffs, i think we should call them Cooldowns and remove the "_CD" suffix that a few of those have

src/main/java/eu/darkbot/utils/ThrowableItem.java
```diff
@@ -0,0 +1,5 @@
+package eu.darkbot.utils;
+
+public interface ThrowableItem {
```
Pablete1234, line 3: *Throwable* is a terrible word to use for anything that is not a type of exception or related. Maybe CastableItem? UsableItem? ConsumableItem ? or something like that

src/main/java/eu/darkbot/api/managers/StarManager.java
```diff
@@ -0,0 +1,32 @@
+package eu.darkbot.api.managers;
+
+import eu.darkbot.api.entities.utils.Map;
+
+import java.util.Set;
+
+public interface StarManager {
+    String[] HOME_MAPS = new String[]{"1-1", "2-1", "3-1"};
+    String[] OUTPOST_HOME_MAPS = new String[]{"1-8", "2-8", "3-8"};
```
Pablete1234, line 9: While final, this is mutable. For the future should probably look into immutable collections for exposed API fields

src/main/java/eu/darkbot/api/managers/StarManager.java
```diff
@@ -0,0 +1,32 @@
...
+        return getMaps().stream()
+                .filter(map -> map.getName().equals(mapName))
+                .findAny()
+                .orElseThrow(MapNotFoundException::new);
+    }
+
+    class MapNotFoundException extends Exception {
+        public MapNotFoundException() {
+        }
+    }
```
Pablete1234, line 31: Include a message that shows the map id or map name searched

src/main/java/eu/darkbot/api/entities/utils/Area.java
```diff
@@ -0,0 +1,15 @@
+package eu.darkbot.api.entities.utils;
+
+import java.awt.Shape;
+
+public interface Area extends Shape {//?
```
Pablete1234, line 5:
```suggestion
public interface Area {
```
Just area, the interface could have different implementations, some square, some not

src/main/java/eu/darkbot/api/entities/utils/Area.java
```diff
@@ -0,0 +1,15 @@
...
+
+    AreaType getAreaType();
+
+    enum AreaType {
+        RECTANGLE,
+        CIRCULAR,
+        TRIANGLE,
+        UNKNOWN
+    }
```
Pablete1234, line 14:
```suggestion
```
It's up to the implementator to decide this.

For Area just consider it a "place", it's shape isn't publicly known. 
Probably the most abstract methods (that will fill in the gaps required to be used properly) should be methods to see if:
 - It `contains` a point
 - If a line `intersects` this area (could default to intersecting with the bounds)
 - `getBounds` returning a rectangle that contains all inner points
 - Possibly a `getPoints()` that returns points arround this area, used for navigation.

src/main/java/eu/darkbot/utils/ArrayUtils.java
```diff
@@ -0,0 +1,18 @@
...
+public class ArrayUtils {
+
+    @SafeVarargs
+    public static <T> List<T> asImmutableList(T... a) {
+        return Collections.unmodifiableList(Arrays.asList(a));
+    }
+
+    public static boolean isEmpty(Collection<?> coll) {
+        return coll == null || coll.isEmpty();
+    }
```
Pablete1234, line 17: I am not convinced that this is a good idea, it can hide behaviour that may not be expected by whoever sees a call to "ArrayUtils.isEmpty" if they do not know what it is

src/main/java/eu/darkbot/utils/EquippableItem.java
```diff
@@ -0,0 +1,6 @@
+package eu.darkbot.utils;
+
+public interface EquippableItem {
+
+    String getExceptionMessage();
```
Pablete1234, line 5: Definetly not. An item does not have an "exception message", if anything give it a name or something more generic

src/main/java/eu/darkbot/utils/ItemNotEquippedException.java
```diff
@@ -0,0 +1,7 @@
+package eu.darkbot.utils;
+
+public class ItemNotEquippedException extends Exception {
+    public ItemNotEquippedException(EquippableItem item) {
+        super(item.getExceptionMessage());
```
Pablete1234, line 5:
```suggestion
        super("Item " + item.getName() + " was not equipped");
```
Something like that

src/main/java/eu/darkbot/utils/Point.java
```diff
@@ -0,0 +1,15 @@
+package eu.darkbot.utils;
+
+public class Point {
```
Pablete1234, line 3: I'd implement locatable here too, and just transform to double if treated as location

src/main/java/eu/darkbot/utils/Point.java
```diff
@@ -0,0 +1,28 @@
+package eu.darkbot.utils;
+
+import eu.darkbot.api.objects.Locatable;
+
+// in future Location extends Point?
```
Pablete1234, line 5: If location and point are both gonna be double, why would you have 2?

Point is integer based and location is double based, i think that's the intended behaviour

src/main/java/eu/darkbot/api/objects/Gui.java
```diff
@@ -0,0 +1,33 @@
...
+    void click(int plusX, int plusY);
+
+    /**
+     * Moves mouse at gui position + plusX/Y
+     */
+    void hover(int plusX, int plusY);
+
+    //should be available in API?
+    long getElementsList(int elementsListId);
+    //and so on
```
Pablete1234, line 32: definetly not, in fact the api should probably not assume "clicks" or anything in gui either, it should present higher-level apis, somewhat like pet that has setGear etc

src/main/java/eu/darkbot/api/managers/StatsManager.java
```diff
@@ -0,0 +1,22 @@
+package eu.darkbot.api.managers;
+
+public interface StatsManager {
+
+    int getLevel();
+    long getRuntime();
```
Pablete1234, line 6:
```suggestion
    long getRunningTime();
```
`Runtime` sounds like a piece of software, not like a time measurement

src/main/java/eu/darkbot/api/managers/PetManager.java
```diff
@@ -0,0 +1,63 @@
...
+
+    int getFuel();
+    int getMaxFuel();
+
+    int getHeat();
+    int getMaxHeat();
+
+    @Override
+    default int getHull() {
+        throw new UnsupportedOperationException("Pet dont have nano hull");
```
Pablete1234, line 41:
```suggestion
        throw new UnsupportedOperationException("Pet has no nano hull");
```
consider it may be simpler to return 0 than to throw an exception, simply pretending a pet *does* have a hull, it just turns out to be of always 0, max 0, and never increases nor decreases

src/main/java/eu/darkbot/api/objects/Circle.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    default boolean containsPoint(double x, double y) {
+        x -= getX();
+        y -= getY();
+        return Math.sqrt(x * x + y * y) <= getRadius();
```
Pablete1234, line 24: dis is slow, u can pre-calculate radius squared: `radiusSq = radius * radius;` and avoid the sqrt

src/main/java/eu/darkbot/api/objects/Gui.java
```diff
@@ -0,0 +1,29 @@
+package eu.darkbot.api.objects;
+
+public interface Gui extends Rectangle {
```
Pablete1234, line 3: i am not convinced that gui (point on screen) should mix with rectangle/shapes (position on the map)

src/main/java/eu/darkbot/api/objects/Gui.java
```diff
@@ -1,6 +1,9 @@
 package eu.darkbot.api.objects;
 
-public interface Gui extends Rectangle {
+public interface Gui extends Point {
+    double getWidth(); //or x2, y2?
+    double getHeight();
```
Pablete1234, line 5: I'd add both, default getX2() { return getX() + getWidth(); }

src/main/java/eu/darkbot/api/objects/Location.java
```diff
@@ -4,42 +4,49 @@
     /**
      * Returns distance to location ox & oy
      */
-    default double distanceTo(double ox, double oy) {
-        ox -= getX();
-        oy -= getY();
-        return Math.sqrt(ox * ox + oy * oy);
-    }
+    double distanceTo(double ox, double oy);
```
Pablete1234, line 7: i think this one can be defaulted, the distance between one location and another is a pretty universal concept as long as you provide x and y for both

src/main/java/eu/darkbot/api/objects/Location.java
```diff
@@ -4,42 +4,49 @@
...
         return distanceTo(other.getX(), other.getY());
     }
 
-    default double angleTo(double ox, double oy) {
-        return Math.atan2(getY() - oy, getX() - ox);
-    }
+    /**
+     * @return angle to other location as radians.
+     */
+    double angleTo(double ox, double oy);
```
Pablete1234: same here, this is a pretty generic operation it can stay as a default, it is not making any extra assumptions about implementation

src/main/java/eu/darkbot/api/objects/Location.java
```diff
@@ -4,42 +4,49 @@
...
 
+    /**
+     * Copies current location into a new {@link Location} object and returns it.
+     */
+    Location copy();
+    Location copy(double plusX, double plusY);
+
+    default Location copy(Location other) {
+        return copy(other.getX(), other.getY());
+    }
```
Pablete1234, line 30: i wouldn't mix, id leave it up for someone to just:
`loc.copy().plus(other)` instead of allowing for `loc.copy(other)`, becaue it is not clear what you're copying where, are you copying loc into other? are you setting loc to be a copy of other?

src/main/java/eu/darkbot/api/objects/Location.java
```diff
@@ -1,10 +1,16 @@
 package eu.darkbot.api.objects;
 
+import static java.lang.Math.pow;
+import static java.lang.Math.sqrt;
```
Pablete1234, line 4: avoid static imports

src/main/java/eu/darkbot/api/objects/Location.java
```diff
@@ -1,10 +1,16 @@
...
+
 public interface Location extends Locatable, Point {
     /**
-     * Returns distance to location ox & oy
+     * @return the distance from the current point {@code (getX(), getY())}
+     *         to the point {@code (ox, oy)}
      */
-    double distanceTo(double ox, double oy);
+    default double distanceTo(double ox, double oy) {
+        return sqrt(pow(getX() - ox, 2) + pow(getY() - oy, 2));
```
Pablete1234, line 12: you can avoid the call to pow for a power of 2, just multiply

src/main/java/eu/darkbot/api/managers/PetAPI.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * @return {@link LocationInfo} of pet locator target
+     * or else null if theres none.
+     */
+    @Nullable
+    LocationInfo getLocatorNpcLoc();
+
+    /**
+     * This method checks if pet have given cooldown.
+     * For example, kamikaze after use need to be cooldowned
```
Pablete1234, line 67:
```suggestion
     * Checks if pet has the given cooldown, which can make gears unavailable temporally.
```

src/main/java/eu/darkbot/api/entities/Ship.java
```diff
@@ -46,14 +46,21 @@
...
+     */
     String getClanTag();
 
     /**
-     * Returns ship faction as int.
-     * 0 = NONE, 1 = MMO, 2 = EIC, 3 = VRU, 4 = KRONOS?
+     * @return {@link Faction}
      */
-    int getFactionId();
+    Faction getFactionId();
```
Pablete1234, line 63: i'd just call it `getFaction`

src/main/java/eu/darkbot/api/entities/Npc.java
```diff
@@ -0,0 +1,14 @@
...
+
+    /**
+     * @return id of the npc
+     */
+    int getNpcId();
+
+    /**
+     * @return true if {@link} has {@code InstaShield} effect
+     */
+    boolean hasInstaShield();
```
Pablete1234, line 13: in the API should just use `hasEffect`, the ish flag was left for legacy reasons

src/main/java/eu/darkbot/api/entities/Portal.java
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Portal type id in generally is id of the portal's graphic.
+     *
+     * @return {@link Portal}'s type id
+     */
+    int getPortalType();
+
+    // TODO: 01.11.2020 not sure which and when.
+    int getFactionId();
```
Pablete1234, line 25: 5-1, 5-2 and 5-3 portals, are from a specific faction, and members from other factions don't see them on the map

src/main/java/eu/darkbot/api/managers/SlotBarAPI.java
```diff
@@ -0,0 +1,37 @@
...
+    @Nullable
+    SlotBar getPremiumBar();
+
+    /**
+     * @return pro action {@link SlotBar} otherwise null if doesn't exists.
+     */
+    @Nullable
+    SlotBar getProActionBar();
+
+    //maybe move search items here?
```
Pablete1234, line 36: probably good to flatten it out a bit, allow too many nested calls

src/main/java/eu/darkbot/api/core/Installable.java
```diff
@@ -0,0 +1,8 @@
+package eu.darkbot.api.core;
+
+import eu.darkbot.api.PluginAPI;
+
+public interface Installable {
+
+    void install(PluginAPI pluginAPI);
```
Pablete1234, line 7: missing uninstall

src/main/java/eu/darkbot/api/core/Tickable.java
```diff
@@ -0,0 +1,8 @@
+package eu.darkbot.api.core;
+
+public interface Tickable {
+
+    void onTick();
+
+    void onStopped();
+}
```
Pablete1234, line 8: I feel like tickable is useless and should be removed.

Module, Behaviour and Task should each implement their own (differently called) name for ticking, and only module and behaviour need an "onStopped" equivalent

src/main/java/eu/darkbot/api/managers/OreAPI.java
```diff
@@ -9,36 +9,40 @@
...
+         * @return true if ore trader window has been opened or closed and its animation is done,
+         * false if animation is not done, or no action is needed to be taken
+         * to change the visibility status of the ore trader window
+         */
+        boolean showTrade(boolean show, BasePoint base);
+    }
 
-    int getAmount(Ore ore);
+    interface Refinery extends Gui {
+        int getAmount(Ore ore);
```
Pablete1234, line 44: The fact that data of amount of ore comes from a refinery, and that refinery is a gui, is largely irrelevant, there's no need to have it on a sub-api

src/main/java/eu/darkbot/api/managers/KeyBindsAPI.java
```diff
@@ -0,0 +1,74 @@
...
+    @Nullable
+    Character getKeyBind(@NotNull Shortcut shortcut);
+
+    /**
+     * Search {@link Character} associated with given {@link Shortcut} and slotNumber in-game.
+     *
+     * @param shortcut   to be looked for
+     * @param slotNumber between 1 - 10
+     * @return {@link Character} associated with {@link Shortcut} otherwise null
+     * @throws IllegalStateException if {@code slotNumber < 1 || slotNumber > 10}
```
Pablete1234, line 22: IllegalArgumentException

src/main/java/com/github/manolo8/darkbot/core/objects/slotbars/SlotBar.java
```diff
@@ -28,23 +38,46 @@ public void update() {
...
+    public class Slot extends UpdatableAuto {
         public int slotNumber;
         public boolean premium; //not sure
         public String slotBarId;
         public @Nullable Item item;
 
+        /*public Item getItem() {
+            return item.address == 0 ? null : item;
+        }*/
+
```
Pablete1234, line 50: can't comment this out or plugins may break

src/main/java/eu/darkbot/api/managers/KeyBindsAPI.java
```diff
@@ -24,6 +26,18 @@
...
 
+    /**
+     * Search {@link Character} associated with given slotEntry
+     *
+     * @param slotEntry with {@link Shortcut} as key and {@link Integer} as value
+     * @return {@link Character} associated with given {@link Map.Entry} otherwise null
+     * @throws IllegalArgumentException if {@code slotEntry.value < 1 || slotEntry.value > 10}
+     */
+    @Nullable
+    default Character getKeyBind(@NotNull Map.Entry<SlotBarAPI.Type, Integer> slotEntry) throws IllegalArgumentException {
```
Pablete1234, line 37: this is a very weird way of passing parameters, i wouldn't make it part of the API

src/main/java/com/github/manolo8/darkbot/core/objects/slotbars/SlotBar.java
Punikekk: its new method

src/main/java/eu/darkbot/api/managers/AttackAPI.java
```diff
@@ -2,105 +2,87 @@
...
-        private final boolean cooldown;
+    boolean hasTarget();
 
-        LaserAmmo() {
-            this(false);
-        }
+    /**
+     * @return currently set target
+     */
+    Optional<Entity> getTarget();
```
Pablete1234, line 18: Entity shouldn't be the attack api target, either make an `Attackable` interface (and set it on ship, and whatever othere ntity that can be attacked) or just rely on `Ship`

src/main/java/eu/darkbot/api/managers/AttackAPI.java
```diff
@@ -2,105 +2,87 @@
...
-            return cooldown;
-        }
-    }
+    /**
+     * Tries to click selected target
+     *
+     * @return true if click was successful
+     * @see #getTarget()
+     */
+    boolean clickTarget();
```
Pablete1234, line 37: Like suggested before, don't name this "click", there may not be a mouse, use `target` or `attack`

src/main/java/eu/darkbot/api/managers/LogAPI.java
```diff
@@ -2,19 +2,31 @@
...
+     */
+    Listener<String> addListener(Listener<String> onMessage);
+
     /**
      * Gets the most recent log message
      *
      * @return the most recent message in your in-game log
      */
-    String getLastMessage();
+    Optional<String> getLastMessage();
```
Pablete1234: I think this method shouldn't even exist, it may confuse developers thinking they can stay up-to-date with log with this, but you'll get 4 log messages at once and miss 3 of them

src/main/java/eu/darkbot/api/managers/MovementAPI.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    void moveTo(double x, double y);
+
+    default void moveTo(@NotNull Locatable destination) {
+        moveTo(destination.getX(), destination.getY());
+    }
+
+    /**
+     * Returns closest distance to destination,
+     * calculates barriers and any other obstacles.
```
Pablete1234, line 76:
```suggestion
     * Returns distance to the closest valid (not inside any obstacle) location
```

src/main/java/eu/darkbot/api/managers/OreAPI.java
```diff
@@ -0,0 +1,69 @@
...
+         * @return true if ore trader window has been opened or closed and its animation is done,
+         * false if animation is not done, or no action is needed to be taken
+         * to change the visibility status of the ore trader window
+         */
+        boolean showTrade(boolean show, BasePoint base);
+    }
+
+    interface Refinery extends Gui {
+        int getAmount(Ore ore);
+    }
```
Pablete1234, line 45: The fact that data of amount of ore comes from a refinery, and that refinery is a gui, is largely irrelevant, there's no need to have it on a sub-api

src/main/java/eu/darkbot/api/managers/PetAPI.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Will change enabled flag for pet.
+     * If was set to {@code true} bot will try to enable the pet,
+     * only if user configured bot to use the pet.
+     */
+    void setEnabled(boolean enabled);
+
+    /**
+     * @return true if pet is active.
```
Pablete1234, line 30:
```suggestion
     * @return true if pet is active (alive and on the map).
```

src/main/java/eu/darkbot/api/managers/RepairAPI.java
```diff
@@ -0,0 +1,27 @@
...
+     *
+     * @throws IllegalStateException if ship is already repaired
+     */
+    void tryRevive(int repairOption) throws IllegalStateException;
+
+    Collection<Integer> getAvailableRepairOptions();
+
+    Optional<String> getLastDestroyerName();
+
+    Optional<LocalDateTime> getLastDeathTime();
```
Pablete1234, line 26: I feel like this may be a bit of an abuse of Optional, this is simple enough that we can just return nulls

Also, use Instant for the time datatype

src/main/java/eu/darkbot/api/managers/StarAPI.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * API to add, manage, retrieve maps.
+ */
+public interface StarAPI extends API {
+    List<String> HOME_MAPS = ArrayUtils.asImmutableList("1-1", "2-1", "3-1");
+    List<String> OUTPOST_HOME_MAPS = ArrayUtils.asImmutableList("1-8", "2-8", "3-8");
```
Pablete1234, line 18: Probably useful to add BLACKLIGHT_MAPS too nowadays since the're used, and maybe PIRATE_MAPS too just for competeness

src/main/java/eu/darkbot/api/managers/StarAPI.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return bounds of the current map
+     */
+    Area.Rectangle getCurrentMapBounds();
+
+    /**
+     * Adds given map into list of maps.
+     *
+     * @return true if was added, false when list already contains given {@link Map}
+     */
+    boolean addMap(@NotNull Map map);
```
Pablete1234, line 35: I'm not convinced that one should just be causally "adding" maps via API, if anything it should be a `getOrCreateMapById` where if a map doesn't exist it will just create a dummy map with that id and store it. It will also save you the allocaton of a new map that would be discarded if you can't add it

src/main/java/eu/darkbot/api/managers/WindowAPI.java
```diff
@@ -0,0 +1,96 @@
...
+    void keyClick(int keyCode);
+
+    default void keyClick(Character keyCode) {
+        if (keyCode != null) keyClick(keyCode.charValue());
+    }
+
+    /**
+     * Moves mouse to x & y coordinates of game window
+     */
+    void mouseMove(double x, double y);  //cast to int here or in implementation?
```
Pablete1234, line 64: probably the param here should be int, since i don't think you can move a mouse to a double coordinate on screen

src/main/java/eu/darkbot/api/objects/Gui.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Moves mouse at gui position + plusX/Y
+     */
+    void hover(int plusX, int plusY);
+
+    /**
+     * @return last {@link LocalDateTime} where gui was visible
+     */
+    LocalDateTime lastVisibleTime();
```
Pablete1234, line 60: use `Instant` not LocalDateTime

src/main/java/eu/darkbot/api/entities/Box.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @see Entity#isRemoved()
+     */
+    boolean isCollected();
+
+    /**
+     * @return time until box is marked as collected
+     */
+    @Nullable
+    LocalDateTime isCollectedUntil();
```
Pablete1234, line 33: Use `Instant` as the time datatype

src/main/java/eu/darkbot/api/utils/Listener.java
```diff
@@ -0,0 +1,13 @@
+package eu.darkbot.api.utils;
+
+import java.util.EventListener;
+
+/**
+ * Any {@link Listener} need to have strong reference
+ * otherwise will be garbage collected.
+ */
+@FunctionalInterface
+public interface Listener<T> extends EventListener {
```
Pablete1234, line 10: This should have no relation to java EventListener

Pablete1234 (changes requested):

src/main/java/eu/darkbot/api/entities/utils/Attackable.java
```diff
@@ -3,6 +3,80 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 public interface Attackable extends Entity, Health, Info {
+
+    /**
+     * @return the target of this {@link Attackable} if present else {@code null}
+     */
+    @Nullable
+    Entity getTarget();
```
Pablete1234, line 15: Unsure if this makes sense.
Imagine those mines in GOP, they probably should be Attackable (so that we can shoot them) but they're never gonna have a target or be aiming or shooting anyone.

Maybe need to distinguish between Attackable and Attacker (where they can attack you), ship implementing both

src/main/java/eu/darkbot/api/managers/OreAPI.java
```diff
@@ -10,7 +10,13 @@
  * API for ore related things such as selling ores or getting ore amount.
  */
 public interface OreAPI extends API {
-    interface Trade extends Gui {
+
+    Trade getTrade();
+
+    interface Trade extends Gui { //move upper?
+
+        int getAmount(Ore ore);
```
Pablete1234, line 18: this should definetly move to upper, getting amount of ore should be in OreAPI

src/main/java/eu/darkbot/api/PluginAPI.java
```diff
@@ -1,10 +1,17 @@
...
+import eu.darkbot.api.utils.Version;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 public interface PluginAPI {
 
+    /**
+     * {@link Version} of the DarkBot.
+     */
+    Version VERSION = Version.of("1.13.17 beta 59");
```
Pablete1234, line 13: probably doesn't belong here, it's not the version of the plugin api, it's the version of the bot

src/main/java/eu/darkbot/DarkBot.java
```diff
@@ -0,0 +1,56 @@
...
+    @Nullable
+    @SuppressWarnings("unchecked")
+    public <T extends eu.darkbot.api.API> T getAPI(@NotNull Class<T> api) {
+        return (T) apis.get(api);
+    }
+
+    @Override
+    @NotNull
+    public <T extends eu.darkbot.api.API> T requireAPI(@NotNull Class<T> api) throws UnsupportedOperationException {
+        T theApi = getAPI(api);
```
Pablete1234, line 34:
```suggestion
    public <T extends eu.darkbot.api.API> T requireAPI(@NotNull Class<T> clazz) throws UnsupportedOperationException {
        T api = getAPI(api);
```
calling something `theApi` looks awful

src/main/java/eu/darkbot/api/PluginAPI.java
```diff
@@ -23,6 +30,17 @@
      */
     @NotNull <T extends API> T requireAPI(@NotNull Class<T> api) throws UnsupportedOperationException;
 
+    /**
+     * @return current used {@link Module}
+     */
+    Module getModule();
```
Pablete1234, line 36: the plugin api doesn't hold the module, the "main" or the "bot" does, consider it an api that you can get from PluginAPI, like `pluginApi.requireAPI(DarkBot.class).getModule()` ?

src/main/java/eu/darkbot/api/plugin/Feature.java
```diff
@@ -0,0 +1,29 @@
...
+     */
+    String name();
+
+    /**
+     * Description of the {@link Feature}
+     */
+    String description();
+
+    /**
+     * Should feature be enabled by default?
```
Pablete1234, line 26: no, you'll enable a feature if you want it otherwise it should stay disabled. Modules are (or rather, have been) an exception since an enabled module does nothing unless you select it

src/main/java/eu/darkbot/logic/modules/MapModule.java
```diff
@@ -0,0 +1,138 @@
...
+                I18n.get("module.map_travel.status.no_next", traveler.target.getName());
+    }
+
+    @Override
+    public void onTickModule() {
+        if (!traveler.isDone()) traveler.tick();
+        if (traveler.isDone()) goBack();
+    }
+
+    private static class Traveler {
```
Pablete1234, line 48: Whis is this traveler embedded inside map module?
Sometimes other parts of the bot may want to travel without necessarily handing over the full control to the map module, the safety finder is an example of this, where it uses a traveler to go back to the previous map

src/main/java/eu/darkbot/logic/modules/MapModule.java
```diff
@@ -0,0 +1,138 @@
...
+            shipTpWait = mapChangeWait = -1;
+            this.target = target;
+            this.done = false;
+        }
+
+        public boolean isDone() {
+            return done;
+        }
+
+        private Location last;
```
Pablete1234, line 86: Having to save a "last" location when LocationInfo literally has that data for you is kinda redundant.
It'd be nice to get an api method for `hero.getLocationInfo().getLast()`

getLocationInfo returns a LocationInfo, that implements locatable as well as having a getLast that returns another locatable

src/main/java/eu/darkbot/api/managers/HeroAPI.java
```diff
@@ -20,6 +20,8 @@
 
     void sendRocketAttack();
 
+    boolean isLogout();
```
Pablete1234, line 23: Kind of a weird name, sounds like an action not a name

src/main/java/com/github/manolo8/darkbot/backpage/BackpageManager.java
```diff
@@ -176,19 +172,19 @@ public String getDataInventory(String params) {
...
-                    .map(m -> m.group(1))
-                    .findFirst().orElse(null);
-
-        } catch (IOException e){
+        try {
+            return getReloadToken(IOUtils.read(input));
+        } catch (IOException e) {
             e.printStackTrace();
             return null;
         }
```
Pablete1234, line 180: This is replacing a buffered reader that will read lines one by one and short-circuit when one matches, with a full download of the whole page to one massive string, and then doing a regex search over that huge string.

Keep it as it was

src/main/java/eu/darkbot/DarkBot.java
```diff
@@ -0,0 +1,9 @@
+package eu.darkbot;
+
+//main class.
+public class DarkBot {
+
+    public static void main(String[] args) {
+        //...
```
Pablete1234, line 7: :thinking:

src/main/java/eu/darkbot/api/PluginAPI.java
```diff
@@ -0,0 +1,55 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+
+public interface PluginAPI extends API {
+
+    /**
+     * @param api to get
+     * @param <T> type of api which extends {@link API}
+     * @return instance of given {@link API} type
```
Pablete1234, line 15:
```suggestion
     * @return instance of given {@link API} type, null if not present
```

src/main/java/eu/darkbot/api/PluginAPI.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    @NotNull <T> T getFeature(String featureId) throws ClassNotFoundException;
+    @NotNull <T> T getFeature(Class<T> feature) throws ClassNotFoundException;
+
+    /**
+     * @param featureId of the feature
+     * @return {@link FeatureInfo} of given feature.
+     * @throws ClassNotFoundException when given feature wasn't found
+     */
+    @NotNull <T> FeatureInfo<T> getFeatureInfo(String featureId) throws ClassNotFoundException;
```
Pablete1234, line 53: you can likely remove the string based methods and just keep the class ones (same for the one in getFeature)

src/main/java/eu/darkbot/api/entities/Box.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    String getHash();
+
+    /**
+     * @return type of the box.
+     */
+    String getTypeName();
+
+    /**
+     * @return {@link BoxInfo} with some settings for {@link Box}
```
Pablete1234, line 24:
```suggestion
     * @return {@link BoxInfo} with some user defined settings for {@link Box}
```

src/main/java/eu/darkbot/api/entities/Entity.java
```diff
@@ -0,0 +1,88 @@
...
+     * @return true if and only if {@link Entity} is loaded in-game and isn't removed
+     * @see #isRemoved()
+     */
+    boolean isValid();
+
+    /**
+     * Checks that entity is removed.
+     * If (isRemoved() == true) then this entity wont be updated anymore.
+     */
+    boolean isRemoved();
```
Pablete1234, line 31: Are these 2 methods complementary? is one the other negated?

If they are the same negated, remove one of the 2 or make it very clear that they are the same
If they are slightly different that one negated of the other, explain the difference in the doc

src/main/java/eu/darkbot/api/entities/Portal.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Portal type id in generally is id of the portal's graphic.
+     *
+     * @return {@link Portal}'s {@link PortalType}
+     */
+    @Nullable
+    PortalType getPortalType();
+
+    // TODO: 01.11.2020 not sure which and when.
```
Pablete1234, line 28: This faction id is for portals that belong to a specific faction, used in 5-3 exit portals to exit thru your factions' portal

src/main/java/eu/darkbot/api/entities/other/Effect.java
```diff
@@ -0,0 +1,31 @@
+package eu.darkbot.api.entities.other;
+
+/**
+ * Represents in-game {@link eu.darkbot.api.entities.Entity}'s effects.
+ */
+// TODO: 01.11.2020 add more effects
```
Pablete1234, line 6: what's missing?

src/main/java/eu/darkbot/api/managers/BotAPI.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Bot's management API
+ */
+public interface BotAPI extends API {
+
+    /**
+     * {@link Version} of the DarkBot.
+     */
+    Version VERSION = Version.of("1.13.17 beta 59");
```
Pablete1234, line 17: probably needs some updating

src/main/java/eu/darkbot/api/objects/Locatable.java
```diff
@@ -0,0 +1,29 @@
...
+     * Creates new instance of {@link Locatable} with given parameters.
+     * @param x coordinate
+     * @param y coordinate
+     * @return {@link Locatable} with given coordinates
+     */
+    static Locatable of(double x, double y) {
+        return new Locatable() {
+            public double getX() { return x; }
+            public double getY() { return y; }
+        };
```
Pablete1234, line 17: Creating an annonymous class is undesirable, if needed you can just create an inner `LocatableImpl` class that implements it

src/main/java/eu/darkbot/api/objects/Point.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public interface Point {
+    /**
+     * Creates new instance of {@link Point} with given parameters.
+     * @param x coordinate
+     * @param y coordinate
+     * @return {@link Point} with given coordinates
+     */
+    static Point of(double x, double y) {
+        return new Point() {
```
Pablete1234, line 14: same as locatable

Pablete1234 (changes requested):
I see a few todos left over and a couple things i've suggested, but it's advancing pretty well

src/main/java/eu/darkbot/DarkBot.java
Punikekk: as main class in da future instead of Bot.java

src/main/java/eu/darkbot/api/entities/other/Effect.java
Punikekk: over 100 effects ids <:KEKSAD:695390101018247198>

Pablete1234 (changes requested):

PR #8, opened 2020-10-21
auto login and start on startup

kekawce:


src/main/java/com/github/manolo8/darkbot/gui/login/SavedLogins.java
```diff
@@ -14,17 +15,40 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
 
 public class SavedLogins extends JPanel implements LoginScreen {
     public LoginForm loginForm;
 
     private char[] password;
     private boolean loaded = false;
-    private final Credentials credentials = LoginUtils.loadCredentials();
+    private static final Credentials credentials = LoginUtils.loadCredentials();
```
Pablete1234, line 26: This is a no-go

Pablete1234:
This should be implemented as a startup parameter passed into the bot, not a config

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -18,6 +19,8 @@ public static void main(String[] args) {
             UIManager.setLookAndFeel(new FlatDarkLaf());
             UIManager.put("Button.arc", 0);
             UIManager.put("Component.arc", 0);
+
+            CommandLineUtils.parse(args);
```
Pablete1234, line 23: What does look & feel have to do with this line? Should be either before or after this try block, but not inside it

src/main/java/com/github/manolo8/darkbot/utils/CommandLineUtils.java
```diff
@@ -0,0 +1,86 @@
...
+
+import com.github.manolo8.darkbot.gui.utils.Popups;
+
+import javax.swing.*;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+
+public class CommandLineUtils {
```
Pablete1234, line 11: This whole class is treated as static, it shouldn't. Consider naming it `StartupParams` or something like that, hold the specific info it needs, and pass it to main as a parameter. In general, avoid static when posslbe

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -21,6 +24,7 @@ public static void main(String[] args) {
         } catch (UnsupportedLookAndFeelException e) {
             e.printStackTrace();
         }
+        STARTUP_PARAMS.parse(args);
```
Pablete1234, line 27: Startup params params = new StartupParams(args);
```suggestion
        StartupParams params = new StartupParams(args);
```

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -21,6 +24,7 @@ public static void main(String[] args) {
...
         SwingUtilities.invokeLater(Main::new);
```
Pablete1234, line 28:
```suggestion
        SwingUtilities.invokeLater(() -> new Main(params));
```

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -2,12 +2,15 @@
...
 import com.github.manolo8.darkbot.config.ConfigManager;
+import com.github.manolo8.darkbot.utils.StartupParams;
 import com.github.manolo8.darkbot.utils.LogUtils;
 
 import javax.swing.*;
 
 public class Bot {
 
+    private static final StartupParams STARTUP_PARAMS = new StartupParams();
+
```
Pablete1234, line 13:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -30,4 +34,8 @@ public static void main(String[] args) {
             Bot.main(args);
         }
     }
+
+    public static StartupParams getStartupParams() {
+        return STARTUP_PARAMS;
+    }
```
Pablete1234, line 40:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -103,9 +112,14 @@ public Main() {
             Popups.showMessageAsync("Error", I18n.get("bot.issue.config_load_failed"), JOptionPane.ERROR_MESSAGE);
 
         API.createWindow();
+        initializeStartupSettings();
```
Pablete1234, line 115:
```suggestion
        initializeStartupSettings(params);
```

src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -103,9 +112,14 @@ public Main() {
...
         start();
     }
 
+    private void initializeStartupSettings() {
+        if (Bot.getStartupParams().AUTO_START) setRunning(true);
+    }
```
Pablete1234, line 121:
```suggestion
    private void initializeStartupSettings(StartupParams params) {
        if (params.AUTO_START) setRunning(true);
    }
```

src/main/java/com/github/manolo8/darkbot/gui/login/LoginForm.java
```diff
@@ -32,6 +33,8 @@ public LoginForm() {
         add(tabbedPane, "span 2, growx, height 112px!");
         add(infoLb, "gapleft 8px, grow 0");
         add(loginBtn, "gapright 8px");
+
+        if (Bot.getStartupParams().AUTO_LOGIN) startLogin();
```
Pablete1234, line 37: login form as a whole should be skipped if autologin is a thing

src/main/java/com/github/manolo8/darkbot/gui/login/SavedLogins.java
```diff
@@ -121,7 +122,9 @@ private void updateList() {
 
     @Override
     public LoginForm.Message tryLogin(LoginData login) {
-        Credentials.User user = users.getSelectedValue();
+        Credentials.User user = Bot.getStartupParams().AUTO_LOGIN
+                ? credentials.getUsers().stream().filter(usr -> usr.u.equals(Bot.getStartupParams().USERNAME)).findFirst().orElse(null)
+                : users.getSelectedValue();
```
Pablete1234, line 127: none of this logic belongs in the SavedLogins gui panel

Pablete1234 (changes requested):
You need to pass the params into configManager.getAPI(params), in there, you can edit the LoginUtils.performUserLogin so that if params have autologin, you directly call to login with what's necessary, never calling a login form at all

src/main/java/com/github/manolo8/darkbot/core/api/NativeApiAdapter.java
```diff
@@ -12,8 +13,8 @@
     private static int nextBotId = 0;
     private int botId = -1;
 
-    public NativeApiAdapter() {
-        this.loginData = LoginUtils.performUserLogin();
+    public NativeApiAdapter(StartupParams params) {
+        this.loginData = params.shouldAutoLogin() ? LoginUtils.performAutoLogin(params) : LoginUtils.performUserLogin();
```
Pablete1234, line 17:
```suggestion
        this.loginData = LoginUtils.performUserLogin(params);
```
pass the params and inside the performUserLogin, do 
`if (params != null && params.shouldAutoLogin()) return LoginUtils.performAutoLogin(params)`
It will simplify having this logic in 3 different api adapters, and will put it in one place. Also allow for params to be null.

src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -103,9 +113,14 @@ public Main() {
             Popups.showMessageAsync("Error", I18n.get("bot.issue.config_load_failed"), JOptionPane.ERROR_MESSAGE);
 
         API.createWindow();
+        initializeStartupSettings(params);
```
Pablete1234, line 116:
```suggestion
        if (params.shouldAutoStart()) setRunning(true);
```
This being a single line used in a single place, i'd just avoid the function

src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -103,9 +113,14 @@ public Main() {
...
         start();
     }
 
+    private void initializeStartupSettings(StartupParams params) {
+        if (params.shouldAutoStart()) setRunning(true);
+    }
+
```
Pablete1234, line 123:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/gui/login/LoginForm.java
```diff
@@ -49,9 +49,9 @@ public void setInfoText(Message val) {
     }
 
     public static class Message {
-        private final boolean error;
-        private final String text;
-        private final String description;
+        final boolean error;
+        final String text;
+        final String description;
```
Pablete1234, line 54:
```suggestion
        private final boolean error;
        private final String text;
        private final String description;
```

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,123 @@
...
+                        username = args[++i];
+                        masterPasswordFile = new File(args[++i]);
+                        autoLogin = true;
+                        break;
+                    case START_COMMAND:
+                        autoStart = true;
+                        break;
+                }
+            }
+        } catch (ArrayIndexOutOfBoundsException e) {
```
Pablete1234, line 90: I'm pretty sure catching an out of bounds exception is not the right way to handle things.
Check for length and avoid the exception being thrown, also, provide a descriptive error message of what's missing in the log, instead of "invalid program arguments" say "Missing arguments for auto-login, usage: -login username path/to/masterpass.txt"

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,123 @@
...
+    public char[] getMasterPassword() throws IOException {
+        BufferedReader br = new BufferedReader(new FileReader(masterPasswordFile));
+
+        StringBuilder sb = new StringBuilder();
+        String s;
+        while ((s = br.readLine()) != null) {
+            sb.append(s);
+        }
+
+        return sb.substring(sb.indexOf("\"") + 1, sb.lastIndexOf("\"")).toCharArray();
```
Pablete1234, line 108: Pretty complicated way of reading a file, also, there's no need to have the `password="asdf"` syntax, especially if you're gonna ignore anything before the first ".

```
List<String> lines = Files.readAllLines(Paths.get(path));
if (lines.size() != 1) ;// some error
return lines.get(0).toCharArray();
```

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public boolean shouldAutoLogin() {
```
Pablete1234, line 111:
```suggestion
    public boolean getAutoLogin() {
```

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,123 @@
...
+        return autoLogin;
+    }
+
+    public String getUsername() {
+        return username;
+    }
+
+    public boolean shouldAutoStart() {
```
Pablete1234, line 119:
```suggestion
    public boolean getAutoStart() {
```

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,123 @@
...
+     * Sample content of master password file:
+     * <pre>
+     *     password="mypassword12345"
+     * </pre>
+     * Leave it empty if you didnt set a master password like this:
+     * <pre>
+     *     password=""
+     * </pre>
+     */
+    private File masterPasswordFile;
```
Pablete1234, line 49:
```suggestion
    private Path masterPasswordPath;
```

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,123 @@
...
+     * This is a txt file containing your master file<br>
+     * Sample content of master password file:
+     * <pre>
+     *     password="mypassword12345"
+     * </pre>
+     * Leave it empty if you didnt set a master password like this:
+     * <pre>
+     *     password=""
+     * </pre>
+     */
```
Pablete1234, line 48:
```suggestion
    /** Plaintext file containing the master password */
```
I'm not sure that verbose comments really help much

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,123 @@
...
+     *     <i>username</i> and <i>filepath</i> are required arguments
+     *     <i>username</i> refers to the account you want to login to
+     *     <i>filepath</i> refers to the filepath to your txt file containing your master password
+     * </pre>
+     * <b>SAMPLE USAGE</b>
+     * <pre>
+     *     -login NotABot C:\Users\Owner\masterpw.txt
+     * </pre>
+     * @see #masterPasswordFile
+     */
```
Pablete1234, line 34:
```suggestion
    /**
     * Command-line argument for auto-login, requires username and a path to a plaintext file with the master password.
     * Example usage: {@code -login NotABot C:\Users\Owner\masterpw.txt}
     * @see #username
     * @see #masterPasswordFile
     */
```
It feels like you're writing documentation for end-users, but this is documentation for developers, keep it simple

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,123 @@
...
+     * </pre>
+     * <b>OPTIONS</b>
+     * <pre>
+     *     This command takes no options
+     * </pre>
+     * <b>SAMPLE USAGE</b>
+     * <pre>
+     *     -start
+     * </pre>
+     */
```
Pablete1234, line 68:
```suggestion
    /** Command-line argument for auto-start, has no parameters. */
```

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -51,6 +52,37 @@ public static LoginData performUserLogin() {
         return loginData;
     }
 
+    public static LoginData performAutoLogin(StartupParams params) {
+        Credentials credentials = loadCredentials();
+        try {
+            credentials.decrypt(params.getMasterPassword());
+        } catch (Exception e) {
+            System.err.println("Couldn't login, check your master password txt file");
```
Pablete1234, line 60:
```suggestion
            System.err.println("Couldn't retreive logins, check your master password file");
```

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -51,6 +52,37 @@ public static LoginData performUserLogin() {
...
+            e.printStackTrace();
+        }
+
+        Credentials.User user = credentials.getUsers()
+                .stream()
+                .filter(usr -> usr.u.equals(params.getUsername()))
+                .findFirst()
+                .orElse(null);
+        if (user == null) {
+            System.err.println("Invalid program arguments, make sure you entered the correct username and filepath");
```
Pablete1234, line 70:
```suggestion
            System.err.println("Couldn't find the provided user in the saved logins. Make sure the username and master password are correct.");
```

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -51,6 +52,37 @@ public static LoginData performUserLogin() {
...
+
+        LoginData loginData = new LoginData();
+        loginData.setCredentials(user.u, user.p);
+
+        System.out.println("Auto logging in (1/2)");
+        usernameLogin(loginData);
+        System.out.println("Loading spacemap (2/2)");
+        findPreloader(loginData);
+
+        if (loginData.getPreloaderUrl() == null || loginData.getParams() == null) System.exit(0);
```
Pablete1234, line 82:
```suggestion
        if (loginData.getPreloaderUrl() == null || loginData.getParams() == null) {
          System.err.println("Could not find preloader url or parameters");
          System.exit(0);
        }
```
Before any random system.exit, log it. If any user reports "it's randomly closing" we need to see why in the log

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,118 @@
...
+        return masterPassword == null ? null : masterPassword.toCharArray();
+    }
+
+    private String get(String key) throws IOException {
+        if (properties == null) {
+            properties = new Properties();
+            properties.load(new InputStreamReader(new FileInputStream(startupPropertiesPath), StandardCharsets.UTF_8));
+            System.out.println("Loaded startup properties file");
+            properties.list(System.out);
+        }
```
Pablete1234, line 74: Do this logic when parsing the parameters, not when trying to get values. Getting a value shouldn't throw an IOException

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,118 @@
...
+            String value = get(key.name);
+            if (value == null) {
+                System.err.println("Unable to retrieve " + key.name + ", make sure you have defined this key inside your properties file");
+                if (key.required) System.exit(-1);
+            }
+            return value;
+        } catch (IOException e) {
+            System.err.println("Unable to retrieve " + key.name + ", make sure you have specified the correct path to your properties file");
+            e.printStackTrace();
+        }
```
Pablete1234, line 89: This whole try-catch will be unnecessary if you edit the above.

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private void parse(String[] args) {
+        for (int i = 0; i < args.length; i++) {
+            switch (args[i]) {
+                case LOGIN_COMMAND:
+                    if (i + 1 >= args.length) {
+                        System.err.println("Missing arguments for auto-login, usage: -login path/to/startup.properties");
+                        System.exit(-1);
+                    }
+                    startupPropertiesPath = args[++i];
```
Pablete1234, line 53:
```suggestion
                    properties = loadLoginProperties(args[++i]);
```
Just load the properties here, make parsing be able to throw that IOException

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,118 @@
...
+    private String[] args;
+    private Properties properties;
+
+    public StartupParams(String[] args) {
+        this.args = args;
+        parse(args);
+        System.out.println(this);
+    }
+
+    private void parse(String[] args) {
```
Pablete1234, line 45:
```suggestion
```
I'd merge parse into the constructor, and make that return the IOException, no system exits required, let exception bubble up to Bot.java and let that be the reason it closes down the bot

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -47,7 +50,57 @@ public static LoginData performUserLogin() {
         Popups.showMessageSync("Login", pane, panel::setDialog);
 
         LoginData loginData = panel.getResult();
-        if (loginData.getPreloaderUrl() == null || loginData.getParams() == null) System.exit(0);
+        if (loginData.getPreloaderUrl() == null || loginData.getParams() == null) {
+            System.err.println("Could not find preloader url or parameters");
+            System.exit(-1);
```
Pablete1234, line 55: I'm pretty sure in this case, it means the user closed the panel to login without selecting a user or logging in in any other way, it's not an error to shutdown with -1, rather, 0, and make the it a System.out instead of err, but mention how the user exited without logging in, not that preloader is missing

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -21,11 +23,12 @@ public static void main(String[] args) {
...
         }
-        SwingUtilities.invokeLater(Main::new);
+        StartupParams params = args.length == 0 ? null : new StartupParams(args);
+        SwingUtilities.invokeLater(() -> new Main(params));
     }
 
     public static class NoOpBot {
-        public static void main(String[] args) {
+        public static void main(String[] args) throws IOException {
             ConfigManager.FORCE_NO_OP = true;
```
Pablete1234, line 32: This was a lazy way of passing what should've been a parameter to the API-side. Feel free to remove NoOpBot inner class and just make it one of the startup params `-no-op`, and treat it on the getAPI method in config manager

src/main/java/com/github/manolo8/darkbot/config/ConfigManager.java
```diff
@@ -102,13 +102,13 @@ private Config loadConfig(Path configFile, Path backupFile) {
         return config;
     }
 
-    public IDarkBotAPI getAPI() {
+    public IDarkBotAPI getAPI(StartupParams params) {
         if (FORCE_NO_OP) return new NoopApiAdapter();
```
Pablete1234, line 106:
```suggestion
        if (params.useNoOp()) return new NoopApiAdapter();
```

Pablete1234 (commented):
Just one last thing you can add to cleanup previous code, otherwise good to go

kekawce:
:pogU:

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,128 @@
...
+                    }
+                    properties = loadLoginProperties(args[++i]);
+                    autoLogin = true;
+                    break;
+                case START_COMMAND:
+                    autoStart = true;
+                    break;
+                case NO_OP_COMMAND:
+                    forceNoOp = true;
+                    break;
```
Pablete1234, line 64:
```suggestion
                case NO_OP_COMMAND:
                    forceNoOp = true;
                    break;
                default:
                    System.out.println("Unknown startup argument: " + args[i]);
```
I'd add a default here to make sure if you mistype an argument it's displayed

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+    }
+
+    private Properties loadLoginProperties(String path) throws IOException {
+        InputStreamReader reader = new InputStreamReader(new FileInputStream(path), StandardCharsets.UTF_8);
+        Properties p = new Properties();
+        p.load(reader);
+        reader.close();
```
Pablete1234, line 73:
```suggestion
        try (InputStreamReader reader = new InputStreamReader(new FileInputStream(path), StandardCharsets.UTF_8)) {
            Properties p = new Properties();
            p.load(reader);
        }
```
Prefer a try-with-resources to explicit closing

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,128 @@
...
+
+        System.out.println("Loaded startup properties file");
+        p.list(System.out);
```
Pablete1234, line 76: printing user & pass to console (and log) ? not a good idea
```suggestion
```

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,128 @@
...
+        PASSWORD("password", false),
+        MASTER_PASSWORD("master-password", false);
+
+        private final String name;
+        private final boolean required;
+        PropertyKey(String name, boolean required) {
+            this.name = name;
+            this.required = required;
+        }
+    }
```
Pablete1234, line 110:
```suggestion
    public enum PropertyKey {
        USERNAME, PASSWORD, MASTER_PASSWORD;
    }
```
Simplify this

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -0,0 +1,128 @@
...
+        return properties.getProperty(key);
+    }
+
+    public String get(PropertyKey key) {
+        String value = get(key.name);
+        if (value == null) {
+            System.err.println("Unable to retrieve " + key.name + ", make sure you have defined this key inside your properties file");
+            if (key.required) System.exit(-1);
+        }
+        return value;
```
Pablete1234, line 96:
```suggestion
        return get(key.name().toLowerCase(Locale.ROOT));
```
Use the enum name directly, let the file use `master_password` instead of the `-`
Avoid exiting here, it is definetly not needed, just return a null

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -47,7 +50,57 @@ public static LoginData performUserLogin() {
...
+                    .findFirst()
+                    .orElse(null);
+            if (user == null) {
+                System.err.println("Couldn't find the provided user in the saved logins. Make sure the username and master password are correct.");
+            }
+        }
+
+        LoginData loginData = new LoginData();
+        loginData.setCredentials(params.get(StartupParams.PropertyKey.USERNAME),
+                (password == null || password.isEmpty()) && user != null ? user.p : password);
```
Pablete1234, line 89:
```suggestion
        String username = params.get(StartupParams.PropertyKey.USERNAME),
                masterPw = params.get(StartupParams.PropertyKey.MASTER_PASSWORD),
                password = params.get(StartupParams.PropertyKey.PASSWORD);
        if (username == null || (masterPw == null && (password == null || password.isEmpty()))) {
            System.err.prirntln("Credentials file requires username and either a password or a master password");
            System.exit(-1);
        }
        if (password == null || password.isEmpty()) {
            Credentials credentials = loadCredentials();
            try {
                char[] masterPassword = params.getMasterPassword();
                if (masterPassword == null) {
                    System.err.println("Both master-password and password are undefined, make sure to at least define one of these fields");
                }
                credentials.decrypt(masterPassword);
            } catch (Exception e) {
                System.err.println("Couldn't retreive logins, check your startup properties file");
                e.printStackTrace();
            }
            
            Credentials.User user = credentials.getUsers()
                    .stream()
                    .filter(usr -> username.equals(usr.u))
                    .findFirst()
                    .orElse(null);
            if (user == null) {
                System.err.println("Couldn't find the provided user in the saved logins. Make sure the username and master password are correct.");
                System.exit(-1);
            } else {
                password = user.p;
            }
        }

        LoginData loginData = new LoginData();
        loginData.setCredentials(username, password);
```

I'd consider moving to a different method that is `private String getPassword(String user, String masterPassword)`, returns null if user not found (one with no exits!). It would simplify this function quite a bit

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -47,10 +50,69 @@ public static LoginData performUserLogin() {
...
+        String username = params.get(StartupParams.PropertyKey.USERNAME);
+        String password = params.get(StartupParams.PropertyKey.PASSWORD);
+        char[] masterPw = params.getMasterPassword();
+
+        if (username == null || (masterPw == null && (password == null || password.isEmpty()))) {
+            System.err.println("Credentials file requires username and either a password or a master password");
+            System.exit(-1);
+        }
+
+        password = getPassword(username, password, masterPw);
```
Pablete1234, line 70:
```suggestion
        if (username != null && (password == null || password.isEmpty)) {
            char[] masterPw = params.getMasterPassword();
            password = getPassword(user, masterPw);
            if (password == null)
                System.err.println("Password for user couldn't be retreived. Check that the user exists and master password is correct.");
        }

        if (username == null || password == null || password.isEmpty()) {
            System.err.println("Credentials file requires username and either a password or a master password");
            System.exit(-1);
        }
```

This way we keep a single exit that will run if a password wasn't provided, and none could be obtained (by using the master pw)

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -47,10 +50,69 @@ public static LoginData performUserLogin() {
...
+                    .orElse(null);
+            if (user == null) {
+                System.err.println("Couldn't find the provided user in the saved logins. Make sure the username and master password are correct.");
+                password = null;
+            } else {
+                password = user.p;
+            }
+        }
+        return password;
+    }
```
Pablete1234, line 114: What's the point of "getPassword" if the requested parameter is the password?
```suggestion
    private static String getPassword(String username, char[] masterPassword) {
        Credentials credentials = loadCredentials();
        try {
            credentials.decrypt(masterPassword);
        } catch (Exception e) {
            System.err.println("Couldn't retreive logins, check your startup properties file");
            e.printStackTrace();
            return null;
        }
        return credentials.getUsers().stream()
                .filter(usr -> username.equals(usr.u))
                .findFirst().map(u -> u.p).orElse(null);
    }
```
Simplify the function, return null if any issue occurs getting the user.
Possibly even consider `orElseThrow` and throw an exception for missing user or something, but that may be a just complicating it for no good reason

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #9, opened 2020-10-23
IGNORE_BOXES flag and logic for Loot&Collect Module

ODAKAB:
I'd still prefer managing it with priority so users could do mixed prioritizing like Box1 > Npc1 > Box2

My opinion:
In this state, it makes more sense than releasing a flag to make it a default feature for Loot&Collect Module


src/main/java/com/github/manolo8/darkbot/modules/LootNCollectorModule.java
```diff
@@ -61,7 +61,7 @@ public void tick() {
                 if (box == null || box.locationInfo.distance(hero) > config.COLLECT.RADIUS
                         || lootModule.attack.target.health.hpPercent() < 0.25) {
                     lootModule.moveToAnSafePosition();
-                } else {
+                } else if (!(lootModule.attack.target.npcInfo.extra.has(NpcExtra.IGNORE_BOXES) && hero.locationInfo.distance(lootModule.attack.target) > lootModule.attack.target.npcInfo.radius*1.2)) {
```
Pablete1234, line 64: This line is way too long, cut it

Pablete1234 (changes requested):

ODAKAB (commented):
.

ODAKAB:
```
if ((lootModule.attack.target.npcInfo.extra.has(NpcExtra.IGNORE_BOXES)
                        && hero.locationInfo.distance(lootModule.attack.target) > lootModule.attack.target.npcInfo.radius*1.2)
                        || box == null || box.locationInfo.distance(hero) > config.COLLECT.RADIUS
                        || lootModule.attack.target.health.hpPercent() < 0.25) {
                    lootModule.moveToAnSafePosition();
                } else {
                    lootModule.setConfig(collectorModule.current.locationInfo.now);
                    collectorModule.tryCollectNearestBox();
                }
```

PR #14, opened 2020-10-28
anti pushing

kekawce:
image


src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -182,6 +183,7 @@ private void tickRunning() {
         guiManager.pet.tick();
         guiManager.group.tick();
         checkRefresh();
+        checkPushing();
```
Pablete1234, line 186: put this on a native behaviour, not literally on main

Pablete1234 (changes requested):

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #24, opened 2020-11-14
Re-login on SID KO

kekawce:
also need punis new darkboat api (version 4) to use this

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/api/DarkBoatAdapter.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/api/DarkBoatAdapter.java
@@ -13,18 +13,14 @@ public class DarkBoatAdapter extends ApiAdapter {
 
     @Override
     public void createWindow() {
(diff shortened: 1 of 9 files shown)


src/main/java/com/github/manolo8/darkbot/core/api/DarkBoatAdapter.java
```diff
@@ -24,6 +22,15 @@ public void createWindow() {
         apiThread.start();
     }
 
+    @Override
+    public void relogin() {
+        super.relogin();
+        String url = "https://" + loginData.getUrl() + "/",
+                sid = "dosid=" + loginData.getSid();
+
+        API.setData(url, sid, loginData.getPreloaderUrl(), loginData.getParams());
```
Pablete1234, line 31: make a setData function with this, and reuse it in createWindow and here (consider the super's relogin utilizing that directly

Pablete1234 (approved):

Pablete1234 (approved):

PR #26, opened 2020-11-17
Plugin Updater

kekawce:


src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginHandler.java
```diff
@@ -120,6 +122,8 @@ private void loadPlugins(File[] pluginFiles) {
             try {
                 pl = new Plugin(pluginFile, pluginFile.toURI().toURL());
                 loadPlugin(pl);
+                checkUpdate(pl);
+                loadPlugin(pl);
```
Pablete1234, line 126: Loading plugins should NOT cause an update, also, loading twice?

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginHandler.java
```diff
@@ -139,16 +143,37 @@ private void loadPlugin(Plugin plugin) throws IOException, PluginLoadingExceptio
...
+                        plugin.getDefinition().version, plDef.version),
+                JOptionPane.INFORMATION_MESSAGE);
+    }
+
+    private PluginDefinition readPlugin(InputStream is) throws IOException {
+        PluginDefinition plDef;
+        try (InputStreamReader isr = new InputStreamReader(is, StandardCharsets.UTF_8)) {
+            plDef = GSON.fromJson(isr, (Type) PluginDefinition.class);
+        }
+        return plDef;
```
Pablete1234, line 174:
```suggestion
        try (InputStreamReader isr = new InputStreamReader(is, StandardCharsets.UTF_8)) {
            return GSON.fromJson(isr, (Type) PluginDefinition.class);
        }
```
I belive that is all you need, since any exception will simply not need a return (no catch)

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginHandler.java
```diff
@@ -139,16 +143,37 @@ private void loadPlugin(Plugin plugin) throws IOException, PluginLoadingExceptio
...
+        try (InputStream is = plDef.download.openConnection().getInputStream()) {
+            Files.copy(is, plugin.getFile().toPath(), StandardCopyOption.REPLACE_EXISTING);
+        }
+        Popups.showMessageAsync(I18n.get("plugins.update.title"),
+                I18n.get("plugins.update.message", plugin.getName(),
+                        plugin.getDefinition().version, plDef.version),
+                JOptionPane.INFORMATION_MESSAGE);
+    }
+
+    private PluginDefinition readPlugin(InputStream is) throws IOException {
```
Pablete1234, line 169: call this readPluginJson or readPluginDefinition or readDefinition. Calling it `readPlugin` makes you think it's going to read the `.jar` file and is misleading

Pablete1234 (changes requested):
This is an extremely naive plugin updater, there are many key features missing here.

 - The user needs to be in control, your software shouldn't break without you touching it
 - Updating needs to be a completely different flow to loading
 - Updating should download to the updates folder, then trigger a refresh
 - Loading should be modified to keep a backup in case of an update
 - A single pop-up or interface should show all information regarding plugin updates, not one pop-up per plugin

I belive the best implementation is that on startup the bot asynchronously checks for plugin updates, and if any exist, makes a download button visible next to plugin update icon. Upon click by the user, it shows a GUI with information related to the updates:
 - For each plugin with an update, show the current version and the new version
 - In case the new version requires a higher bot version, don't allow the download and make it obvious to the user that the updated version wouldn't run.
 - Buttons to update individual plugins should appear, as well as a button to update all plugins.
 - Probably should consider adding some kind of progress indicator when downloading plugins via this gui.
 - When all selected plugin updates have been downloaded, trigger a plugin refresh.
 - Consider maybe adding an automatic revert if features fail to load after the update, but this can be a feature for a later date (as long as the user can manually revert)

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginName.java
```diff
@@ -1,15 +1,48 @@
...
+    private final Plugin plugin;
+
+    private final UpdateProgressBar progressBar;
+    private final JLabel progressLabel;
+    private final UpdateButton updateButton;
+
+    volatile boolean canReload;
+    UpdateTask task;
+
+    PluginName(Main main, Plugin plugin, UpdateProgressBar progressBar, JLabel progressLabel) {
```
Pablete1234, line 35: This is looking very, VERY complicated for what a "name" should be. This MUST be simplified to the minimum

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginUpdateHeader.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private final PluginHandler pluginHandler;
+
+    private final Title title;
+    private final UpdateAllButton updateAllButton;
+    private volatile UpdateProgressBar progressBar;
+
+    private final List<PluginName> names;
+
+    PluginUpdateHeader(Main main, List<PluginName> namess) {
```
Pablete1234, line 33: That moment when a gui element has a List of other GUI elements, but it's not to display them but rather to do logic on them

That logic MUST go away from PluginName (GUI) into a logic-based object

Pablete1234 (changes requested):

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,289 @@
...
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+
+//todoo i18n
+public class PluginUpdater {
+
+    private Date lastChecked = new Date();
```
Pablete1234, line 25: either go fully primitive and use a long (like most of the code) or just use a java8 Instant

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,289 @@
...
+
+    public final List<Plugin> UPDATES = new ArrayList<>();
```
Pablete1234, line 27: Just avoid holding the list to plugins, use pluginHandler.getPlugins() and ignore the plugins update statuses that aren't AVAILABLE

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,289 @@
...
+        void setText(String text) {
+            progressLabel.setText(text);
+        }
+    }
+
+    public PluginUpdater(Main main) {
+        this.pluginHandler = main.pluginHandler;
+    }
+
+    public void scheduleUpdateChecker() {
```
Pablete1234, line 58: Get  rid of this method as a whole, `checkUpdates()` should already be asynchronous

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,289 @@
...
+
+    public boolean hasNoUpdates() {
+        return UPDATES.isEmpty();
+    }
+
+    public Date getLastChecked() {
+        return lastChecked;
+    }
+
+    public void checkUpdates() {
```
Pablete1234, line 92: This needs to be an async method, start a SwingWorker to search updates, and disable the button that allows searching while already running

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,289 @@
...
+
+    private final PluginHandler pluginHandler;
+
+    private UpdateProgressBar mainProgressBar;
+    private final List<PluginProgress> pluginProgresses = new ArrayList<>();
+
+    private static class PluginProgress {
+        private final Plugin plugin;
+        private final UpdateProgressBar progressBar;
+        private final JLabel progressLabel;
```
Pablete1234, line 37: Feels like ur holding a bit too much of GUI on logic-code

Consider having just PluginCard (s) and having a method there that is setUpdateProgress(x), that automatically update progress bar, sets it visible or invisible, as well as the label etc

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginCard.java
```diff
@@ -29,8 +30,14 @@
         setColor(plugin.getIssues());
         plugin.getIssues().addListener(this::setColor);
 
+        UpdateProgressBar progressBar = new UpdateProgressBar();
+        JLabel progressLabel = new JLabel();
+        main.pluginUpdater.add(plugin, progressBar, progressLabel);
```
Pablete1234, line 35: Plugin updater shouldn't need to be holding this stuff.

Make progressbar and progress label attributes saved in PluginCard, add a public method in PluginCard for setUpdateProgress(x) that just updates the label and progressbar.
The PluginCard s should probably still not be sent to the PluginUpdater, but rather available in PluginPanel.getCard(plugin) or something along the lines

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginDisplay.java
```diff
@@ -42,15 +45,20 @@ private JComponent setupUI() {
...
+        pluginPanel.add(new PluginUpdateHeader(pluginUpdater, this));
         Stream.concat(
                 pluginHandler.LOADING_EXCEPTIONS.stream().map(ExceptionCard::new),
                 Stream.concat(
                         pluginHandler.FAILED_PLUGINS.stream(),
                         pluginHandler.LOADED_PLUGINS.stream()
                 ).map(pl -> new PluginCard(main, pl, main.featureRegistry))
         ).forEach(pluginPanel::add);
+
+        //todo maybe also add another icon here when update is available?
```
Pablete1234, line 61: A plugins with a blue `!` would be good for that

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginDisplay.java
```diff
@@ -42,15 +45,20 @@ private JComponent setupUI() {
         return scrollPane;
     }
 
-    private void refreshUI() {
+    void refreshUI() {
         pluginPanel.removeAll();
+        pluginUpdater.clear();
+
+        pluginPanel.add(new PluginUpdateHeader(pluginUpdater, this));
```
Pablete1234, line 52: The update header shouldn't be inside the plugin panel, that's the scroll pane.

Modify setupUI to include:
- `if (pluginPanel != null) return;` 
 - `add(new PluginUpdateHeader(...));`
 - `add(scrollPane)` at the end
 - make the method return void.

Then on line 34 remove the call to "add", as well as the IF, just call setupUI();

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginDisplay.java
```diff
@@ -1,61 +1,83 @@
...
+    private void setupUI() {
+        if (pluginPanel != null) return;
         pluginPanel = new JPanel(new MigLayout("wrap 1, fillx", "[fill]", ""));
-        JScrollPane scrollPane = new JScrollPane(pluginPanel, JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED, JScrollPane.HORIZONTAL_SCROLLBAR_NEVER);
 
-        scrollPane.setBorder(BorderFactory.createEmptyBorder());
-        scrollPane.getVerticalScrollBar().setUnitIncrement(25);
-        return scrollPane;
+        add(new PluginUpdateHeader(pluginUpdater));
+        add(new ScrollPane(pluginPanel));
```
Pablete1234, line 46: There is absolutely no need to subclass JScrollPane here, it's a pretty concise 3-line code used only on one place
```suggestion
        JScrollPane scrollPane = new JScrollPane(pluginPanel, JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED, JScrollPane.HORIZONTAL_SCROLLBAR_NEVER);

        scrollPane.setBorder(BorderFactory.createEmptyBorder());
        scrollPane.getVerticalScrollBar().setUnitIncrement(25);
        add(scrollPane);
```

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginDisplay.java
```diff
@@ -1,61 +1,83 @@
...
     }
 
-    private void refreshUI() {
+    public void refreshUI() {
+        removeAll();
```
Pablete1234, line 50: Don't remove all, keep stuff where it is and where it belongs

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginDisplay.java
```diff
@@ -1,61 +1,83 @@
...
                 pluginHandler.LOADING_EXCEPTIONS.stream().map(ExceptionCard::new),
                 Stream.concat(
                         pluginHandler.FAILED_PLUGINS.stream(),
                         pluginHandler.LOADED_PLUGINS.stream()
                 ).map(pl -> new PluginCard(main, pl, main.featureRegistry))
         ).forEach(pluginPanel::add);
-        pluginTab.setIcon(UIUtils.getIcon(pluginHandler.LOADING_EXCEPTIONS.isEmpty() && pluginHandler.FAILED_PLUGINS.isEmpty() ? "plugins" : "plugins_warn"));
+
+        add(new PluginUpdateHeader(pluginUpdater));
+        add(new ScrollPane(pluginPanel));
```
Pablete1234, line 62: Don't re-add stuff, keep the old objects whenever possible, this isn't any dynamic part of the GUI (like plugin cards are) so keep it as it is

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,284 @@
...
+                checkUpdates(null);
+            }
+        }, 0, Time.DAY);
+    }
+
+    public void setPluginDisplay(PluginDisplay display) {
+        pluginDisplay = display;
+    }
+
+    public void setMainProgressBar(UpdateProgressBar progressBar) {
```
Pablete1234, line 58: Shouldn't be needed if you have the plugin display directly

src/main/java/com/github/manolo8/darkbot/gui/plugins/UpdateProgressBar.java
```diff
@@ -0,0 +1,11 @@
+package com.github.manolo8.darkbot.gui.plugins;
+
+import javax.swing.*;
+
+public class UpdateProgressBar extends JProgressBar {
+    public UpdateProgressBar() {
+        super();
```
Pablete1234, line 7:
```suggestion
```
If this is barely used, make sure it is needed, if all you do is set 2 values, may as well avoid subclassing

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/extensions/util/Version.java
```diff
@@ -31,10 +31,21 @@ public Version(String version) {
...
-        if (beta == -1 && alpha == -1) beta = Integer.MAX_VALUE;
-        if (alpha == -1) alpha = Integer.MAX_VALUE;
+
+        int tmpBeta = getInt(matcher, 8);
+        int tmpAlpha = getInt(matcher, 10);
+        beta = tmpBeta == -1 && tmpAlpha == -1 ? Integer.MAX_VALUE : tmpBeta;
+        alpha = tmpAlpha == -1 ? Integer.MAX_VALUE : tmpAlpha;
+    }
+
+    public Version(Version version) {
```
Pablete1234, line 41: Is this needed/used?

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,288 @@
...
+                Files.copy(plugin.getFile().toPath(), PluginHandler.PLUGIN_OLD_PATH.resolve(plugin.getFile().getName()), StandardCopyOption.REPLACE_EXISTING);
+                progressManager.setUpdateProgress(ProgressManager.State.DOWNLOADING);
+
+                FileUtils.createDirectory(PluginHandler.PLUGIN_UPDATE_PATH);
+                Files.copy(is, PluginHandler.PLUGIN_UPDATE_PATH.resolve(plugin.getFile().getName()), StandardCopyOption.REPLACE_EXISTING);
+                progressManager.setUpdateProgress(ProgressManager.State.WAITING_RELOAD);
+
+                if (updateAllTask != null) return null;
+
+                progressManager.setUpdateProgress(ProgressManager.State.RELOADING);
```
Pablete1234, line 223: You're setting waiting reload & reloading at the same time, consider just doing just one of the 2

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,288 @@
...
+        protected void done() {
+            plugin.setUpdateStatus(failed ? Plugin.UpdateStatus.FAILED : Plugin.UpdateStatus.UP_TO_DATE);
+            if (updateAllTask != null || failedUpdate.length() <= 0) return;
+
+            Popups.showMessageAsync("Update failed", failedUpdate, JOptionPane.ERROR_MESSAGE);
+        }
+
+        @Override
+        protected Void doInBackground() throws Exception {
+            progressManager.setUpdateProgress(ProgressManager.State.STARTING);
```
Pablete1234, line 205: You can't modify UI code from background thread. You MUST do this from the `process` method of the worker.
Call `publish(State.STARTING)`, and on process grab the latest state and push it via `card.setUpdateProgress(state)` (see the comment in the ProgressManager class)

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/extensions/plugins/Plugin.java
```diff
@@ -33,10 +41,26 @@ public void setDefinition(PluginDefinition definition) {
...
+    }
+
+    public void setUpdateStatus(UpdateStatus status) {
+        updateStatus = status;
+    }
+
+    public void add(PluginIssue issue) {
+        updateIssues.getIssues().add(issue);
+    }
+
```
Pablete1234, line 55:
```suggestion
```
This is a no-go, since when is "adding" to a plugin is intuitive for "adding to the update issues" ? It'd be very confusing for any other develiper, just use `plugin.getUpdateIssues().add(issue)` which is very explicit about what it's doing

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,258 @@
...
+            @Override
+            protected void done() {
+                pluginDisplay.refreshUI();
+            }
+
+            @Override
+            protected Void doInBackground() {
+                for (Plugin plugin : pluginHandler.LOADED_PLUGINS) {
+                    plugin.getUpdateIssues().getIssues()
+                            .removeIf(pl -> pl.getMessage().equals(I18n.get("plugins.update_issues.download_failed")));
```
Pablete1234, line 74: Make this check against a constant for the plugin issue, doesn't seem like this has parameters

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,258 @@
...
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
+
+//todoo i18n
+public class PluginUpdater {
+
+    public final List<PluginException> UPDATING_EXCEPTIONS = new ArrayList<>();
+    private long lastChecked = System.currentTimeMillis();
```
Pablete1234, line 25:
```suggestion
    private volatile long lastChecked = System.currentTimeMillis();
```
This is used by different threads, make it volatile

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    private class UpdateAllTask extends SwingWorker<Void, Integer> {
+
+        private int progress = 0;
+        private final List<UpdateTask> updateTasks = new ArrayList<>();
+
+        UpdateAllTask() {
+            List<Plugin> availableUpdates = pluginHandler.getAvailableUpdates().collect(Collectors.toList());
+            availableUpdates.forEach(pl -> updateTasks.add(new UpdateTask(pl, this)));
```
Pablete1234, line 133:
```suggestion
            availableUpdates = pluginHandler.getAvailableUpdates().map(pl -> new UpdateTask(pl, this).collect(Collectors.toList());
```
no need to collect to a list to later do a for-each, you can simply collect to one list directly

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    public void updateAll() {
+        new UpdateAllTask().execute();
+    }
+
+    private class UpdateAllTask extends SwingWorker<Void, Integer> {
+
+        private int progress = 0;
+        private final List<UpdateTask> updateTasks = new ArrayList<>();
```
Pablete1234, line 129:
```suggestion
        private final List<UpdateTask> updateTasks;
```

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,258 @@
...
+
+            if (failed) {
+                card.setUpdateProgress(PluginCard.UpdateStatus.FAILED);
+                // refreshUI() if updating individually and failed to generate ExceptionCards
+                pluginDisplay.refreshUI();
+            }
+            else {
+                plugin.setUpdateStatus(Plugin.UpdateStatus.UP_TO_DATE);
+                card.setUpdateProgress(PluginCard.UpdateStatus.DONE);
+            }
```
Pablete1234, line 216:
```suggestion
            if (!failed) plugin.setUpdateStatus(Plugin.UpdateStatus.UP_TO_DATE);
            card.setUpdateProgress(failed ? PluginCard.UpdateStatus.FAILED : PluginCard.UpdateStatus.DONE);
            if (failed) pluginDisplay.refreshUI(); // Ensure the exception card is shown
```

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginUpdater.java
```diff
@@ -0,0 +1,258 @@
...
+                FileUtils.createDirectory(PluginHandler.PLUGIN_OLD_PATH);
+                Files.copy(plugin.getFile().toPath(), PluginHandler.PLUGIN_OLD_PATH.resolve(plugin.getFile().getName()), StandardCopyOption.REPLACE_EXISTING);
+                publish(PluginCard.UpdateStatus.DOWNLOADING);
+
+                FileUtils.createDirectory(PluginHandler.PLUGIN_UPDATE_PATH);
+                Files.copy(is, PluginHandler.PLUGIN_UPDATE_PATH.resolve(plugin.getFile().getName()), StandardCopyOption.REPLACE_EXISTING);
+
+                if (isUpdatingAll) {
+                    publish(PluginCard.UpdateStatus.INDIVIDUALLY_DONE);
+                    plugin.setUpdateStatus(Plugin.UpdateStatus.UP_TO_DATE);
```
Pablete1234, line 238: the up-to-date is set here for all-updates, but on done for single-updates? I think you should make that happen on the same place, probably right before this `if (isUpdatingAll) {`

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginName.java
```diff
@@ -8,8 +9,10 @@
 
 class PluginName extends JPanel {
 
-    PluginName(PluginDefinition definition) {
+    PluginName(Plugin plugin) {
         super(new MigLayout("", "[]5px[]5px[]5px[]"));
+
+        PluginDefinition definition = plugin.getDefinition();
```
Pablete1234, line 15: Undo this change, just the definition is required to print the name

src/main/java/com/github/manolo8/darkbot/utils/FileUtils.java
```diff
@@ -0,0 +1,17 @@
+package com.github.manolo8.darkbot.utils;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class FileUtils {
+    public static void createDirectory(Path path) {
```
Pablete1234, line 8: Probably shound name `ensureDirectoryExists` to indicate it may not always create it

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginUpdateHeader.java
```diff
@@ -0,0 +1,119 @@
...
+        private void refresh() {
+            setEnabled(true);
+
+            StringBuilder toolTipText = new StringBuilder();
+            pluginHandler.getAvailableUpdates()
+                    .map(pl -> I18n.get("plugins.config_button.update_all.desc",
+                            pl.getName(),
+                            pl.getDefinition().version,
+                            pl.getUpdateDefinition().version) + "\n")
+                    .forEach(toolTipText::append);
```
Pablete1234, line 98:
```suggestion
            String toolTipText = pluginHandler.getAvailableUpdates()
                    .map(pl -> I18n.get("plugins.config_button.update_all.desc",
                            pl.getName(),
                            pl.getDefinition().version,
                            pl.getUpdateDefinition().version))
                    .collect(Collectors.joining("\n");
```
You can also get rid of removing the last `\n` etc

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginName.java
```diff
@@ -1,15 +1,31 @@
...
 class PluginName extends JPanel {
 
-    PluginName(PluginDefinition definition) {
-        super(new MigLayout("", "[]5px[]5px[]5px[]"));
+    private final Plugin plugin;
+    private final PluginUpdater pluginUpdater;
+
+    private final UpdateButton updateButton;
+
+    PluginName(Plugin plugin, PluginUpdater pluginUpdater) {
```
Pablete1234, line 21: Too much logic for a "name", look at the other comment but essentially keep the button functionality outside, simply pass as parameter the button to display, keep the button as part of the plugin card

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginUpdateHeader.java
```diff
@@ -0,0 +1,117 @@
...
+        this.updateAllButton = new UpdateAllButton();
+        this.checkUpdateButton = new CheckUpdateButton();
+
+        progressBar.setVisible(false);
+        progressBar.setBorderPainted(false);
+
+        add(title);
+        add(updateAllButton);
+        add(checkUpdateButton);
+        add(progressBar, "dock south, spanx");
```
Pablete1234, line 48:
```suggestion
        add(this.title = new Title());
        add(this.updateAllButton = new UpdateAllButton());
        add(this.checkUpdateButton = new CheckUpdateButton());
        add(this.progressBar = new JProgressBar(), "dock south, spanx");
        
        progressBar.setBorderPainted(false);
        
        refreshUI();
```

You're repeating some code from here and the refreshUI method, you can simplify it a bit

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginUpdateHeader.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public JProgressBar getProgressBar() {
+        return progressBar;
+    }
+
+    private class Title extends JLabel {
+        private Title() {
+            super(titleFormatter.format(new Object[]{status, dateFormatter.format(pluginUpdater.getLastChecked())}));
+        }
+    }
```
Pablete1234, line 68: Just use a jlabel for this directly, you don't even need to set a value at the start since you'll call refreshUI which will call setText for you already

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginUpdateHeader.java
```diff
@@ -0,0 +1,117 @@
...
+        add(progressBar, "dock south, spanx");
+    }
+
+    void refreshUI() {
+        updateAllButton.refresh();
+        progressBar.setVisible(false);
+
+        checkUpdateButton.setEnabled(true);
+        status = !pluginUpdater.hasAnyUpdates() ? I18n.get("plugins.config_button.up_to_date") + " ," : "";
+        title.setText(titleFormatter.format(new Object[]{status, dateFormatter.format(pluginUpdater.getLastChecked())}));
```
Pablete1234, line 57: Not convinced that this titleFormatter is a good idea, you're wrapping a message format in a message format.

```
title.setText(I18n.get("plugins.config_button.update_header",
        !pluginUpdater.hasAnyUpdates() ? I18n.get("plugins.config_button.up_to_date") : "",
        dateFormatter.format(pluginUpdater.getLastChecked()))
```
You can get rid of titleFormatter.

Also, it may make more sense that if there are some updates, it's shown in a message like `{0} updates available` instead of an empty string

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginUpdateHeader.java
```diff
@@ -0,0 +1,107 @@
...
+    void refreshUI() {
+        updateAllButton.refresh();
+        progressBar.setVisible(false);
+
+        checkUpdateButton.setEnabled(true);
+        String status = !pluginUpdater.hasAnyUpdates()
+                ? I18n.get("plugins.config_button.up_to_date")
+                : I18n.get("plugins.config_button.updates_available", pluginHandler.getAvailableUpdates().count());
+        title.setText(I18n.get("plugins.config_button.update_header",
+                status, dateFormatter.format(pluginUpdater.getLastChecked())));
```
Pablete1234, line 53: i'm kinda surprised that the dateFormatter can suck up a `long` directly, but i'd try to use java8 dates for this

Convert the long into an actual instant, then you can use a dateFormatter to format that instant
`Instant lastUpdate = Instant.ofEpochMilli(pluginUpdater.getLastChecked());`

Keep in mind you'll need to swap out the older `SimpleDateFormat` with a java 8 `DateTimeFormatter`.
`DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("EEE, d MMM, hh:mm a");`

Lastly, for i18n purposes you can consider using the date formats provided by java, this way users with different preferences will get the date format in their regional format, example:
`DateTimeFormatter dateFormatter = DateTimeFormatter.ofLocalizedDateTime(FormatStyle.FULL, FormatStyle.FULL)`

The issue is you're trying to print a date without a year, and that has no real standard (see https://stackoverflow.com/questions/9720750/java-local-timeformat-without-year) so you can probably ignore this for now

src/main/java/com/github/manolo8/darkbot/gui/plugins/PluginUpdateHeader.java
```diff
@@ -0,0 +1,107 @@
...
+
+import javax.swing.*;
+import java.awt.event.ActionEvent;
+import java.text.SimpleDateFormat;
+import java.util.Locale;
+import java.util.stream.Collectors;
+
+public class PluginUpdateHeader extends JPanel {
+
+    private static final SimpleDateFormat dateFormatter = new SimpleDateFormat("EEE, d MMM, hh:mm a", Locale.ROOT);
```
Pablete1234, line 18:
```suggestion
    private static final SimpleDateFormat dateFormatter = new SimpleDateFormat("EEE, d MMM, hh:mm a");
```
Do not force locale upon users, see the other comment tho cause this will need to be swapped by a java8 DateTimeFormatter

Pablete1234 (changes requested):

Pablete1234 (approved):
Lots of effort put into this one, now let's see how well it actually does

PR #27, opened 2020-11-19
re-login on sid improvement

kekawce:


Pablete1234:
Different and more fail-safe implementation is already added

PR #29, opened 2020-12-12
Correct IGNORE_BOXES description

ODAKAB:
I suddenly noticed that it's possible to go further further away if there are collectables under 400(default) chained.
So, I'm correcting this mistake in description.


Pablete1234:
The current description is correct, you will not collect boxes while flying to the npc, but nothing prevents collecting when you've arrived to the npc (you're close)

PR #33, opened 2020-12-15
Fix clickable.enabled was not updating

ODAKAB:


ODAKAB:
changed to a proper fix with punis suggestion

DarkBOT.iml
```diff
@@ -1,27 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
Pablete1234, line 1: Leave this file unmodified

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #36, opened 2020-12-28
added bigpoint's user agent for standalone BackpageManager#getConnection

Punikekk:


Pablete1234:
Seeing this one a bit too late, pushed similar fix directly to master already

PR #37, opened 2021-01-02
RefinementGui fix

kekawce:


dm94:
You have to put it in the other if

if (name != null && !name.isEmpty() && !name.contains("ore"))

PR #39, opened 2021-01-03
Fix group members not updating once after size is 0

ODAKAB:


ODAKAB:
This might also fix group invites not being sent after server restarts.

ODAKAB:
I can get rid of hasGroup by checking array size instead.
Or can spam members.clear() before returning and get rid of hasGroup again.
Tell me how would you like to have it.

Pablete1234:
Fixed in cb355a1

PR #40, opened 2021-01-04
Added error popups

LeffPL:
The provided code will check what was the real cause why the bot wasn't able to load the API and display the proper error message.

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/config/ConfigManager.java
+++ b/src/main/java/com/github/manolo8/darkbot/config/ConfigManager.java
@@ -9,25 +9,22 @@ import com.github.manolo8.darkbot.core.api.DarkFlashApiAdapter;
 import com.github.manolo8.darkbot.core.api.NativeApiAdapter;
 import com.github.manolo8.darkbot.core.api.NoopApiAdapter;
 import com.github.manolo8.darkbot.gui.utils.Popups;
+import com.github.manolo8.darkbot.utils.APIErrors;
 import com.github.manolo8.darkbot.utils.StartupParams;
+
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 
 import javax.swing.*;
-import java.awt.*;
-import java.awt.event.MouseAdapter;
-import java.awt.event.MouseEvent;
+
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.IOException;
-import java.net.URI;
-import java.net.URISyntaxException;
+
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-
-import static com.github.manolo8.darkbot.gui.utils.Popups.showMessageSync;
 
 public class ConfigManager {
 
@@ -110,6 +107,7 @@ public class ConfigManager {
     public IDarkBotAPI getAPI(StartupParams params) {
         if (params.useNoOp()) return new NoopApiAdapter();
         try {
+
             if (config.BOT_SETTINGS.API_CONFIG.API == 0) return new DarkBotApiAdapter();
             else if (config.BOT_SETTINGS.API_CONFIG.API == 1) return new DarkFlashApiAdapter(params);
             else if (config.BOT_SETTINGS.API_CONFIG.API == 2) return new DarkBoatAdapter(params);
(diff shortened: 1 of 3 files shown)


Pablete1234 (changes requested):
This code is making many assumptions, many of them may never be true. 
General changes required:
  - You should try to extract most if not all of the strings to the translatable strrings file
  - Encapsulate this code elsewhere, the catch clausule of the error shouldn't be 200 lines long
  - Make the error handling happen first, user-oriented error eplaining later. Otherwise an error in your code may mask away the original error
   - You must not ever check for the system it's running on,  you should check what the error itself is. If the error says it can't run because it needs a 64bit system, then obviously that's a requirement, but not all apis may have this requirement and the code shouldn't assume it

Leff-PL:
Now the catch clausule is much shorter and I removed checking if the OS is 32-bit or not, since I'm not able to get this info from the error message (it's exactly the same as for 32-bit java).

The original error message remains untouched.

src/main/java/com/github/manolo8/darkbot/utils/APIErrors.java
```diff
@@ -0,0 +1,70 @@
...
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import static com.github.manolo8.darkbot.gui.utils.Popups.showMessageSync;
+
+public class APIErrors {
+
```
Pablete1234, line 14:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/utils/APIErrors.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private final String RUNTIME_URL  = "https://darkbot.eu/downloads/Runtimes4DarkBot.exe";
+    private final String RUNTIME_LINK = "Download link: <a href=\"#\">" + RUNTIME_URL;
```
Pablete1234, line 17: When you open an HTML tag you have to close it afterwards

Also have you tested if by adding an A with an actual href it lets you click it without needing to add a mouse listener?

src/main/java/com/github/manolo8/darkbot/utils/APIErrors.java
```diff
@@ -0,0 +1,70 @@
...
+    JPanel pnl = new JPanel();
+
+    private void displayAndExit() {
+        JOptionPane options = new JOptionPane(this.pnl,
+                JOptionPane.ERROR_MESSAGE,
+                JOptionPane.DEFAULT_OPTION,
+                null, new Object[]{}, null
+        );
+        showMessageSync("Error", options);
+        System.exit(0);
```
Pablete1234, line 28: Excuse me what?

None of these errors are any reason to close the bot. The bot itself can run perfectly fine, it simply will have to be on no-op API.

src/main/resources/lang/strings_en.properties
```diff
@@ -475,3 +475,5 @@ colors.fonts.big=Big
 colors.fonts.mid=Medium
 colors.fonts.small=Small
 colors.fonts.tiny=Tiny
+api.errors.unsupported_java=The selected API requires a 64-bit java, but you ran the bot on a 32-bit one<br/>To get a 64-bit java, consider installing our runtime:
+api.errors.non_ansi_path=The bot folder path contains non-ANSI characters. Rename the folders<br/> where they occur or move the bot folder to D:/ and then run again
```
Pablete1234, line 479: Make these messages the FULL html, with parameters for the things you may want to add separately, you can add parameters like `{0}` or `{1}` etc, then when doing `I18n.get` you add the parameters:

`I18n.get("some.string.resource", "param0", "param1")`

src/main/java/com/github/manolo8/darkbot/utils/APIErrors.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public void displayUnsupportedJavaError() {
+        JLabel unSuppJava    = new JLabel("<html>" + I18n.get("api.errors.unsupported_java"));
```
Pablete1234, line 32: You open many html tags that you never close, nevertheless, move the whole message to the string itself and use parameters (see the other comment)

src/main/java/com/github/manolo8/darkbot/utils/APIErrors.java
```diff
@@ -0,0 +1,70 @@
...
+        JLabel clickableLink = new JLabel("<html>" + RUNTIME_LINK);
+        clickableLink.setCursor(Cursor.getPredefinedCursor(Cursor.HAND_CURSOR));
+        clickableLink.addMouseListener(new MouseAdapter() {
+            @Override
+            public void mouseClicked(MouseEvent e) {
+                try {
+                    Desktop.getDesktop().browse(new URI(RUNTIME_URL));
+                } catch (URISyntaxException | IOException ignored) {}
+            }
+        });
```
Pablete1234, line 42: This is quite ugly to add as a standalone, consider adding a util JHyperLink or similar that extends JLabel and has the logic for this

src/main/java/com/github/manolo8/darkbot/config/ConfigManager.java
```diff
@@ -103,22 +107,34 @@ private Config loadConfig(Path configFile, Path backupFile) {
...
             e.printStackTrace();
-            config.BOT_SETTINGS.API_CONFIG.API = 4;
-            Popups.showMessageAsync(
-                    "API failed to load",
-                    "The API you had selected is not able to load.\n" +
-                    "You probably do not have the required DLL in your lib folder.\n" +
-                    "The bot will start on no-operation API, change it in the settings and restart.",
-                    JOptionPane.ERROR_MESSAGE);
+
+            APIErrors err = new APIErrors();
```
Pablete1234, line 120: An instance of APIErrors doesn't feel justified here whatsoever, make the class have static methods.

I think the only thing you should need is something along the lines of:

`String description = ApiErrors.createDescription(e)`

You can append that to the end of the original message, you can move it to the resources too and make it html in itself, with a parameter being the description.

Could end up looking something like:

```
The API you had selected is not able to load.
You probably do not have the required DLL in your lib folder.
The bot will start on no-operation API, change it in the settings and restart

In this case, it seems like you may be missing runtimes, download them here:
https://... 
```
The last part of that message would depend on what you find to be the issue

src/main/java/com/github/manolo8/darkbot/utils/APIErrors.java
```diff
@@ -0,0 +1,70 @@
...
+
+import static com.github.manolo8.darkbot.gui.utils.Popups.showMessageSync;
+
+public class APIErrors {
+
+
+    private final String RUNTIME_URL  = "https://darkbot.eu/downloads/Runtimes4DarkBot.exe";
+    private final String RUNTIME_LINK = "Download link: <a href=\"#\">" + RUNTIME_URL;
+
+    JPanel pnl = new JPanel();
```
Pablete1234, line 19: Make this class stateless and make everything static

src/main/java/com/github/manolo8/darkbot/utils/APIErrors.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public void displayInvalidPathError(String path) {
+        String invalidPathMsg  = "<html>" + I18n.get("api.errors.non_ansi_path") + "<br/><br/>";
+        String formattedPath   = this.formatPath(path);
+        pnl.add(new JLabel(invalidPathMsg + formattedPath));
+        this.displayAndExit();
+    }
+
+    public String formatPath(String str) {
+        String[] folderNames = str.split("\\\\");
```
Pablete1234, line 57: Use the filepath separator constant, not a hard-coded `\`

src/main/java/com/github/manolo8/darkbot/config/ConfigManager.java
```diff
@@ -103,22 +107,34 @@ private Config loadConfig(Path configFile, Path backupFile) {
...
+
+            final String WORKING_DIR  = System.getProperty("user.dir");
```
Pablete1234, line 122: Get the path from simply new File() or Paths, no need to access a system property for this

src/main/java/com/github/manolo8/darkbot/config/ConfigManager.java
```diff
@@ -103,22 +107,34 @@ private Config loadConfig(Path configFile, Path backupFile) {
...
+            final String ERR_MSG      = e.getMessage();
```
Pablete1234, line 123: The error is a nice thing if you wanna show it to the user, you can also use getLocalizedMessage to show it in the users' language

src/main/java/com/github/manolo8/darkbot/config/ConfigManager.java
```diff
@@ -103,22 +107,34 @@ private Config loadConfig(Path configFile, Path backupFile) {
...
+            final String JAVA_32_ERR  = "not found in resource path";
+
+            if      (ERR_MSG.contains(JAVA_32_ERR))          err.displayUnsupportedJavaError();
```
Pablete1234, line 126: However, you should not be checking a contains in an error message. You never know if a different JVM implementation may show a different message and the program should never break due to that. Don't rely on the message, rely on the error instance

src/main/java/com/github/manolo8/darkbot/config/ConfigManager.java
```diff
@@ -103,22 +107,34 @@ private Config loadConfig(Path configFile, Path backupFile) {
...
+            else if (!WORKING_DIR.matches("[\\x00-\\x7F]+")) err.displayInvalidPathError(WORKING_DIR);
```
Pablete1234, line 127: Lastly, a suggested above, make the actual checking/deciding what error you show hidden away inside ApiErrors, none of that logic should be here

src/main/java/com/github/manolo8/darkbot/utils/APIErrors.java
```diff
@@ -0,0 +1,70 @@
...
+        String charPointers  = str
+                .replaceAll("[\\x00-\\x7F]", "&nbsp")
+                .replaceAll("[^\\x00-\\x7F]", "^");
+        StringBuilder path = new StringBuilder("<p style=\"font-family: Consolas\">");
+        for (String name : folderNames) {
+            path.append(!name.matches("[\\x00-\\x7F]+")
+                    ? "<font color=\"#ff7d7d\">" + name + "/</font>"
+                    : name + '/'
+            );
+        }
```
Pablete1234, line 67: There's no need to do any for-loop, you can replace all of this with regex directly

Pablete1234 (changes requested):

Pablete1234:
Since this has already been partially implemented and doesn't have much activity, i'll close

PR #42, opened 2021-02-06
Tweak min wait before revive

ODAKAB:
If you set WAIT_BEFORE_REVIVE as low as 2 sometimes it just causes you to be stuck on death screen without the repair popup and if you refresh you find out client/server didn't register the repair.


pajtor:
I would put step = 1 and min value 5

ODAKAB:
Step 10 doesn't make sense for me too but issue isn't related to it, change it on my patch-1 or open new pr if you think so :peeposhrug:

PR #49, opened 2021-02-09
Fix attacker assumes bugged while npc has ish

ODAKAB:
Attacker thinks attack is bugged if npc hp doesn't drop but also we are in range(under 700)
Added npc ish check to improve hp check, tested and works.

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
+++ b/src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
@@ -110,7 +110,7 @@ public class NpcAttacker {
             sendAttack(250, 5000, true);
         } else if (!hero.isAttacking(target) || !hero.isAiming(target)) {
             sendAttack(1500, 5000, false);
-        } else if (target.health.hpDecreasedIn(1500) || hero.locationInfo.distance(target) > 700) {
+        } else if ((target.health.hpDecreasedIn(1500) || target.hasEffect(16)) || hero.locationInfo.distance(target) > 700) {
             isAttacking = Math.max(isAttacking, System.currentTimeMillis() + 2000);
         } else if (System.currentTimeMillis() > isAttacking) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -110,7 +110,7 @@ protected void tryAttackOrFix() {
             sendAttack(250, 5000, true);
         } else if (!hero.isAttacking(target) || !hero.isAiming(target)) {
             sendAttack(1500, 5000, false);
-        } else if (target.health.hpDecreasedIn(1500) || hero.locationInfo.distance(target) > 700) {
+        } else if ((target.health.hpDecreasedIn(1500) || target.hasEffect(16)) || hero.locationInfo.distance(target) > 700) {
```
Pablete1234, line 113: Avoid using magic numbers, use the effect constants when possible target.hasEffect(Effect.NPC_ISH)

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
ODAKAB: paranthesis was there to specify ish check is meant to work together with hp check which isn't related to distance, helps reader by being more understandable for future
if you agree i'll change to effect name only

ODAKAB:
Logic exceeds one line after last change, can group it under a method isAttackValid or similar

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
ODAKAB: https://github.com/darkbot-reloaded/DarkBot/pull/49/commits/22e09ec75c29ed4437f946bc19509667c9592bc1

Pablete1234 (approved):

PR #52, opened 2021-02-20
added getId()

pajtor:
in order to make it work like Ammo way of getting ID, I suggest this method


Pablete1234:
There's very little reason for this method to exist at all, how the formation is handled in-game is detail implementation of the api itself, and on the api user side you should simply use the enum itself as parameter, not such a string

PR #55, opened 2021-03-04
Cleanup NpcAttacker

ODAKAB:
My mistake, sorry.


Pablete1234 (commented):
This is incorrect, the `ish` variable on npcs is deprecated and should remain that way

ODAKAB:
> This is incorrect, the `ish` variable on npcs is deprecated and should remain that way

it doesn't update itself? i didn't pay attention if doesn't

PR #57, opened 2021-03-15
Attempt to fix zeta portal's recognizing target map.

Punikekk:


Pablete1234:
Thanks for the PR and showing the issue, it's been now fixed in c0e9d01

PR #61, opened 2021-03-19
Continue of API implementation.

Punikekk:


src/main/java/eu/darkbot/api/entities/other/Formation.java
```diff
@@ -70,12 +70,16 @@ public static Formation of(int formationId) {
...
     public static Formation of(String id) {
         for (Formation formation : values())
-            if (formation.id != null && id.endsWith(formation.id))
+            if (formation.endsWith(id))
                 return formation;
 
         return null;
     }
 
+    public boolean endsWith(String formationId) {
```
Pablete1234, line 79: I'd avoid this, it makes little sense, may make the impression that this is a string. if you want to name it `matches(String formationId)` or something else sure, but don't use `endsWith`, how it matches is an implementation detail

Pablete1234 (approved):

src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -219,7 +221,7 @@ private void tickLogic(boolean running) {
 
     private void checkRefresh() {
         if (config.MISCELLANEOUS.REFRESH_TIME == 0 ||
-                System.currentTimeMillis() - lastRefresh < config.MISCELLANEOUS.REFRESH_TIME * 60 * 1000) return;
+                System.currentTimeMillis() - lastRefresh < (long) config.MISCELLANEOUS.REFRESH_TIME * 60 * 1000) return;
```
Pablete1234, line 224:
```suggestion
                System.currentTimeMillis() - lastRefresh < config.MISCELLANEOUS.REFRESH_TIME * 60 * 1000L) return;
```

src/main/java/com/github/manolo8/darkbot/backpage/FlashResManager.java
```diff
@@ -45,6 +49,8 @@ public void tick() {
                 || currLang.equals(lang)) return;
 
         try {
+            inGameLocale = new Locale(currLang);
```
Pablete1234, line 52: wrap this in a try-catch with the specific exception.
From experience a `new Locale` with a random string (maybe a valid locale in another lang, but not in java) will throw an exception, catch it separately to the generic catch here

src/main/java/com/github/manolo8/darkbot/core/manager/GroupManager.java
```diff
@@ -218,4 +218,68 @@ public int idx(boolean leader) {
...
+        kick(id);
+    }
+
+    @Override
+    public List<? extends eu.darkbot.api.objects.group.GroupMember.Invite> getInvites() {
+        return invites;
+    }
+
+    @Override
+    public void openInvites() {
```
Pablete1234, line 277: This api method alongside a few others in group api make very little sense. This isn't "opening invites" this is just "enforcing the setting to set invites to open or closed", which is too reliant on the settings themselves.

Should consider removing this method from the api, have it be an implementation detail. 

Because of how "independent" the group manager works on its own (it reads settings and invites ppl with the tag, or leaves group, ETC) it maybe makes more sense for the API methods to be focused arround those settings more than acting on the group itself with "sync" methods, ie: maybe makes more sense to add a method that says "accept invite from X player whenever it appears" rather than a method to accept it directly that needs to be called in the precise moment that it is sent

src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
```diff
@@ -72,28 +79,43 @@
...
-        @SuppressWarnings("UnnecessaryLocalVariable")
-        Gui guiFix = gui; // Workaround for a java compiler assertion bug having issues with types
+    @SuppressWarnings({"unchecked", "CastCanBeRemovedNarrowingVariableType"})
+    public <T extends Gui> T register(String key, Class<T> gui) {
+        Gui guiFix = pluginAPI.requireInstance(gui);
         this.guis.addLazy(key, guiFix::update);
         this.registeredGuis.add(guiFix);
-        return gui;
+
+        // Workaround for a java compiler assertion bug having issues with types
```
Pablete1234, line 117: The compiler bug is not in the last line, it's about `gui::update` not working with a generic type, so we use `guiFix` that is "hard-coded" to be an instance of a Gui class just so that java compiler doesn't blow up. I don't think the comment should be on the last line.

src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
```diff
@@ -239,4 +261,61 @@ public boolean canTickModule() {
         return main.hero.locationInfo.isLoaded();
     }
 
+    private final RectangleImpl gameScreenRect = new RectangleImpl();
+
+    @Override
+    public Area.Rectangle getViewBounds() {
+        gameScreenRect.set(0, 0, MapManager.clientWidth, MapManager.clientHeight);
```
Pablete1234, line 268: ??
a rectangle that always starts at 0,0 makes no sense whatsoever. The view bounds should start on the top-left corner of your visible area, usually your location sits in the middle of this rectangle

src/main/java/com/github/manolo8/darkbot/core/manager/RepairManager.java
```diff
@@ -45,6 +54,10 @@ public void tick() {
 
         killerName = API.readMemoryString(API.readMemoryLong(repairAddress + 0x68));
         repairOptions.update(API.readMemoryLong(repairAddress + 0x58));
+
+        repairOptionsList.clear();
+        for (int i = 0; i < repairOptions.getSize(); i++)
+            this.repairOptionsList.add(repairOptions.get(i));
```
Pablete1234, line 60: should try to use `sync` in swf collection interface, it avoids having to re-make the whole list

src/main/java/com/github/manolo8/darkbot/core/objects/group/MemberInfo.java
```diff
@@ -56,4 +56,44 @@ public int getShield() {
     public int getMaxShield() {
         return maxShield;
     }
+
+    @Override
+    public boolean hpDecreasedIn(int time) {
+        return false;
```
Pablete1234, line 62: uhm...

src/main/java/eu/darkbot/impl/managers/AttackApiImpl.java
```diff
@@ -0,0 +1,116 @@
...
+        this.hero = hero;
+    }
+
+    @Override
+    public @Nullable Attackable getTarget() {
+        return target;
+    }
+
+    @Override
+    public @Nullable Attackable setTarget(@Nullable Attackable attackable) {
```
Pablete1234, line 36: not too sure about what the utility of previous target really is. If one place wants to use it, it can use getTarget before calling setTarget, i wouldn't return it every time and have to do the extra variable here

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/manager/RepairManager.java
Punikekk: have to implement it

Punikekk:
any ideas what to add/remove/change?

Pablete1234 (approved):

PR #62, opened 2021-03-20
Improve RSBing

ODAKAB:
1. Automated RSB cooldown detection.

2. Kept hardcoded time limits to prevent infinite spam if hotbar gets bugged.
With help of change number 1, I was able to add extra flexibility so now they have more chance to work in little laggy situations.

3. Attempted to do empowered bursts with Prometheus/Scatter Lasers, but commented them because it wasn't stable.
See TODOs in code.


ODAKAB:
What do you think about this idea:

Keep a variable lastShootedTime, update it to System.currentTimeMillis() whenever we lose effect 98.
If we are between lastShootedTime+4500 and lastShootedTime+5500, cease shooting until effect 98 is seen.

This should work fine without lag, can you tell me how can we stop attacking or help it.

ODAKAB:
https://github.com/darkbot-reloaded/DarkBot/pull/62/commits/668a1fcadc0aae5cd86d32e0f95a68b1fbf4e9da as requested

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -144,9 +144,12 @@ private boolean shouldSab() {
...
     private boolean shouldRsb() {
         if (!main.config.LOOT.RSB.ENABLED || !target.npcInfo.extra.has(NpcExtra.USE_RSB)) return false;
-        if (usedRsb < System.currentTimeMillis() - main.config.LOOT.RSB.AMMO_REFRESH) usedRsb = System.currentTimeMillis();
+        boolean isReady = bar.findItemById("ammunition_laser_rsb-75").map(i -> i.activatable).orElse(false);
 
-        return usedRsb > System.currentTimeMillis() - 50;
+        if (isReady && usedRsb < System.currentTimeMillis() - 2000) usedRsb = System.currentTimeMillis();
+                // && (!main.config.LOOT.RSB.EMPOWERED_BURST || hero.hasEffect(98)) TODO: Find a way to determine lastShootedTime, doesn't work stable if you don't guess it before effect.
+
+        return isReady && usedRsb > System.currentTimeMillis() - 100;
```
pajtor, line 151:
```suggestion
 if (isReady && usedRsb < (System.currentTimeMillis() - 2000)) usedRsb = System.currentTimeMillis();
```

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -144,9 +144,12 @@ private boolean shouldSab() {
 
     private boolean shouldRsb() {
         if (!main.config.LOOT.RSB.ENABLED || !target.npcInfo.extra.has(NpcExtra.USE_RSB)) return false;
-        if (usedRsb < System.currentTimeMillis() - main.config.LOOT.RSB.AMMO_REFRESH) usedRsb = System.currentTimeMillis();
+        boolean isReady = bar.findItemById("ammunition_laser_rsb-75").map(i -> i.activatable).orElse(false);
 
-        return usedRsb > System.currentTimeMillis() - 50;
+        if (isReady && usedRsb < System.currentTimeMillis() - 2000) usedRsb = System.currentTimeMillis();
+                // && (!main.config.LOOT.RSB.EMPOWERED_BURST || hero.hasEffect(98)) TODO: Find a way to determine lastShootedTime, doesn't work stable if you don't guess it before effect.
```
pajtor:
```suggestion
 if (isReady && usedRsb < System.currentTimeMillis() - 2000) usedRsb = System.currentTimeMillis();
```

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -144,9 +144,12 @@ private boolean shouldSab() {
 
     private boolean shouldRsb() {
         if (!main.config.LOOT.RSB.ENABLED || !target.npcInfo.extra.has(NpcExtra.USE_RSB)) return false;
-        if (usedRsb < System.currentTimeMillis() - main.config.LOOT.RSB.AMMO_REFRESH) usedRsb = System.currentTimeMillis();
+        boolean isReady = bar.findItemById("ammunition_laser_rsb-75").map(i -> i.activatable).orElse(false);
 
-        return usedRsb > System.currentTimeMillis() - 50;
+        if (isReady && usedRsb < System.currentTimeMillis() - 2000) usedRsb = System.currentTimeMillis();
```
pajtor, line 149:
```suggestion
 if (isReady && usedRsb < (System.currentTimeMillis() - 2000)) usedRsb = System.currentTimeMillis();
```

Pablete1234: This is uneeded due to preference of operators. There can be no confusion as you can't compute a boolean and then do `bool - long`.

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -144,9 +144,11 @@ private boolean shouldSab() {
...
+
+        return isReady && usedRsb > System.currentTimeMillis() - 100;
```
Pablete1234, line 151:
```suggestion
        if (isReady && usedRsb < System.currentTimeMillis() - 1000) usedRsb = System.currentTimeMillis();
        return usedRsb > System.currentTimeMillis() - 50;
```
Let's keep the min-selection timeout the same (50ms) as it seemed to work right, and let's lower the min delay of activation here. There's no reason to keep it that high when it will be activated almost instantly

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -144,9 +144,11 @@ private boolean shouldSab() {
 
     private boolean shouldRsb() {
         if (!main.config.LOOT.RSB.ENABLED || !target.npcInfo.extra.has(NpcExtra.USE_RSB)) return false;
-        if (usedRsb < System.currentTimeMillis() - main.config.LOOT.RSB.AMMO_REFRESH) usedRsb = System.currentTimeMillis();
+        boolean isReady = bar.findItemById("ammunition_laser_rsb-75").map(i -> i.activatable).orElse(false);
```
Pablete1234, line 147: This breaks support for any other burst-based ammo (eg: the x7 ammo), should keep that into consideration and search for the item that is actually being selected by the key

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
ODAKAB: Wrong, all rapid salvo batteries share same cooldown, like regular how rockets do.

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SlotBarsProxy.java
```diff
@@ -12,9 +13,22 @@
     //public SlotBar premiumBar    = new SlotBar(); //104
     //public SlotBar proActionBar  = new SlotBar(); //112
 
+    private Main main;
```
Pablete1234, line 16: Did you even test this change? this main is never initialized.

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SlotBarsProxy.java
```diff
@@ -12,9 +13,22 @@
...
+
     @Override
     public void update() {
         this.categoryBar.update(API.readMemoryLong(address + 88));
         this.standardBar.update(API.readMemoryLong(address + 96));
     }
+
+    public String getItemIdByKey(Character c) {
+        SettingsProxy.KeyBind keybind = main.facadeManager.settings.getAtChar(c);
```
Pablete1234, line 25: This will throw a NPE.

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SlotBarsProxy.java
ODAKAB: Tested, and failed as i wrote in commit name. Can you tell me how should i initialize or what to do?

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SlotBarsProxy.java
ODAKAB: What to do? I used `if (keybind == null) return null;` below

Pablete1234: read the comment above. `main` will be null

ODAKAB: How can i initialize, i have no idea about that. inside update() ?

ODAKAB:
Working on a different approach

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -143,9 +145,17 @@ private boolean shouldSab() {
...
-
+        if (!main.config.LOOT.RSB.ENABLED || main.config.LOOT.RSB.KEY == null
+                || !target.npcInfo.extra.has(NpcExtra.USE_RSB)) return false;
+        String itemId = main.facadeManager.slotBars.standardBar.slots.stream()
+                .filter(slot -> Objects.equals(main.facadeManager.settings.getCharCode(valueOf
+                        ("SLOTBAR_" + (slot.slotNumber == 10 ? 0 : slot.slotNumber))), main.config.LOOT.RSB.KEY))
+                .findFirst().map(slot -> slot.item != null ? slot.item.id : null).orElse(null);
+        if (!Objects.equals(itemId, "ammunition_laser_rsb-75")
+                && !Objects.equals(itemId, "ammunition_laser_rcb-140")) return false; // TODO: Check bool cooldown from enum, after API gets merged.
+        boolean isReady = bar.findItemById(itemId).map(i -> i.activatable).orElse(false);
```
Pablete1234, line 156: Add this method to SettingsProxy:
```java
@Nullable
public KeyBind getKeyBind(Character ch) {
    if (ch == null) return null;
    for (int i = 0; i < keycodes.length; i++) {
        if (keycodes[i] == ch) return KeyBind.values()[i];
    }
    return null;
}
```

Add this method to SlotBarsProxy (as well as uncommenting premiumBar):
```java
@Nullable
public Slot getSlot(KeyBind key) {
        if (key == null) return null;
        String keyStr = keybind.toString();
        
        SlotBar sb = keyStr.startsWith("SLOTBAR") ? standardBar :
                keyStr.startsWith("PREMIUM") ? premiumBar : null;
        
        return sb == null ? null : sb.slots.get(Integer.parseInt(keyStr.split("_")[1]));
}
```

Then this code can be simplified to just:
```suggestion
        KeyBind key = settings.getKeyBind(main.config.LOOT.RSB.KEY);
        Slot slot = slotBars.get(key);
        if (slot == null || slot.item == null) return false;
        boolean isReady = bar.findItemById(slot.item.id).map(i -> i.activatable).orElse(false);
```

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -143,9 +145,17 @@ private boolean shouldSab() {
...
     private boolean shouldRsb() {
-        if (!main.config.LOOT.RSB.ENABLED || !target.npcInfo.extra.has(NpcExtra.USE_RSB)) return false;
-        if (usedRsb < System.currentTimeMillis() - main.config.LOOT.RSB.AMMO_REFRESH) usedRsb = System.currentTimeMillis();
-
+        if (!main.config.LOOT.RSB.ENABLED || main.config.LOOT.RSB.KEY == null
+                || !target.npcInfo.extra.has(NpcExtra.USE_RSB)) return false;
+        String itemId = main.facadeManager.slotBars.standardBar.slots.stream()
+                .filter(slot -> Objects.equals(main.facadeManager.settings.getCharCode(valueOf
+                        ("SLOTBAR_" + (slot.slotNumber == 10 ? 0 : slot.slotNumber))), main.config.LOOT.RSB.KEY))
+                .findFirst().map(slot -> slot.item != null ? slot.item.id : null).orElse(null);
```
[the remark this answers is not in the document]
ODAKAB: it was warning me when i try to do `.map(slot -> slot.item.id)` seperatng it to two maps solves that issue, or it was never an issue?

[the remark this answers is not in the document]
Pablete1234: Because you're using an `Optional`, doing `.map(a -> null)` will not be any exception, it will simply become an `empty` optional and when you later do another `.map` it won't throw an exception, it will simply continue to be an `empty` optional.

however if you try to `.map(a -> null.something)` it will throw NPE, as you tried to get some inner variable of a null reference

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -143,9 +145,17 @@ private boolean shouldSab() {
...
+        if (!Objects.equals(itemId, "ammunition_laser_rsb-75")
+                && !Objects.equals(itemId, "ammunition_laser_rcb-140")) return false; // TODO: Check bool cooldown from enum, after API gets merged.
```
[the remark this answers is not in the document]
ODAKAB: You wanted to restrict it because you told me you want it to not work if user sets a regular ammo like x1
And no, timer of item isn't distinguishable when it's not in cooldown. We can't automatically know if it's a cooldowned item or not.

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -143,9 +145,17 @@ private boolean shouldSab() {
     }
 
     private boolean shouldRsb() {
-        if (!main.config.LOOT.RSB.ENABLED || !target.npcInfo.extra.has(NpcExtra.USE_RSB)) return false;
-        if (usedRsb < System.currentTimeMillis() - main.config.LOOT.RSB.AMMO_REFRESH) usedRsb = System.currentTimeMillis();
-
+        if (!main.config.LOOT.RSB.ENABLED || main.config.LOOT.RSB.KEY == null
+                || !target.npcInfo.extra.has(NpcExtra.USE_RSB)) return false;
+        String itemId = main.facadeManager.slotBars.standardBar.slots.stream()
```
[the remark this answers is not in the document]
ODAKAB: Previous approach which i abandoned because couldn't construct main was having it ready in SlotbarsProxy. Future-ready with ordinal check. https://github.com/ODAKAB/DarkBot/commit/da4319845ac670015feb93a51811d5e02b731f2f

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -143,9 +145,17 @@ private boolean shouldSab() {
...
+                .filter(slot -> Objects.equals(main.facadeManager.settings.getCharCode(valueOf
+                        ("SLOTBAR_" + (slot.slotNumber == 10 ? 0 : slot.slotNumber))), main.config.LOOT.RSB.KEY))
```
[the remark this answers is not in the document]
ODAKAB: I don't understand how can i reduce complexity of this. Help needed.

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
ODAKAB: Will do, after you decide that you want this approach or https://github.com/ODAKAB/DarkBot/commit/da4319845ac670015feb93a51811d5e02b731f2f fixed.

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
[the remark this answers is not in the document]
Pablete1234: I never said i want it not to work. I said if something is wrong it should be apparent to the user. Making it directly never trigger is not apparent to the user at all to start with, it's simply disabling the feature

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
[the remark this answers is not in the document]
Pablete1234: for each slot in the slotbar you're searching for the key code and then comparing, simply look at my proposed solution below where there's just one after the other, instead of one inside the other

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
Pablete1234: I think it's best for it to be toggling on & off the ammo if you happen to pick an RSB ammo that isn't rsb, and it to work fine if you happen to put another random ability in there

ODAKAB: I mean
```
public String getItemIdByKey(Character c) {
        SettingsProxy.KeyBind keybind = main.facadeManager.settings.getAtChar(c);
        if (keybind == null) return null;

        if (keybind.ordinal() <= 9) {
            return standardBar.slots.stream().filter(s -> s.slotNumber == keybind.ordinal() + 1)
                    .filter(s -> s.item != null).map(s -> s.item.id).findFirst().orElse(null);
        }
        else return null;
    }
```
in SlotbarsProxy of old commit

Pablete1234: simply use the methods i provided, this one doesn't handle premium bar

ODAKAB: Okay I'll start applying today. But this was going to handle premium bar and anything in future just with keybind.ordinal value.

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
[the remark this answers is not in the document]
ODAKAB: With 1000ms It will be pretty annoying and cause spam around support channels. If it's RSB, it must be it. Would be okay if we could restrict it to items which have cooldowns.

ODAKAB:
Applied everything you want, removed wrong item prevention, tested and works.

Pablete1234 (approved):

PR #75, opened 2021-04-16
Added formation check by slot

Punikekk:


src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -140,18 +141,35 @@ public boolean setMode(Config.ShipConfig config) {
         return setMode(config.CONFIG, config.FORMATION);
     }
 
+    private boolean formationCheckBySlot = true;
```
Pablete1234, line 144: put this above alongside the rest of the state for hero manager

also rename it to something that makes more sense, i'm not sure if this is `shouldCheckFormationBySlot` or `didCheckBySlot` or what

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -140,18 +141,35 @@ public boolean setMode(Config.ShipConfig config) {
...
 
         if ((this.formation != form || checkFormation) && System.currentTimeMillis() - formationTime > 3500L) {
             Main.API.keyboardClick(this.formation = form);
-            if (formation != null) this.formationTime = System.currentTimeMillis();
+            if (formation != null) {
+                this.formationTime = System.currentTimeMillis();
+                this.formationCheckBySlot = false;
+            }
+
+        } else if (!formationCheckBySlot && System.currentTimeMillis() - formationTime > 600L) {
```
Pablete1234, line 161: 600 ms is way too little, we should never be clicking to switch formations constantly so quickly

also this will only trigger if it's been in the 3500 ms  timeframe since last formation? does this ever trigger?

Pablete1234 (changes requested):

PR #76, opened 2021-04-28
add args -config for set a custom config at startup

fabio1999ita:
ex: -configs palladium

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/Main.java
+++ b/src/main/java/com/github/manolo8/darkbot/Main.java
@@ -51,7 +51,7 @@ import java.util.stream.Stream;
 
 public class Main extends Thread implements PluginListener {
 
-    public static final Version VERSION      = new Version("1.13.17 beta 87");
+    public static final Version VERSION      = new Version("1.13.17 beta 92");
     public static final Object UPDATE_LOCKER = new Object();
     public static final Gson GSON            = new GsonBuilder()
             .setPrettyPrinting()
@@ -316,11 +316,11 @@ public class Main extends Thread implements PluginListener {
         try {
             SwingUtilities.invokeAndWait(() -> {
                 this.config = configManager.loadConfig(config);
-                mapManager.updateAreas();
+                mapManager.updateAreas(true);
                 pluginHandler.updateConfig(); // Get plugins to update what features are enabled
                 featureRegistry.updateConfig(); // Update the features & configurables
                 form.updateConfiguration(); // Rebuild config gui
-                setModule(module, true);
+                setModule(null, true);
             });
         } catch (InterruptedException | InvocationTargetException e) {
             e.printStackTrace();
--- a/src/main/java/com/github/manolo8/darkbot/Main.java
+++ b/src/main/java/com/github/manolo8/darkbot/Main.java
@@ -101,10 +101,10 @@ public class Main extends Thread implements PluginListener {
         super("Main");
         VerifierChecker.getAuthApi().setupAuth();
 
-        if (params.startConfigs() == null)
+        if (params.startConfig() == null)
             configManager.loadConfig(null);
         else
-            config = configManager.loadConfig(params.startConfigs());
+            config = configManager.loadConfig(params.startConfig());
 
         API = configManager.getAPI(params);
         API.setSize(config.BOT_SETTINGS.API_CONFIG.width, config.BOT_SETTINGS.API_CONFIG.height);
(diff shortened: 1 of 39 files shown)


src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -121,7 +123,9 @@ public Main(StartupParams params) {
 
         API.createWindow();
         if (params.getAutoStart()) setRunning(true);
+
```
Pablete1234, line 126:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -121,7 +123,9 @@ public Main(StartupParams params) {
...
         start();
+
```
Pablete1234, line 128:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -113,6 +113,8 @@ public Main(StartupParams params) {
         this.pluginHandler.updatePluginsSync();
         this.pluginHandler.addListener(this);
 
+        if (params.startConfigs() != null) config = configManager.loadConfig(params.startConfigs());
```
Pablete1234, line 116: By doing this here, you're loading the config twice, things like the API are comming from the main config instead of the new one, same for the size of the window.
Make line 66 just `Config config;` 
Initialize the config (calling loadConfig) with either null or the other start config parameter on line 103 (after setupAuth, but before configManager.getAPI)

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -41,6 +41,13 @@
     private static final String NO_OP_COMMAND = COMMAND_PREFIX + "no-op";
     private boolean forceNoOp = false;
 
+    /**
+     * Command-line argument for set start-config, has no parameters
+     * Used for debugging
+     */
+    private static final String START_CONFIGS = COMMAND_PREFIX + "configs";
+    private String startConfigs = null;
```
Pablete1234, line 49:
```suggestion
    /**
     * Command-line argument for set start-config, has no parameters
     * Used for debugging
     */
    private static final String START_CONFIG = COMMAND_PREFIX + "config";
    private String startConfig = null;
```
Config, it's a single one, not plural. Also fix the copy-pasted comment

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -62,6 +69,12 @@ public StartupParams(String[] args) throws IOException {
                 case NO_OP_COMMAND:
                     forceNoOp = true;
                     break;
+                case START_CONFIGS:
+                    if (i + 1 >= args.length)
+                        System.err.println("Missing arguments for config, usage: -configs nameofconfig");
```
Pablete1234, line 74:
```suggestion
                        System.err.println("Missing arguments for config, usage: -config configname");
```

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -106,6 +119,10 @@ public boolean useNoOp() {
         return forceNoOp;
     }
 
+    public String startConfigs(){
+        return startConfigs;
```
Pablete1234, line 123:
```suggestion
    public String startConfig() {
        return startConfig;
```

Pablete1234 (changes requested):

fabio1999ita:
should be ok now, sorry for spam of commit...

src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -100,6 +100,12 @@
     public Main(StartupParams params) {
         super("Main");
         VerifierChecker.getAuthApi().setupAuth();
+
+        if (params.startConfig() == null)
+            configManager.loadConfig(null);
+        else
+            config = configManager.loadConfig(params.startConfig());
```
Pablete1234, line 107: This expression can be simplified to
```suggestion
        config = configManager.loadConfig(params.startConfig());
```

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -106,6 +119,10 @@ public boolean useNoOp() {
         return forceNoOp;
     }
 
+    public String startConfig(){
```
Pablete1234, line 122:
```suggestion
    public String startConfig() {
```

Pablete1234 (changes requested):

PR #77, opened 2021-05-26
added basic captcha api support

anjeshshrestha:


src/main/java/com/github/manolo8/darkbot/utils/CaptchaAPI.java
```diff
@@ -0,0 +1,16 @@
...
+    static CaptchaAPI getInstance(){
+        return INSTANCE;
+    }
+
+    /**
+     * Pass a website page and it will parse the g-site-key and solve using external service.
+     * @param webpage a page that has been consumed into a string
+     * @return A solved g-captcha-response key from 2Captcha Solver Service
+     */
+    String solveCaptcha(String webpage);
```
Pablete1234, line 15:
```suggestion
     * Create solving captcha parameters for the provided webpage html string.
     * @param the url this webpage was obtained from
     * @param webpage as an HTML string
     * @return form parameters to include as captcha response solution
     */
    Map<String, String> solveCaptcha(URL url, String webpage);
```
Passing in the URL allows you to send it over to the resolver service, which is needed, instead of hard-coding the url in the implementation

For the response map, this implementation can just be, a:
`Collections.singletonMap("g-captcha-response", response);`

However other types of captchas may use different responses (or more of them)

src/main/java/com/github/manolo8/darkbot/utils/CaptchaAPI.java
```diff
@@ -0,0 +1,16 @@
+package com.github.manolo8.darkbot.utils;
+
+public interface CaptchaAPI {
+    String TWOCAPTCHA_PATH = "lib/captchasolver.jar";
+    CaptchaAPI INSTANCE = ReflectionUtils.createInstance("eu.darkbot.captcha.CaptchaSolver", TWOCAPTCHA_PATH);
+    static CaptchaAPI getInstance(){
+        return INSTANCE;
+    }
```
Pablete1234, line 8:
```suggestion
    String SOLVER_PATH = "lib/captchasolver.jar";
    CaptchaAPI INSTANCE;

    static {
        try {
            INSTANCE = ReflectionUtils.createInstance("eu.darkbot.captcha.CaptchaSolver", SOLVER_PATH);
        } catch (Exception e) {
            System.out.printLn("No captcha resolver is configured, if you're not trying to use one you can safely ignore this message");
        }
    }

    static CaptchaAPI getInstance() {
        return INSTANCE;
    }
```

Make sure you handle the exception, as it's not a required class

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -118,17 +117,37 @@ public static void usernameLogin(LoginData loginData) {
...
     public static void usernameLogin(LoginData loginData, String domain) {
-        String loginUrl = Http.create("https://" + domain + ".darkorbit.com/")
-                .consumeInputStream(LoginUtils::getLoginUrl);
+        String frontPage = "";
+        String gResponse = "";
+        try {
+            frontPage = IOUtils.read(Http.create("https://" + domain + ".darkorbit.com/").getInputStream());
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
```
Pablete1234, line 126:
```suggestion
        String frontPage = IOUtils.read(Http.create("https://" + domain + ".darkorbit.com/").getInputStream());
```

If this is the exception handling you're gonna do (none) better not do it at all, just bubble the exception up instead.

Make this method throw IOException, let the higher-up usernameLogin method handle the exception (by trying a diff domain).

Also make this method private if possible, it's an implentation detail

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -118,17 +117,37 @@ public static void usernameLogin(LoginData loginData) {
     }
 
     public static void usernameLogin(LoginData loginData, String domain) {
```
Pablete1234, line 119:
```suggestion
    private static void usernameLogin(LoginData loginData, String domain) throws IOException {
```

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -118,17 +117,37 @@ public static void usernameLogin(LoginData loginData) {
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
 
+        try {
+            CaptchaAPI captcha = CaptchaAPI.getInstance();
+            gResponse = captcha.solveCaptcha(frontPage);
+        }catch (Exception e){
+            e.printStackTrace();
+        }
```
Pablete1234, line 133:
```suggestion
    Map<String, String> extraPostParams = Collections.emptyMap();
    CaptchaAPI solver = CaptchaAPI.getInstance();
    if (solver != null) {
        try {
            extraPostParams = solver.solveCaptcha(frontPage);
        } catch (Exception e) {
            System.out.printLn("Captcha solver failed to resolve login captcha");
            e.printStackTrace();
        }
    }
```
Only run if the captcha solver exists, and make sure to say a bit more in console

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -167,6 +186,16 @@ private static String replaceParameters(String params) {
...
     }
 
+    private static String getLoginUrl(String in) {
+        Matcher match = LOGIN_PATTERN.matcher(in);
+        if (match.find()){
+            String login_pattern = match.group(1).replace("&amp;", "&");
+            System.out.println(login_pattern);
+            return login_pattern;
+        }
+        throw new WrongCredentialsException();
```
Pablete1234, line 196: i doubt credentials are wrong if you didn't try to log-in yet, find a better exception or just return a null and handle it higher up

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -167,6 +186,16 @@ private static String replaceParameters(String params) {
         return params;
     }
 
+    private static String getLoginUrl(String in) {
+        Matcher match = LOGIN_PATTERN.matcher(in);
+        if (match.find()){
+            String login_pattern = match.group(1).replace("&amp;", "&");
+            System.out.println(login_pattern);
+            return login_pattern;
+        }
```
Pablete1234, line 195:
```suggestion
        Matcher match = LOGIN_PATTERN.matcher(in);
        if (match.find()) return match.group(1).replace("&amp;", "&");
```

never use underscores in local variable names, also no need for this print

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -118,17 +117,37 @@ public static void usernameLogin(LoginData loginData) {
...
+                Http.create(loginUrl, Method.POST)
+                        .setParam("username", loginData.getUsername())
+                        .setParam("password", loginData.getPassword())
+                        .closeInputStream();
+            else
+                Http.create(loginUrl, Method.POST)
+                        .setParam("username", loginData.getUsername())
+                        .setParam("password", loginData.getPassword())
+                        .setParam("g-recaptcha-response", gResponse)
+                        .closeInputStream();
```
Pablete1234, line 150:
```suggestion
            Http http = Http.create(loginUrl, Method.POST)
                        .setParam("username", loginData.getUsername())
                        .setParam("password", loginData.getPassword());
            extraPostParams.forEach(http::setParam);
            http.closeInputStream();
```

Simply add whatever extra params exist directly, it could be 0, or many, not relevant to our implementation

Pablete1234 (changes requested):

anjeshshrestha (commented):
updated in the last 2 commits

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -167,6 +185,20 @@ private static String replaceParameters(String params) {
         return params;
     }
 
+    private static String getLoginUrl(String in) {
+        Matcher match = LOGIN_PATTERN.matcher(in);
+        if (match.find()) return match.group(1).replace("&amp;", "&");
+        //find better exception
+        throw new WrongCredentialsException();
```
Pablete1234, line 192: Maybe just return a null and handle the exception higher up

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -112,23 +113,40 @@ private static String getPassword(String username, char[] masterPassword) {
...
+        if (solver != null) {
+            try {
+                extraPostParams = solver.solveCaptcha(url, frontPage);
+            } catch (Exception e) {
+                System.out.println("Captcha solver failed to resolve login captcha");
+                e.printStackTrace();
+            }
+        }
 
+        String loginUrl = getLoginUrl(frontPage);
```
Pablete1234, line 140: after this, handle if loginURL == null throw a new exception
Consider making a LoginException that WrongCredentialsException extends, so that we can throw it arround here (and also make sure it's cached in the login form

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -167,6 +190,20 @@ private static String replaceParameters(String params) {
...
+        throw new LoginURLException();
+    }
+
+    /**
+     * @Deprecated
+     * This method is no longer used to get login url
+     * @param in http connection to url
+     * @return login url parsed
+     */
+
```
Pablete1234, line 206: Just remove it, it's private and unused right?

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -210,4 +247,13 @@ public WrongCredentialsException(String s) {
         }
     }
 
+    public static class LoginURLException extends  WrongCredentialsException {
```
Pablete1234, line 250: What i ment was the oposite:

have a `LoginException` that is the more generic one
then have `WrongCredentialsException extends LoginException`

Then make the code catching WrongCredentialsException catch the more generic LoginException, although i think that code already catches Exception so it should be fine

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -113,22 +114,44 @@ public static void usernameLogin(LoginData loginData) {
...
-                    .closeInputStream();
+            Http http = Http.create(loginUrl, Method.POST)
+                        .setParam("username", loginData.getUsername())
+                        .setParam("password", loginData.getPassword());
+            extraPostParams.forEach(http::setParam);
+            if(http.getContent().contains("bgcdw_errors_all_wrapper")){
+                System.out.println("Wrong Credentials");
+                throw new WrongCredentialsException();
+            }
+
```
Pablete1234, line 154:
```suggestion
            if (http.getContent().contains("bgcdw_errors_all_wrapper"))
                throw new WrongCredentialsException();
```

Although i think this is simply handled lower down by the fact that no DOSID cookie will be found, so i'm not sure that this makes much of a difference

Pablete1234 (changes requested):
Almost there, just a couple final touches

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
anjeshshrestha: yup you are right i didnt see it was being handled

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -199,7 +218,17 @@ public static void saveCredentials(Credentials c, char[] passwd) throws GeneralS
...
+        public LoginException(){
+            this("Login Exception");
+        }
+
+        public LoginException(String s) {
+            super(s);
+        }
+    }
+
+    public static class WrongCredentialsException extends  LoginException{
```
Pablete1234, line 231:
```suggestion
    public static class LoginException extends RuntimeException {
        public LoginException(String s) {
            super(s);
        }
    }

    public static class WrongCredentialsException extends LoginException {
```

Didn't notice wrong credentials was extending illegal arg ex, probably shouldn't it's a good time to change that

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -167,12 +187,11 @@ private static String replaceParameters(String params) {
...
-        return new BufferedReader(new InputStreamReader(in)).lines()
-                .map(LOGIN_PATTERN::matcher)
-                .filter(Matcher::find)
-                .map(matcher -> matcher.group(1).replace("&amp;", "&"))
-                .findFirst().orElseThrow(WrongCredentialsException::new);
+    private static String getLoginUrl(String in) {
+        Matcher match = LOGIN_PATTERN.matcher(in);
+        if (match.find()) return match.group(1).replace("&amp;", "&");
+
+        throw new LoginException();
```
Pablete1234, line 194:
```suggestion
        throw new LoginException("Failed to get login URL in frontpage");
```
Make exceptions descriptive when possible

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -113,22 +114,41 @@ public static void usernameLogin(LoginData loginData) {
         try {
             usernameLogin(loginData, "www");
         } catch (Exception e) {
-            usernameLogin(loginData, "lp");
+            try {
+                usernameLogin(loginData, "lp");
+            } catch (IOException ex) {
+                throw new RuntimeException(ex);
```
Pablete1234, line 120:
```suggestion
                throw new LoginException(ex);
```
Let's reuse our exception properly (also keep in mind you'll need to add a constructor with a cause exception to pass to super(ex) )

Pablete1234 (changes requested):
Hopefully final fixes

Pablete1234 (approved):
Good job on implementing the feature, i'll merge once i get confirmation that this works as intended

PR #78, opened 2021-05-27
add hide argument

fabio1999ita:
add hide argument to startup of bot for auto hide API

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/Main.java
+++ b/src/main/java/com/github/manolo8/darkbot/Main.java
@@ -144,19 +144,6 @@ public class Main extends Thread implements PluginListener {
         API.createWindow();
         if (params.getAutoStart()) setRunning(true);
         start();
-        if (params.getAutoHide()) {
-            Thread hide = new Thread(() -> {
-                while (true) {
-                    Time.sleep(2000);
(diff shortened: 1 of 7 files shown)


src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -144,6 +144,19 @@ public Main(StartupParams params) {
         API.createWindow();
         if (params.getAutoStart()) setRunning(true);
         start();
+        if (params.getAutoHide()) {
+            Thread hide = new Thread(() -> {
```
Pablete1234, line 148: Implement this in the API adapter not in main, consider that it may be better to simply modify darkboat api to handle this rather than having a hacky 2nd thread that will keep on waiting until it's valid to set it to minimized, but regardless of that you can likely have the isValid method in DarkboatApiAdapter make the check and after returning true sending a setMinimized and modifying an internal state.

Keep in mind that `setMinimized` does NOT work on all systems and there's extra logic required to use either setMinimized or setVisible depending on a user setting for fully hide API, that this must respect

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #79, opened 2021-05-29
Refresh if jumping portal takes too long

ODAKAB:
not tested, review with caution i was a bit sleepy

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
+++ b/src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
@@ -6,12 +6,15 @@ import com.github.manolo8.darkbot.core.manager.HeroManager;
 import com.github.manolo8.darkbot.core.utils.Drive;
 import com.github.manolo8.darkbot.core.utils.Location;
 
+import static com.github.manolo8.darkbot.Main.API;
+
 public class PortalJumper {
 
     private final HeroManager hero;
     private final Drive drive;
     private Portal last;
     private long nextMoveClick;
+    private long tryingToJumpSince;
 
     public PortalJumper(HeroManager hero) {
         this.hero = hero;
@@ -44,15 +47,22 @@ public class PortalJumper {
             if (!gm.group.isValid() || gm.group.size < minGroupSize) return;
         }
 
-        hero.jumpPortal(target);
-
         if (target != last) {
             last = target;
+            tryingToJumpSince = System.currentTimeMillis();
             nextMoveClick = System.currentTimeMillis() + 5000;
         } else if (System.currentTimeMillis() > nextMoveClick && !target.clickable.enabled) {
             hero.drive.clickCenter(true, target.locationInfo.now);
             nextMoveClick = System.currentTimeMillis() + 10000;
         }
+
+        if (tryingToJumpSince != 0 && System.currentTimeMillis() > tryingToJumpSince + 120000) {
+            System.out.println("Triggering refresh: jumping portal took too long");
+            tryingToJumpSince = 0;
+            API.handleRefresh();
+        }
+
+        hero.jumpPortal(target);
     }
 
     public void travelAndJump(Portal target) {
--- a/src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
+++ b/src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
@@ -15,6 +15,7 @@ public class PortalJumper {
     private Portal last;
     private long nextMoveClick;
     private long tryingToJumpSince;
+    private long lastJumpTick;
 
     public PortalJumper(HeroManager hero) {
         this.hero = hero;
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
```diff
@@ -44,15 +47,22 @@ public void jump(Portal target) {
             if (!gm.group.isValid() || gm.group.size < minGroupSize) return;
         }
 
-        hero.jumpPortal(target);
```
Pablete1234, line 47: Try to keep this up here, if you move it down, jumping occurs after clicking which can lead to more desyncs

src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
```diff
@@ -44,15 +47,22 @@ public void jump(Portal target) {
...
         if (target != last) {
             last = target;
+            tryingToJumpSince = System.currentTimeMillis();
             nextMoveClick = System.currentTimeMillis() + 5000;
         } else if (System.currentTimeMillis() > nextMoveClick && !target.clickable.enabled) {
             hero.drive.clickCenter(true, target.locationInfo.now);
             nextMoveClick = System.currentTimeMillis() + 10000;
         }
+
+        if (tryingToJumpSince != 0 && System.currentTimeMillis() > tryingToJumpSince + 120000) {
```
Pablete1234, line 59: You probably want this as an else-if from the above, so if the desync is known (clickable isn't enabled, so you need to still travel towards it) it won't be trying to refresh

Pablete1234 (changes requested):
Looks like a fine start but it requires a couple changes

Consider that you are not resetting tryingToJumpSince, so in the event that you start trying to jump a portal and then decide you want to instead continue on the same map, and a couple minutes later you try to jump that portal again it will instantly trigger a refresh since tryingToJump will only be reset if target portal changes in this portal jumper, you need to handle that case better

src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
ODAKAB: there isn't a synchronous block to cause that, 1ms doesn't cause a desync

src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
ODAKAB: clickable is always enabled on desync since it happens after jump(probably) check both screenshots in bug report

ODAKAB:
> Looks like a fine start but it requires a couple changes
> 
> Consider that you are not resetting tryingToJumpSince, so in the event that you start trying to jump a portal and then decide you want to instead continue on the same map, and a couple minutes later you try to jump that portal again it will instantly trigger a refresh since tryingToJump will only be reset if target portal changes in this portal jumper, you need to handle that case better

I don't know how would you like to have this, is it possible to make it static and so MapTraveler's and SafetyFinder's PortalJumper tryingToJumpSince variable will be same one? Then we can reset it on pausing bot.

src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
Pablete1234: I'm aware of the specific bug you're trying to solve, my point is precisely that because clickable IS enabled, you should only do this on after an `else` from above instead of always

src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
Pablete1234: you can't make assumptions about synchronism on the call, they're undefined behavior and api-dependant, however, it'd be better if they were sync in the future which is why code should make a best effort for that behavior

Pablete1234:
> > Looks like a fine start but it requires a couple changes
> > Consider that you are not resetting tryingToJumpSince, so in the event that you start trying to jump a portal and then decide you want to instead continue on the same map, and a couple minutes later you try to jump that portal again it will instantly trigger a refresh since tryingToJump will only be reset if target portal changes in this portal jumper, you need to handle that case better
> 
> I don't know how would you like to have this, is it possible to make it static and so MapTraveler's and SafetyFinder's PortalJumper tryingToJumpSince variable will be same one? Then we can reset it on pausing bot.

The best way i can think of is to keep track of how often you're being called, if the last time `jump` was called is over X time, reset the state of the jumper.

The SafetyManager does something similar: https://github.com/darkbot-reloaded/Darkbot/blob/master/src/main/java/com/github/manolo8/darkbot/modules/utils/SafetyFinder.java#L114

src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
ODAKAB: merged to else if chain

src/main/java/com/github/manolo8/darkbot/modules/utils/PortalJumper.java
ODAKAB: reverted to original position

ODAKAB:
I'm not sure for which parts and how you would like to use reset method, so i didn't involve it.

Pablete1234 (approved):

PR #80, opened 2021-05-31
Reworked HeroItemsAPI, fixed some bugs, added FutureResult

Punikekk:


Punikekk:
you can review if you have some time

src/main/java/eu/darkbot/api/future/FutureResult.java
```diff
@@ -0,0 +1,60 @@
...
+    Optional<R> getResult();
+
+    /**
+     * @return current {@link Status} of this {@link FutureResult}
+     */
+    Status getStatus();
+
+    /**
+     * May return false if current status is already {@link Status#COMMITTED}
+     * and mayInterrupt parameter is set to false.
```
Pablete1234, line 20: ok sure, but what does it do?

Does it have side-effects?
Does it mark task as completed?
Will it result in onDone being called?

The javadocs on this class are a bit lacking for how important of a class it is, document all the states in Status rigurously

src/main/java/eu/darkbot/api/future/ItemFutureResult.java
```diff
@@ -0,0 +1,42 @@
+package eu.darkbot.api.future;
+
+import eu.darkbot.api.items.ItemUseResult;
+import eu.darkbot.api.items.Item;
+
+import java.util.Optional;
+import java.util.function.Consumer;
+
+public interface ItemFutureResult extends FutureResult<ItemUseResult> {
```
Pablete1234, line 9: just delete this file

src/main/java/eu/darkbot/api/items/ItemUseFlag.java
```diff
@@ -0,0 +1,25 @@
+package eu.darkbot.api.items;
+
+import java.util.function.Predicate;
+
+public enum ItemUseFlag {
```
Pablete1234, line 5: bit of javadoc on this class and each flag would be nice

src/main/java/eu/darkbot/api/items/ItemUseResult.java
```diff
@@ -0,0 +1,65 @@
+package eu.darkbot.api.items;
+
+import eu.darkbot.api.managers.HeroItemsAPI;
+
+import java.util.function.Consumer;
+
+/**
+ * {@link HeroItemsAPI#useItem(SelectableItem, ItemUseFlag...)} results
```
Pablete1234, line 8: What a class does is not defined by where it is used,  you can still link or @see it but it is not the definition

`The result after attempting to use an in-game item, usually via {@link ...}`

src/main/java/eu/darkbot/api/items/ItemUseResult.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public enum ItemUseResult {
+    /**
+     * Item was used successfully
+     */
+    SUCCESS,
+    /**
+     * Failed to use an item
```
Pablete1234, line 16: But why? any examples of cases? Is this a generic fallback fail? explain it

src/main/java/eu/darkbot/api/items/ItemUseResult.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    FAILED,
+    /**
+     * Item isn't ready
```
Pablete1234, line 20: But what does ready mean? defining a word with itself isn't too helpful if anyone is confused about what this means when reading our javadocs

src/main/java/eu/darkbot/api/items/ItemUseResult.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Item have insufficient quantity, non-positive, <= 0
+     * Only if {@link ItemUseFlag#POSITIVE_QUANTITY} flag was used
+     *
+     * @see ItemUseFlag#POSITIVE_QUANTITY
+     */
+    INSUFFICIENT_QUANTITY;
+
+    /**
+     * Is executed only on {@link #SUCCESS} result
```
Pablete1234, line 45: This is too strict and limiting definition of the API, if in the future we added other successful results we wouldd be forced to count them as ifFailed not to break the contract.

Say that something along the lines of: `It's executed on successful using of the item, like {@link #SUCCESS}` that way you don't close off the posibility of more than just success being there.

Also maybe consider adding `isSuccessful` that currently returns `this == SUCCESS` but in the future could be expanded. You can put this explanation on `isSuccessful` directly and make ifSuccessful and ifFailed reference it

src/main/java/eu/darkbot/api/managers/HeroItemsAPI.java
```diff
@@ -1,136 +1,46 @@
...
-     * @return list of items associated with given {@link Category}
-     */
-    Collection<? extends Item> getItems(@NotNull HeroItemsAPI.Category category);
-
     /**
-     * Search every {@link Category} for given {@code itemId}.
-     *
-     * @param itemId to be searched for
-     * @return first encounter of given item id
+     * @param selectableItem to get {@link Item} instance
```
Pablete1234, line 17: this is some odd wording

src/main/java/eu/darkbot/api/managers/HeroItemsAPI.java
```diff
@@ -1,136 +1,46 @@
...
-    }
+    Optional<Item> getItem(@NotNull SelectableItem selectableItem);
 
     /**
-     * Search {@link Category} for given {@code itemId}.
-     *
-     * @param category to be searched
-     * @param itemId   to be looked for in given category
-     * @return first encounter of given item id
+     * Returns {@link Collection<Item>} of given {@link ItemCategory}
```
Pablete1234, line 23:
```suggestion
     * @param itemCategory the category to get items from
     * @return a {@link Collection<Item>} with all the items inside the given {@link ItemCategory}
```

src/main/java/eu/darkbot/api/managers/HeroItemsAPI.java
```diff
@@ -1,136 +1,46 @@
...
 
     /**
-     * Filters items in every {@link Category}
+     * This method checks if given {@link Item} can be used in-game.
+     * If the item isn't in any of the action bars it may not be selectable.
      *
-     * @param filter the items with
-     * @return filtered {@link Optional<Item>} or {@link Optional#empty()}
+     * @param item to check
+     * @return non empty optional if item can be used
```
Pablete1234, line 32: I'm pretty sure there's more to it than just this. if the item doesn't exist it won't be non-empty

src/main/java/eu/darkbot/api/managers/HeroItemsAPI.java
```diff
@@ -1,136 +1,46 @@
...
-                return Optional.of(item);
-
-        return Optional.empty();
-    }
-
-    /**
-     * Represents all available categories of {@link HeroItemsAPI} items.
+     * @param selectableItem to be used
+     * @param itemFlags      which this method must respect
+     * @return use result of the selectableItem
```
Pablete1234, line 43: These explanations are again a bit poor

"API should check ..." should api check? are you talking to implementors or users? Make sure these javadocs are aimed at users for the API, not implementors

selectable item has bad wording, don't mix the name of the parameter with the text as if it was part of the same sentence, it just looks odd, this is done in a few other places

itemFlags are not explained at all here "which this method must respect" explains nothing
in the return a link to selectable item result would be helpful for the longer explanation

src/main/java/eu/darkbot/shared/future/FutureQueue.java
```diff
@@ -0,0 +1,22 @@
+package eu.darkbot.shared.future;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+public class FutureQueue<T> implements Runnable {
+    private static ScheduledExecutorService sharedExecutor;
+
+    public FutureQueue(boolean shareThread) {
```
Pablete1234, line 10: I don't think we have a use for this, this class should probably be removed

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SlotBarsProxy.java
```diff
@@ -54,40 +58,56 @@ public boolean isProActionBarVisible() {
...
+        return (!toggleProAction || settings.pressKeybind(SettingsProxy.KeyBind.TOGGLE_PRO_ACTION))
+                && settings.pressKeybind(SettingsProxy.KeyBind.of(slotBarType, slotNumber))
+                ? ItemUseResult.SUCCESS : ItemUseResult.FAILED;
+
     }
 
-    @Override
-    public Collection<? extends eu.darkbot.api.objects.Item> getItems(@NotNull Category category) {
-        return categoryBar.get(category).items;
+    private Optional<? extends eu.darkbot.api.items.Item> findItem(SelectableItem item) {
```
Pablete1234, line 100: i'm pretty sure `get` is more common in the java world, and should probably be prefered over `find` since in this case it fits just as well

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SlotBarsProxy.java
```diff
@@ -94,20 +91,38 @@ public boolean isProActionBarVisible() {
...
+            } else if (!flag.test(item))
+                return flag.getFailResult();
+        }
+
+        return !isUseItem ? null : Stream.of(DEFAULT_ITEM_FLAGS)
+                .filter(f -> f != ItemFlag.NONE)
+                .filter(f -> !f.test(item))
+                .map(ItemFlag::getFailResult)
+                .findFirst().orElse(null);
     }
```
Pablete1234, line 112:
```suggestion
    private ItemUseResult checkItemFlags(Item item, ItemFlag... flags) {
        if (flags == null || flags.isEmpty()) flags = DEFAULT_ITEM_FLAGS;
        for (ItemFlag flag : flags) {
            if (flag == ItemFlag.NONE && flags.length > 1)
                    throw new IllegalArgumentException("ItemFlag.NONE can't be combined with other flag!");
            if (!flag.test(item)) return flag.getFailResult();
        }
        return null;
    }
```
I'm not exactly sure why the isUseItem parameter is even passed, if it's so that you can decide if you should add the default flags, just convert the first line in my suggestion in a return null, and always pass the default item flags from useItem method if the user doesn't set any.

Also, make the NONE flag simply a dummy test/result, so it can be treated just like any other without requiring a special case

Pablete1234 (changes requested):

src/main/java/eu/darkbot/api/items/ItemFlag.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    NOT_SELECTED(item -> !item.isSelected(), ItemUseResult.ALREADY_SELECTED),
+    /**
+     * Item need to have positive quantity, >0
+     */
+    POSITIVE_QUANTITY(item -> item.getQuantity() > 0, ItemUseResult.INSUFFICIENT_QUANTITY),
+    /**
+     * If this flag is passed, cancels every item flag.
+     */
+    NONE(item -> true, null);
```
Pablete1234, line 36:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SlotBarsProxy.java
```diff
@@ -55,39 +59,64 @@ public boolean isProActionBarVisible() {
...
-        return categoryBar.hasCategory(category);
+    private ItemUseResult checkItemFlags(Item item, ItemFlag... flags) {
+        for (ItemFlag flag : flags) {
+            if (flag == ItemFlag.NONE) { // kinda may be not throw if none flag was passed last and test failed for another flag.
+                if (flags.length > 1)
+                    throw new IllegalArgumentException("You cannot pass NONE flag with any other!");
+                return null;
+            } else if (!flag.test(item))
+                return flag.getFailResult();
+        }
```
Pablete1234, line 103:
```suggestion
        for (ItemFlag flag : flags) {
            if (!flag.test(item)) return flag.getFailResult();
        }
```

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #81, opened 2021-06-25
Added DiscordWebhook.java for sending discord webhooks

pajtor:
This class is capable of sending custom discord webhook message

I have also added example usage in the class comments, leaving it here as well.

 Example usage:

 DiscordWebhook webhook = new DiscordWebhook("https://discordapp.com/api/webhooks/...");
 webhook.setContent("Any message!");
 webhook.setAvatarUrl("https://your.awesome/image.png");
 webhook.setUsername("Custom Usernames!");
 webhook.setTts(true);
 webhook.addEmbed(new DiscordWebhook.EmbedObject()
 .setTitle("Title")
 .setDescription("This is a description")
 .setColor(Color.RED)
 .addField("1st Field", "Inline", true)
 .addField("2nd Field", "Inline", true)
 .addField("3rd Field", "No-Inline", false)
 .setThumbnail("https://kryptongta.com/images/kryptonlogo.png")
 .setFooter("Footer text", "https://kryptongta.com/images/kryptonlogodark.png")
 .setImage("https://kryptongta.com/images/kryptontitle2.png")
 .setAuthor("Author Name", "https://kryptongta.com", "https://kryptongta.com/images/kryptonlogowide.png")
 .setUrl("https://kryptongta.com"));
 webhook.addEmbed(new DiscordWebhook.EmbedObject()
 .setDescription("Just another added embed object!"));
 webhook.execute(); //Handle exception


pajtor:
https://gist.github.com/k3kdude/fba6f6b37594eae3d6f9475330733bdb

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -0,0 +1,412 @@
...
+ .addField("2nd Field", "Inline", true)
+ .addField("3rd Field", "No-Inline", false)
+ .setThumbnail("https://kryptongta.com/images/kryptonlogo.png")
+ .setFooter("Footer text", "https://kryptongta.com/images/kryptonlogodark.png")
+ .setImage("https://kryptongta.com/images/kryptontitle2.png")
+ .setAuthor("Author Name", "https://kryptongta.com", "https://kryptongta.com/images/kryptonlogowide.png")
+ .setUrl("https://kryptongta.com"));
+ webhook.addEmbed(new DiscordWebhook.EmbedObject()
+ .setDescription("Just another added embed object!"));
+ webhook.execute(); //Handle exception
```
Pablete1234, line 37:
```suggestion
    Source: https://gist.github.com/k3kdude/fba6f6b37594eae3d6f9475330733bdb
```

Just properly link-back to the original source, if someone wants examples they can get them  there

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -0,0 +1,412 @@
...
+                builder.append(++i == entrySet.size() ? "}" : ",");
+            }
+
+            return builder.toString();
+        }
+
+        private String quote(String string) {
+            return "\"" + string + "\"";
+        }
+    }
```
Pablete1234, line 410: Since we have GSON as a library, let's just use the GSON JSONObject instead of re-implementing an object-to-json library inside this file

Pablete1234 (changes requested):

pajtor:
I am not sure if I refactored it correctly

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -0,0 +1,339 @@
...
+
+                    jsonFields.add(jsonField);
+                }
+//                jsonEmbed.add("fields", jsonFields.toArray());
+                jsonFields.forEach(jsonField -> jsonEmbed.add("fields", jsonField));
+                embedObjects.add(jsonEmbed);
+            }
+//            json.add("embeds", embedObjects.toArray());
+            embedObjects.forEach(embedObject -> json.add("embeds", embedObject));
+        }
```
Pablete1234, line 138: this whole thing can be replaced with a one-liner call to GSON that serializes the content, simply make sure to annotate the fields that don't match (ie: iconUrl needs to be annotated so that the name will be "icon_url" when serialized. see Hangar > Item for an example), 
You'll need either an adapter for color, or, simpler, change the field type to int and make the setter do the conversion

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -0,0 +1,339 @@
...
+        connection.setDoOutput(true);
+        connection.setRequestMethod("POST");
+
+        OutputStream stream = connection.getOutputStream();
+        stream.write(json.toString().getBytes());
+        stream.flush();
+        stream.close();
+
+        connection.getInputStream().close(); //I'm not sure why but it doesn't work without getting the InputStream
+        connection.disconnect();
```
Pablete1234, line 153: use our `Http` library class for this instead, should be just a couple linese using it

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -0,0 +1,258 @@
...
+
+    public void execute() throws IOException {
+        if (this.content == null && this.embeds.isEmpty()) {
+            throw new IllegalArgumentException("Set content or add at least one EmbedObject");
+        }
+
+        Http http = Http.create(this.url, Method.POST);
+        HttpURLConnection conn = http.getConnection();
+        conn.addRequestProperty("Content-Type", "application/json");
+        conn.addRequestProperty("User-Agent", "Java-DiscordWebhook-BY-Gelox_");
```
Pablete1234, line 64: You don't need to set this agent, we already have one
```suggestion
```

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -0,0 +1,258 @@
...
+        conn.setDoOutput(true);
```
Pablete1234, line 65: i'm pretty certain you don't need to set this, Http does it when you use method POST

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -0,0 +1,258 @@
...
+
+        OutputStream stream = conn.getOutputStream();
+        stream.write(GSON.toJson(this).getBytes());
+        stream.flush();
+        stream.close();
+
+        http.closeInputStream();
+        conn.disconnect();
```
Pablete1234, line 73: This is quite a clusterfuck

To start with, Gson allows writing directly to an output stream, second of all, use try-with-resources instead of manually closing streams:

something like this:
```java
try (OutputStreamWriter writer = new OutputStreamWriter(http.getOutputStream()) {
    GSON.toJson(this, writer);
}
```

pajtor:
Doesn't work

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Adaptation of original source: https://gist.github.com/k3kdude/fba6f6b37594eae3d6f9475330733bdb
+ */
+public class DiscordWebhook {
+    private static final Gson GSON = new Gson();
+    private final List<EmbedObject> embeds = new ArrayList<>();
+    private String content;
+    private String username;
+    @SerializedName("avatar_url") private String avatarUrl;
+    private boolean tts;
```
Pablete1234, line 23:
```suggestion
    private static final Gson GSON = new Gson();
    
    private String content;
    private String username;
    @SerializedName("avatar_url") private String avatarUrl;
    private boolean tts;
    private final List<EmbedObject> embeds = new ArrayList<>();
```

It seems like the order of the fields & getters/setters is a bit over the place, it'd be nice of both the fields and the getters/setters followed the order in https://discord.com/developers/docs/resources/webhook#execute-webhook

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -0,0 +1,238 @@
...
+    public static class EmbedObject {
+        private final List<Field> fields = new ArrayList<>();
+        private String title;
+        private String description;
+        private String url;
+        private Color color;
+        private Footer footer;
+        private Thumbnail thumbnail;
+        private Image image;
+        private Author author;
```
Pablete1234, line 65:
```suggestion

        private String title;
        private String description;
        private String url;
        private Color color;
        private Footer footer;
        private Image image;
        private Thumbnail thumbnail;
        private Author author;
        private final List<Field> fields = new ArrayList<>();
```

Same thing here, with https://discord.com/developers/docs/resources/channel#embed-object

it also seems like some fields have getters and setters scattered in here, make them properly ordered like: getA, setA, getB, setB, currently i can see getAuthor, getFields, setFooter, setAuthor, in that order, which is odd

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/utils/DiscordWebhook.java
```diff
@@ -43,6 +45,10 @@ public void addEmbed(EmbedObject embed) {
         this.embeds.add(embed);
     }
 
+    public void addEmbed(Function<EmbedObject, EmbedObject> function) {
+        addEmbed(function.apply(new EmbedObject()));
+    }
+
```
Pablete1234, line 51: This is just stupid, it really does not help make code any shorter from the calling client

if you want to make this better-to-use, make builder-style setters:
```java
public DiscordWebhook setX(X x) {
    this.x = x;
    return this;
}
```
as this allows chaning calls from the client using the api:
```java
new DiscordWebhook()
    .setX("asdf")
    .setY("asdf2)
    .execute("url");
```

pajtor: so should I remove this way?

Pablete1234 (approved):

PR #83, opened 2021-06-26
Initial commit of JButtonField

pajtor:


Pablete1234:
How exactly is this editor useful in any way?

pajtor:
¯\_(ツ)_/¯ because I need buttons to perform an action in config, but I will use the beforeConfig method then

Pablete1234:
> ¯_(ツ)_/¯ because I need buttons to perform an action in config, but I will use the beforeConfig method then

The behavior on this button cannot be edited as it is right now, which is why i'm asking how is this useful

pajtor:
Will try doing this later in PluginAPI instead

PR #89, opened 2021-08-17
added stat reset feature to ExtraButtons

Zodey-hub:


src/main/java/com/github/manolo8/darkbot/gui/titlebar/ExtraButton.java
```diff
@@ -3,7 +3,9 @@
 import com.github.manolo8.darkbot.Main;
 import com.github.manolo8.darkbot.config.ColorScheme;
 import com.github.manolo8.darkbot.config.ConfigEntity;
+import com.github.manolo8.darkbot.config.types.suppliers.DisplayFlag;
```
Pablete1234, line 6: This import shouldn't be here
```suggestion
```

src/main/java/com/github/manolo8/darkbot/gui/titlebar/ExtraButton.java
```diff
@@ -150,7 +152,13 @@ public void actionPerformed(ActionEvent e) {
                 ConfigEntity.changed();
                 main.getGui().updateConfiguration();
             }));
-
+            list.add(create("reset_stats", e -> {
+                main.statsManager.earnedUridium = 0;
+                main.statsManager.earnedCredits = 0;
+                main.statsManager.earnedExperience = 0;
+                main.statsManager.earnedHonor = 0;
+                main.guiManager.deaths = 0;
```
Pablete1234, line 160: This is force-setting values in statsManager, instead, it should be calling the method in stats manager that resets it

Keep in mind this is not resetting runtime for example, which is unintended behavior as it would break uri/h etc

Pablete1234 (commented):
Additionally, you need to add a key in strings_en.properties for this extra menu entry, so that it has a proper text and not just "stats_reset"

Pablete1234 (approved):

PR #91, opened 2021-08-24
added draw pet stats

Zodey-hub:


Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/config/ColorScheme.java
```diff
@@ -15,6 +15,7 @@
     public @Option Color PORTALS = new Color(0xAEAEAE);
     public @Option Color HERO = new Color(0x22CC22);
     public @Option @Col(alpha = false) Color TRAIL_BASE = new Color(0xE0E0E0);
+    public @Option Color FUEL = new Color(0xF2F2F2);
```
Pablete1234, line 18: This is missing a the translation key `cs.fuel` in the strings_en file

Pablete1234 (approved):

Zodey-hub:
Do I need to test it with OBS then send it?

Pablete1234:
If you want to record a bit of how it's running etc that'd be nice, i still want to checkout the code locally and run it for a bit to double check everything runs smoothly

Zodey-hub:
Okey, 5 min and I send it

Zodey-hub:
https://streamable.com/7kd09z

Pablete1234:
From what i see, while loading pet health bar is shown, that's a bug.

The constructor for Pet that has no parameters should set `removed = true;` which will be updated later when addresses are found etc

PR #94, opened 2021-08-28
Improve precision of library path

ODAKAB:
Prevents loading from unwanted paths on odd cases.

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/utils/LibUtils.java
+++ b/src/main/java/com/github/manolo8/darkbot/utils/LibUtils.java
@@ -8,7 +8,7 @@ public class LibUtils {
      * @return Absolute path of library with extension based on OS
      */
     public static String getLibPath(String library) {
-        return Paths.get("lib", library).toAbsolutePath().toString() + getExtension();
+        return Paths.get("lib", library + getExtension()).toAbsolutePath().toString();
     }
 
     /**
--- a/src/main/java/com/github/manolo8/darkbot/utils/LibUtils.java
+++ b/src/main/java/com/github/manolo8/darkbot/utils/LibUtils.java
@@ -4,10 +4,18 @@ import java.nio.file.Paths;
 
 public class LibUtils {
     /**
-     * Loads specified library with improved precision of path, goals to avoid loading from unwanted places.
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/manolo8/darkbot/utils/LibUtils.java
```diff
@@ -0,0 +1,13 @@
...
+import java.nio.file.Paths;
+
+public class LibUtils {
+    /**
+     * Loads specified library with improved precision of path, goals to avoid loading from unwanted places.
+     * @param library Filename of library including extension, e.g. "DarkBoatAPI.dll"
+     */
+    public static void loadLibrary(String library) {
+        System.load(Paths.get("lib", library).toAbsolutePath().toString());
+    }
```
Pablete1234, line 12: Don't call system.load from here, make this util just about returning a path, and call system.load in the class where it is used:

`System.load(LibUtils.getPath("DarkBoatAPI"));`
Also remove the need for the extension when calling, and add in the dll or so depending on the OS it runs on

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/utils/LibUtils.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.nio.file.Paths;
+
+public class LibUtils {
+    /**
+     * @param library Filename of library without extension, e.g. "DarkBoatAPI"
+     * @return Absolute path of library with extension based on OS
+     */
+    public static String getLibPath(String library) {
+        return Paths.get("lib", library).toAbsolutePath().toString() + getExtension();
```
Pablete1234, line 11:
```suggestion
        return Paths.get("lib", library + getExtension()).toAbsolutePath().toString();
```
it's the file that has an extension, not an appended thing at the end of the path

Pablete1234 (changes requested):

Pablete1234 (approved):

ODAKAB:
Still works on my end after https://github.com/darkbot-reloaded/DarkBot/pull/94/commits/e4e6a928d0fb988626d8f3dc995662fcae9275cc

PR #98, opened 2021-10-09
Add method to check if GroupManager is busy

ODAKAB:
Good for legacy to use in behaviours


ODAKAB:
nvm biggest issue is game itself again, it bugs out because it's constantly updating textbox with player name, so constantly empties if it's a npc

PR #101, opened 2021-10-13
Prevent handling group while it's loading after jump

ODAKAB:
didn't do a live test but seems ok
need reviews

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/manager/GroupManager.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/manager/GroupManager.java
@@ -69,10 +69,10 @@ public class GroupManager extends Gui {
     }
 
     public void tick() {
-        if (group.isValid()) lastValidTime = System.currentTimeMillis();
-        if (!isLoaded()) return;
-
         if (group.address == 0) return;
+
+        if (group.isValid()) lastValidTime = System.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/manolo8/darkbot/core/manager/GroupManager.java
```diff
@@ -67,6 +69,9 @@ public void update() {
     }
 
     public void tick() {
+        if (group.isValid()) lastValidTime = System.currentTimeMillis();
+        if (!isLoaded()) return;
+
         if (group.address == 0) return;
```
Pablete1234, line 75:
```suggestion
        if (group.address == 0) return;

        if (group.isValid()) lastValidTime = System.currentTimeMillis();
        else if (System.currentTimeMillis() - lastValidTime < 10_000L) return; // Wait until reacting to group being invalid
```

from a logic pov, if the address is 0 you don't even need to check for if the group is valid,there is just no group whatsoever, and so, it shouldn't be marked as valid or even try to tick.

Second of all i'd avoid adding an extra isLoaded function for this as it can be solved with just an if-else

ODAKAB: okay, this one is missing a `&& System.currentTimeMillis != lastValidTime`

PR #102, opened 2021-11-03
Compress png resources (lossless)

ODAKAB:
using [PNGGauntlet](https://pnggauntlet.com/)


dm94:
@ODAKAB  Try this: https://squoosh.app/

Pablete1234:
These are 500 byte images, not 1mb images, there's little more to gain and even if there is anything to gain it's just not worth the effort

ODAKAB:
> @ODAKAB Try this: https://squoosh.app/

couldn't figure out which options are the best, and also lossless
imo pnggauntlet already pushes it to the limits and very cpu intensive by nature

PR #103, opened 2021-11-05
Ignoring NPCs radius before first attack

QR4SH3D:
The bot keeps a distance between target and hero and it takes time to shoot the target but with this it will ignore the radius before the first shot


Pablete1234:
Hey, if it ignores the radius before first attack, what radius does it use? 0?

Is this setting aimed at keeping a safer distance or at keeping a more aggressive distance?

What would the alternative of simply making radius smaller look like, why would that not be viable?

What is the main use-case of this?

QR4SH3D:
Hai
1. 300
2. More aggressive, only before the first shot.
3. Because if you keep close range the target will still hit you while killing but if it only comes close to the NPCs before the first shot to target it I realized when I keep the Kristallon for example on 600 (which is a perfect radius for me to not get hit) it sometimes takes a while to lock/shoot I don't know why but using this solved it for me I can provide a video of what I mean if you think thats necessary 👍

Pablete1234:
I'd consider renaming the feautre to Agressive Approach (to kinda hit at Agressive Follow) which is probably more descriptive of what this is, and instead of hardcoding 300 radius also do a % of original, maybe 75% just like agressive follow or if that doesn't work well enough try 60% or 50%.
Make sure it doesn't stack with agressive follow

QR4SH3D:
Yeah that name sounds better I will play with the percentages and see what is better I didn't get what you meant with "Make sure it doesn't stack with agressive follow" can you explain that again? Thanks

QR4SH3D:
I made little changes is it okay like this? still don't know what you meant with that sentence ^

Pablete1234:
What i mean with it is that if you approach an npc with loe hp, it should not do both 75% from aggressive follow AND 55% from aggressive approach, it should only do one of the 2

QR4SH3D:
Well aggressive approach does not constantly have the radius % it only does it when shooting the first shot it isn't meant to get 55% closer all the time just before it shoots so I don't get what you're saying now because aggressive follow is when NPC is on lowHp and makes the radius a bit smaller and aggressive approach is getting closer only the first time before he shoots not constantly

Pablete1234:
> Well aggressive approach does not constantly have the radius % it only does it when shooting the first shot it isn't meant to get 55% closer all the time just before it shoots so I don't get what you're saying now because aggressive follow is when NPC is on lowHp and makes the radius a bit smaller and aggressive approach is getting closer only the first time before he shoots not constantly

Aggressive approach does 55% less when approaching, Aggressive follow does 25% less when the npc is under 25% hp. If you are approaching an NPC that is at less than 25% hp, we don't want to be using both 55% and 25% less, as it rersults in like ~67% less. 
The idea is that it should just take aggressive approach (55% less) and if not, try aggressive follow (25% less), but not both at once

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -133,10 +136,12 @@ public double modifyRadius(double radius) {
         if (target.health.hpPercent() < 0.25 && target.npcInfo.extra.has(NpcExtra.AGGRESSIVE_FOLLOW)) radius *= 0.75;
         if (target != hero.target || !hero.isAttacking(target) || castingAbility()) radius = Math.min(550, radius);
         else if (!target.locationInfo.isMoving() || target.health.hpPercent() < 0.25) radius = Math.min(600, radius);
+        if (target.npcInfo.extra.has((NpcExtra.AGGRESSIVE_APPROACH)) && !closeRange) radius *= 0.55;
+        radius += bar.findItemById("ability_zephyr_mmt").map(i -> i.quantity).orElse(0d) * 5;
 
-        return radius + bar.findItemById("ability_zephyr_mmt").map(i -> i.quantity).orElse(0d) * 5;
+        if (main.hero.locationInfo.distance(target) < radius + 0.45) closeRange = true;
```
Pablete1234, line 142: This 0.45 is an absolute value, i'm pretty sure it's intended to be a multiplying factor
The issue is this may never be true, if you try to get to a distance of 600 with an npc, if the npc is moving away from you you will never get 600 units close.
When you change this `closeRange` variable may not be reliable at all as it is here

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -133,10 +136,12 @@ public double modifyRadius(double radius) {
...
+        return radius;
     }
-
```
Pablete1234, line 139: keep the empty line separating functions

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -130,13 +133,15 @@ private void sendAttack(long minWait, long bugTime, boolean normal) {
...
     public double modifyRadius(double radius) {
-        if (target.health.hpPercent() < 0.25 && target.npcInfo.extra.has(NpcExtra.AGGRESSIVE_FOLLOW)) radius *= 0.75;
+        if (target.npcInfo.extra.has((NpcExtra.AGGRESSIVE_APPROACH)) && !closeRange) radius *= 0.55;
+        else if (target.health.hpPercent() < 0.25 && target.npcInfo.extra.has(NpcExtra.AGGRESSIVE_FOLLOW)) radius *= 0.75;
         if (target != hero.target || !hero.isAttacking(target) || castingAbility()) radius = Math.min(550, radius);
         else if (!target.locationInfo.isMoving() || target.health.hpPercent() < 0.25) radius = Math.min(600, radius);
+        radius += bar.findItemById("ability_zephyr_mmt").map(i -> i.quantity).orElse(0d) * 5;
 
-        return radius + bar.findItemById("ability_zephyr_mmt").map(i -> i.quantity).orElse(0d) * 5;
+        if (main.hero.locationInfo.distance(target) < radius) closeRange = true;
```
Pablete1234, line 142: If you put this at the very start of the function, it may be viable to see it happening
but in here, with the already reduced radius, it will likely never happen that you get closer

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -130,13 +133,15 @@ private void sendAttack(long minWait, long bugTime, boolean normal) {
...
+        return radius;
     }
-
```
Pablete1234, line 139: keep the empty line separating functions

src/main/java/com/github/manolo8/darkbot/modules/utils/NpcAttacker.java
```diff
@@ -42,6 +42,8 @@
     protected boolean sab; // If shooting SAB right now
     protected boolean rsb; // If shooting RSB right nos
 
+    protected boolean closeRange; // Ignore radius before first shot
```
Pablete1234, line 45:
```suggestion
    protected boolean closeRange; // If we've been close enough to shoot this npc yet
````
Update this comment, the variable is not about ignoring, the variable is only about keeping track of if we've been close enough

Pablete1234 (changes requested):

Pablete1234 (approved):

ODAKAB:
I am also working on a similar context for Eternal Gate, sadly i can confirm this is a placebo because game is really unstable.

ODAKAB:
Compare with https://github.com/darkbot-reloaded/DarkBot/pull/110 please.

PR #105, opened 2021-11-12
Fix ItemTimer not updating

ODAKAB:
yes

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/objects/slotbars/Item.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/objects/slotbars/Item.java
@@ -71,7 +71,7 @@ public class Item extends UpdatableAuto {
         @Override
         public void update() {
             if (address == 0) {
-                resetTimer();
+                reset();
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


pajtor:
no

src/main/java/com/github/manolo8/darkbot/core/objects/slotbars/Item.java
```diff
@@ -80,10 +82,20 @@ public void update() {
...
+            if (address == 0) {
+                resetTimer();
+                return;
+            }
 
             this.startTime = API.readMemoryDouble(address + 80);
             this.itemDelay = API.readMemoryDouble(address + 88);
         }
+
+        public void resetTimer() {
```
Pablete1234, line 94: A function inside itemtimer does not need to call itself resetTimer, this is already the timer
just rename it to reset
```suggestion
        public void reset() {
```

Pablete1234 (changes requested):

Pablete1234 (changes requested):
This does not compile

Pablete1234 (approved):

PR #109, opened 2021-11-29
Performance + bugfixes

cybooo:
Fixing some small mistakes.

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/gui/players/PlayerRenderer.java
+++ b/src/main/java/com/github/manolo8/darkbot/gui/players/PlayerRenderer.java
@@ -12,7 +12,6 @@ import javax.swing.border.EmptyBorder;
 import java.awt.*;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
 
 public class PlayerRenderer extends JPanel implements ListCellRenderer<PlayerInfo> {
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/manolo8/darkbot/gui/players/PlayerRenderer.java
```diff
@@ -83,8 +84,8 @@ public void repaint(Rectangle r) {}
     @Override
     protected void firePropertyChange(String propertyName, Object oldValue, Object newValue) {
         // Strings get interned...
-        if (propertyName == "text"
-                || ((propertyName == "font" || propertyName == "foreground")
+        if (Objects.equals(propertyName, "text")
+                || ((Objects.equals(propertyName, "font") || Objects.equals(propertyName, "foreground"))
```
Pablete1234, line 88: I am very well aware that strings should not be compared with ==, however, this piece of code comes from the JDK.

The strings are interned and it is known that == will be reliable, also for performance reasons it is much quicker to identity equality than to check the actual equality of content.

Have you actually encountered a situation where this is an issue or are you just changing these based off of IDE warnings?


For reference purposes, the JDK DefaultListCellRenderer: https://github.com/openjdk/jdk/blob/master/src/java.desktop/share/classes/javax/swing/DefaultListCellRenderer.java#L260

cybooo: Wasn't aware that that piece of code comes from the JDK. In that case, it should be alright - Yes, string comparison with == did cause me issues, but in this case, it is probably meant to compare the reference of the String, which would be correct.

cybooo: Commit reverted.

Pablete1234: I meant if you had found an issue where this specific function using == didn't work.

I am well aware that string identity equality does not work for the normal use-case anywhere else and that it WILL break and cause issues

PR #113, opened 2021-12-17
Add linux support

Alph4rd:


dm94:
I am completely unfamiliar with how it works.
But wouldn't it be possible to detect which operating system is being used and on that basis activate or deactivate functions?
I know that many offsets would have to be changed, but based on a global variable it could be set to use one or the other.

fabio1999ita:
> I am completely unfamiliar with how it works. But wouldn't it be possible to detect which operating system is being used and on that basis activate or deactivate functions? I know that many offsets would have to be changed, but based on a global variable it could be set to use one or the other.

yeah, should be also easy and need to adapt the download of API too

Pablete1234:
> But wouldn't it be possible to detect which operating system is being used and on that basis activate or deactivate functions? I know that many offsets would have to be changed, but based on a global variable it could be set to use one or the other.

Yes, when this PR is to be merged, it will be with such capability to adapt automatically to the OS, atm, the changes in here have been slowly being refactored/merged into master to make this PR more lightweight (less OS-specific things), and when it gets to be a minimum set of differences, those differences will have to work off of a toggle based on the OS, which darkbot will handle automatically.

Regarding browser native APIs, the libraries for this are not redistributed alongside the rest of windows-based apis (yet) because this isn't merged (yet). When this gets official support and is merged into master, the main libs will simply download the correct natives for linux which will be diff to the ones from windows.

src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -224,7 +224,7 @@ public static class WindowPosition {
 
         public @Option @Visibility(Level.ADVANCED) APIConfig API_CONFIG = new APIConfig();
         public static class APIConfig {
-            public @Option @Dropdown BrowserApi BROWSER_API = BrowserApi.KEKKA_PLAYER;
+            public @Option @Dropdown BrowserApi BROWSER_API = BrowserApi.TANOS_API;
```
Pablete1234, line 227:
```suggestion
            public @Option @Dropdown BrowserApi BROWSER_API = OSUtils.isLinux() ? BrowserApi.TANOS_API : BrowserApi.KEKKA_PLAYER;
```

src/main/java/com/github/manolo8/darkbot/core/manager/MapManager.java
```diff
@@ -421,7 +421,9 @@ private Location getEnemyLocatorTarget() {
     }
 
     private boolean findMarker(long spriteArray, double scale, Location result) {
-        int size = API.readMemoryInt(spriteArray, 0x40, 0x18);
+        long array_data = API.readMemoryLong(spriteArray, 0x40);
+        if (array_data == 0 ) return false;
+        int size = API.readMemoryInt(array_data + 0x20);
```
Pablete1234, line 426: these are equivalent except for the 0x18 -> 0x20 offset, readMemoryLong checks at every step of the way the ptr is valid now

src/main/java/com/github/manolo8/darkbot/core/manager/MapManager.java
```diff
@@ -337,12 +337,12 @@ private void updateBounds() {
 
             renderValidated = false;
             is3DView = !main.settingsManager.is2DForced()
-                       && ByteUtils.readObjectName(API.readLong(viewAddress + 208)).contains("HUD");
+                && ByteUtils.readObjectName(API.readLong(viewAddress + 208)).contains("HUD");
```
Pablete1234, line 340: Undo the change plz

src/main/java/com/github/manolo8/darkbot/core/api/TanosAdapter.java
```diff
@@ -0,0 +1,174 @@
...
+        private void sendNotification(String notification, int... args) {
+            if (botInstaller.screenManagerAddress.get() == 0) return;
+
+            long[] tagged = new long[args.length];
+            for (int i = 0; i < args.length; i++) {
+                tagged[i] = ByteUtils.tagInteger(args[i]);
+            }
+
+            tanos.sendNotification(botInstaller.screenManagerAddress.get(), notification, tagged);
+        }
```
Pablete1234, line 142: This whole piece should be moved to somewhere more reusable, potentially either GameApiImpl, or GameApi.DirectInteraction, where it expects another method (sendNotification) to be implemented and just calls that, that makes it so kekka and tanos don't need to copy-paste this block

src/main/java/com/github/manolo8/darkbot/core/manager/MouseManager.java
```diff
@@ -1,6 +1,7 @@
 package com.github.manolo8.darkbot.core.manager;
 
 import com.github.manolo8.darkbot.Main;
+import com.github.manolo8.darkbot.core.entities.Box;
```
Pablete1234, line 4:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/core/utils/factory/EntityFactory.java
```diff
@@ -132,8 +132,11 @@ public static EntityBuilder find(long address) {
     }
 
     private static String getAssetId(long address) {
-        long temp = API.readMemoryLong(address, 48, 48, 16) & ByteUtils.ATOM_MASK;
-        return API.readMemoryString(temp, 64, 32, 24, 8, 16, 24).trim();
+        // trait list
+        
+        long trait = API.readMemoryLong(address, 0x30, 0x30, 0x10) & ByteUtils.ATOM_MASK;
+        String r = API.readMemoryString(trait, 0x40, 0x20, 0x18, 0x10, 0x10, 0x18).trim();
+        return r;
```
Pablete1234, line 139:
```suggestion
        long trait = API.readMemoryLong(address, 0x30, 0x30, 0x10) & ByteUtils.ATOM_MASK;
        return API.readMemoryString(trait, 0x40, 0x20, 0x18, 0x10, 0x10, 0x18).trim();
```

src/main/java/com/github/manolo8/darkbot/utils/ApiErrors.java
```diff
@@ -26,7 +26,7 @@ private static String getMessage(BrowserApi api) {
             return "You're using a <strong>removed API</strong>, use the recommended API!";
         if (api == BrowserApi.NO_OP_API)
             return "Error loading the no-op api, how did this happen?";
-        if (api != BrowserApi.KEKKA_PLAYER)
+        if (api != BrowserApi.TANOS_API)
```
Pablete1234, line 29:
```suggestion
        if (api != (OSUtil.isLinux() ? BrowserApi.TANOS_API : BrowserApi.KEKKA_PLAYER))
```

maybe `OSUtil.getDefaultAPI` that just does this logic and re-use it in the 3 places it's used at?

Pablete1234 (changes requested):
This is generally looking very, very compatible with the current codebase, unlike in the past where it was many, many changes.

I believe the main thing is sprite offset change from 0x18 to 0x20 in a few places, all other offset changes need to be double-checked but i believe they're just dec to hex conversions with the same value.

Overall very happy to see this be so close to a merge

Pablete1234 (approved):

PR #115, opened 2021-12-21
Frozen lab synk map and zone

anjeshshrestha:
need to test if reading zone location is correct or not


src/main/java/com/github/manolo8/darkbot/core/objects/facades/FrozenLabyrinthProxy.java
```diff
@@ -9,13 +9,16 @@
     public double time;
     public int keys;
     public String labStatus;
-
+    public String synkMap;
+    public int synkZone;
     public void update() {
```
Pablete1234, line 14:
```suggestion
    public int synkZone;

    public void update() {
```

PR #116, opened 2021-12-21
Add possibility to disable death limit

ODAKAB:
Requested feature: https://discord.com/channels/523159099870019584/564190622492262402/916678583014023208


Pablete1234:
I'm going to reject this one because you can set the limit to a very high number for essentially the same behavior, and i think users may be more likely to be confused by if they set deaths to 0 to make the bot stop it causing it to never stop on death

PR #117, opened 2021-12-23
Add possibility to disable master password first time

fabio1999ita:
old vs new (the layout can still be improved)

image

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/gui/login/SavedLogins.java
+++ b/src/main/java/com/github/manolo8/darkbot/gui/login/SavedLogins.java
@@ -127,7 +127,7 @@ public class SavedLogins extends JPanel implements LoginScreen {
             button.setEnabled(pass.getPassword().length > 0 || e.getStateChange() == ItemEvent.SELECTED);
         });
         pass.getDocument().addDocumentListener((GeneralDocumentListener) e -> 
-                button.setEnabled(pass.getPassword().length > 0 || check.isSelected());
+                button.setEnabled(pass.getPassword().length > 0 || check.isSelected()));
         button.addActionListener(e -> SwingUtilities.getWindowAncestor(button).setVisible(false));
 
         JOptionPane pane = new JOptionPane(panel, JOptionPane.PLAIN_MESSAGE, JOptionPane.DEFAULT_OPTION, null, new Object[]{}, null);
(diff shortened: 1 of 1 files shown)


fabio1999ita:
video

fabio1999ita:
video

Pablete1234:
Currently you're using the Message in the jdialog for the string, then adding an "option" (usually, the list of buttons to show on bottom-right) that is a whole jpanel with everything

Instead, send just one jpanel as the message, which contains a jlabel, the field, and the button
lastly, make the button be the windows' default button, that way it will appear blue hinting the user that it should be their action to click

fabio1999ita:
video

src/main/java/com/github/manolo8/darkbot/gui/login/SavedLogins.java
```diff
@@ -109,14 +111,41 @@ private void updateList() {
...
+
+        check.addItemListener(e -> {
+            if (e.getStateChange() == ItemEvent.SELECTED) {
+                pass.setEditable(false);
+                button.setEnabled(true);
+            } else {
+                pass.setEditable(true);
+                if (new String(pass.getPassword()).isEmpty())
+                    button.setEnabled(false);
+            }
```
Pablete1234, line 133:
```suggestion
            pass.setEditable(e.getStateChange() != ItemEvent.SELECTED);
            button.setEnabled(pass.getPassword().length > 0 || e.getStateChange() == ItemEvent.SELECTED);
```

This logic can be cleaned up & simplified a bit

src/main/java/com/github/manolo8/darkbot/gui/login/SavedLogins.java
```diff
@@ -109,14 +111,41 @@ private void updateList() {
...
+        });
+        pass.getDocument().addDocumentListener((GeneralDocumentListener) e -> button.setEnabled(!new String(pass.getPassword()).isEmpty() || check.isSelected()));
```
Pablete1234, line 135:
```suggestion
        pass.getDocument().addDocumentListener((GeneralDocumentListener) e -> 
                button.setEnabled(pass.getPassword().length > 0 || check.isSelected());
```
This line is too long to read comfortably, split it into a new line, and also avoid the new String allocation each time

Pablete1234 (changes requested):
Just a couple final touchups but it looks good :+1: 

keep up the good work

PR #118, opened 2021-12-23
added background image setting

anjeshshrestha:


anjeshshrestha:
also added getImage function to support user specified extension

src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -227,6 +227,7 @@
             public @Option @Num(max = 300, step = 1) int TRAIL_LENGTH = 15;
             public @Option boolean MAP_START_STOP = false;
             public @Option(key = "colors") ColorScheme cs = new ColorScheme();
+            public @Option boolean CUSTOM_BACKGROUND = false;
```
Pablete1234, line 230: This doesn't look like an easy way for a user to configure this custom background

Consider making it a File that the user can select via a FileChooser, or even an Image or a wrapper of an image, and an editor that essentially updates the path from a file picker

So you have essentially an object looking like:
```java
class ImageWrapper {
    private String path;
    private transient Image image;
    private transient boolean loaded;
    
    public void setPath(String path) {
        this.path = path;
        this.loaded = false;
    }
    
    public Image getImage() {
        if (loaded) return image;
        loaded = true;
        try {
            return image = ... ;
        } catch (Exception e) {
            e.printStackTrace():
            return null;
        }
    }
}
```

The editor for this type of ImageWrapper just calls setPath whenever the user selects a different path from a file picker, and then from mapDrawer you can just call getImage and if it's null do nothing

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -158,13 +160,25 @@ public void setup(Main main) {
...
+
     }
 
     @Override
     protected void paintComponent(Graphics g) {
         if (main == null) return;
         Graphics2D g2 = setupDraw(g);
 
+        if(main.config.BOT_SETTINGS.MAP_DISPLAY.CUSTOM_BACKGROUND && image != null){
+            g2.setComposite(AlphaComposite.getInstance(AlphaComposite.SRC_OVER, 0.3f));
```
Pablete1234, line 178: It'd be cool to be able to configure that 0.3f as a transparency parameter in the config, but it's fine if it stays like this too

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -158,13 +160,25 @@ public void setup(Main main) {
         this.ships = main.mapManager.entities.ships;
         this.battleStations = main.mapManager.entities.battleStations;
         this.basePoints = main.mapManager.entities.basePoints;
+
+        try {
+            image = UIUtils.getImage("background");
+        }catch (Exception e){
```
Pablete1234, line 166:
```suggestion
        } catch (Exception e) {
```
Missing spaces

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -158,13 +160,25 @@ public void setup(Main main) {
...
+        }
+
     }
 
     @Override
     protected void paintComponent(Graphics g) {
         if (main == null) return;
         Graphics2D g2 = setupDraw(g);
 
+        if(main.config.BOT_SETTINGS.MAP_DISPLAY.CUSTOM_BACKGROUND && image != null){
```
Pablete1234, line 177:
```suggestion
        if (main.config.BOT_SETTINGS.MAP_DISPLAY.CUSTOM_BACKGROUND && image != null) {
```
Missing spaces

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -158,13 +160,25 @@ public void setup(Main main) {
...
+            g2.setComposite(AlphaComposite.getInstance(AlphaComposite.SRC_OVER, 0.3f));
+            g2.drawImage(image,0,0,width,height,null);
```
Pablete1234, line 179:
```suggestion
            g2.drawImage(image, 0, 0, width, height, null);
```
Missing spaces

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
anjeshshrestha: that is really good idea, i added that in

src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -227,6 +225,12 @@
...
+            public @Option @Editor(JPercentField.class) double TRANSPARENCY = 0.3f;
+            @Option
+            @Editor(value=JFileOpener.class, shared = true)
+            public String CHOOSER = "";
```
Pablete1234, line 232: You're gonna need to make an editor for specifically images, make it either extend JFileOpener or be similar (mostly copy-pasted from JFileOpener).
Make the type it saves to be the ImageWrapper defined in an earlier comment (https://github.com/darkbot-reloaded/DarkBot/pull/118/files#r775158000), make it so the editor calls setPath when a user picks a different file in the file chooser, and call getImage from map drawer

src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -239,6 +232,10 @@
             public @Option @Number(max = 300, step = 1) int TRAIL_LENGTH = 15;
             public @Option boolean MAP_START_STOP = false;
             public @Option("colors") ColorScheme cs = new ColorScheme();
+            public @Option boolean CUSTOM_BACKGROUND = false;
+            public @Option @Percentage double TRANSPARENCY = 0.3f;
+            public @Option ImageWrapper CHOOSER = new ImageWrapper();
```
Pablete1234, line 237: I'd personally bundle these inside a sub-confid, that will make ti clearer that "transparency" is also about the image

Custom Background: 
    - Enabled
    - Opacity
    - Image

src/main/java/com/github/manolo8/darkbot/config/ImageWrapper.java
```diff
@@ -0,0 +1,33 @@
...
+
+import javax.swing.*;
+import java.awt.*;
+
+public class ImageWrapper {
+    private String path;
+    private transient Image image;
+    private transient boolean loaded;
+
+    public String getPath(){
```
Pablete1234, line 11:
```suggestion
    public String getPath() {
```

src/main/java/com/github/manolo8/darkbot/config/ImageWrapper.java
```diff
@@ -0,0 +1,33 @@
...
+    public void setPath(String path) {
+        this.path = path;
+        this.loaded = false;
+    }
+
+    public Image getImage() {
+        if (loaded) return image;
+        loaded = true;
+        try {
+            return image = new ImageIcon(this.path).getImage() ;
```
Pablete1234, line 26:
```suggestion
            return image = new ImageIcon(this.path).getImage();
```

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -140,6 +143,8 @@ public void mouseDragged(MouseEvent e) {
                 hero.drive.move(undoTranslateX(e.getX()), undoTranslateY(e.getY()));
             }
         });
+
+        //image = new ImageIcon(main.config.BOT_SETTINGS.MAP_DISPLAY.CHOOSER).getImage();
```
Pablete1234, line 147:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/gui/tree/EditorProvider.java
```diff
@@ -1,22 +1,10 @@
...
-import com.github.manolo8.darkbot.gui.tree.editors.FontEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.MultiDropdownEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.NumberEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.PercentEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.PlayerTagEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.RangeEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.ShipModeEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.StringEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.TableEditor;
+import com.github.manolo8.darkbot.gui.tree.editors.*;
```
Pablete1234, line 7: disable wildcard imports in your IDE

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImageOpener.java
```diff
@@ -0,0 +1,72 @@
...
+import eu.darkbot.api.config.util.OptionEditor;
+import com.github.manolo8.darkbot.gui.utils.Strings;
+import eu.darkbot.api.config.ConfigSetting;
+
+import javax.swing.*;
+import javax.swing.filechooser.FileNameExtensionFilter;
+import java.awt.*;
+import java.io.File;
+
+public class ImageOpener extends JButton implements OptionEditor<ImageWrapper> {
```
Pablete1234, line 14:
```suggestion
public class ImagePicker extends JButton implements OptionEditor<ImageWrapper> {
```
I think image picker is a better suited name for this

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImageOpener.java
```diff
@@ -0,0 +1,72 @@
...
+                fc = new JFileChooser(new File(".")) {
+                    @Override
+                    protected JDialog createDialog(Component parent) throws HeadlessException {
+                        JDialog dialog = super.createDialog(parent);
+                        dialog.setAlwaysOnTop(true);
+                        return dialog;
+                    }
+                };
+                FileNameExtensionFilter filter = new FileNameExtensionFilter(
+                        "PNG, JPG & GIF Images", "png","jpeg","jpg", "gif");
```
Pablete1234, line 31:
```suggestion
                        "PNG, JPG & GIF Images", "png", "jpeg", "jpg", "gif");
```
are you sure that  "PNG, JPG & GIF Images" is a valid filter for this? (i have no clue but it looks off)

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImageOpener.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public JComponent getEditorComponent(ConfigSetting<ImageWrapper> configSetting) {
+        setEditing(configSetting.getValue());
+        return this;
+    }
+
+    private void setEditing(ImageWrapper image){
+        this.image = image; //do i need this, i dont think its needed
+        setText(image.getPath());
```
Pablete1234, line 59: I would simplify the text-representation of this, showing a whole system path is often impractical (too long)
You can probably just trim for the last N sections of the path based on a max aim length, say, 80 chars max, so instead of:
c:\Users\User\Desktop\Images\darkbot\image.jpg
it would maybe trim to:
...\Images\darkbot\image.jpg

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImageOpener.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class ImageOpener extends JButton implements OptionEditor<ImageWrapper> {
+    private JFileChooser fc;
+    private ImageWrapper image;
+
+    public ImageOpener(){
+        putClientProperty("JButton.buttonType", "square");
+        addActionListener(e -> {
+            if (fc == null) {
+                fc = new JFileChooser(new File(".")) {
```
Pablete1234, line 22: you should probabily initialize the folder to be in the path of the current image wrapper, if it has a path

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImageOpener.java
```diff
@@ -0,0 +1,72 @@
...
+                    }
+                };
+                FileNameExtensionFilter filter = new FileNameExtensionFilter(
+                        "PNG, JPG & GIF Images", "png","jpeg","jpg", "gif");
+                fc.setFileFilter(filter);
+                fc.setAcceptAllFileFilterUsed(false);
+            }
+            if(image == null){
+                image = new ImageWrapper();
+            }
```
Pablete1234, line 37: You should probably do this in getEditorComponent directly, like `setEditing(val == null ? new img : val)`

src/main/java/com/github/manolo8/darkbot/gui/utils/UIUtils.java
```diff
@@ -20,12 +21,16 @@
...
 
     public static ImageIcon getIcon(String name) {
-        return new ImageIcon(new ImageIcon(UIUtils.class.getResource("/" + name + ".png")).getImage()
+        return new ImageIcon(new ImageIcon(Objects.requireNonNull(UIUtils.class.getResource("/" + name + ".png"))).getImage()
                 .getScaledInstance(16, 16, Image.SCALE_SMOOTH));
     }
 
     public static Image getImage(String name) {
-        return new ImageIcon(UIUtils.class.getResource("/" + name + ".png")).getImage();
+        return getImage(name,"png");
```
Pablete1234, line 29:
```suggestion
        return getImage(name, "png");
```

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -167,7 +172,11 @@ public void setup(Main main) {
     protected void paintComponent(Graphics g) {
         if (main == null) return;
         Graphics2D g2 = setupDraw(g);
-
+        if(main.config.BOT_SETTINGS.MAP_DISPLAY.CUSTOM_BACKGROUND && this.main.config.BOT_SETTINGS.MAP_DISPLAY.CHOOSER.getImage() != null){
+            g2.setComposite(AlphaComposite.getInstance(AlphaComposite.SRC_OVER, (float) main.config.BOT_SETTINGS.MAP_DISPLAY.TRANSPARENCY));
+            g2.drawImage(this.main.config.BOT_SETTINGS.MAP_DISPLAY.CHOOSER.getImage(),0,0,width,height,this);
```
Pablete1234, line 177:
```suggestion
            g2.drawImage(this.main.config.BOT_SETTINGS.MAP_DISPLAY.CHOOSER.getImage(), 0, 0, width, height, this);
```

src/main/java/com/github/manolo8/darkbot/gui/utils/UIUtils.java
```diff
@@ -7,6 +7,7 @@
 import javax.swing.border.Border;
 import javax.swing.border.MatteBorder;
 import java.awt.*;
+import java.util.Objects;
 
 public class UIUtils {
```
Pablete1234, line 12: are any of the changes in UIUtils really needed? can't see them used anywhere

Pablete1234 (changes requested):
You also (probably accidentally) left in a png in resources

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImageOpener.java
anjeshshrestha: theses are what i have tested with so far and seem to work, need to test with bmp but this covers the majority

src/main/java/com/github/manolo8/darkbot/gui/utils/UIUtils.java
anjeshshrestha: think it was auto import for something, cause of ide

anjeshshrestha: wait its in 

`public static ImageIcon getIcon(String name) {
        return new ImageIcon(new ImageIcon(Objects.requireNonNull(UIUtils.class.getResource("/" + name + ".png"))).getImage()
                .getScaledInstance(16, 16, Image.SCALE_SMOOTH));
    }`

Pablete1234: i mean the changes in the class as a whole, not the import

src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -29,6 +29,7 @@
 import eu.darkbot.shared.modules.LootCollectorModule;
 import org.jetbrains.annotations.Nullable;
 
+
```
Pablete1234, line 32:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -240,6 +241,12 @@
             public @Option boolean MAP_START_STOP = false;
             public @Option("colors") ColorScheme cs = new ColorScheme();
         }
+        public @Option CustomBackground CUSTOM_BACKGROUND = new CustomBackground();
```
Pablete1234, line 244:
```suggestion

        public @Option CustomBackground CUSTOM_BACKGROUND = new CustomBackground();
```

src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -240,6 +241,12 @@
...
+        public static class CustomBackground{
```
Pablete1234, line 245:
```suggestion
        public static class CustomBackground {
```

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -167,7 +170,11 @@ public void setup(Main main) {
     protected void paintComponent(Graphics g) {
         if (main == null) return;
         Graphics2D g2 = setupDraw(g);
-
+        if(main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.ENABLED && this.main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.IMAGE.getImage() != null){
```
Pablete1234, line 173:
```suggestion
        if (main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.ENABLED && this.main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.IMAGE.getImage() != null) {
```

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImagePicker.java
```diff
@@ -0,0 +1,76 @@
...
+            if (fc.showOpenDialog(null) == JFileChooser.APPROVE_OPTION) {
+                image.setPath(fc.getSelectedFile().getAbsolutePath());
+                setText(Strings.fileName(image.getPath()));
+            }
+        });
+
+    }
+    @Override
+    public void setText(String text) {
+        if(text != null && text.contains("\\") && text.length() > 20){
```
Pablete1234, line 46: use File.separator (or something like that, i'm not 100% sure on what it was called), do not hard-code `\`. On linux systems it will be `/`

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImagePicker.java
```diff
@@ -0,0 +1,76 @@
...
+            super.setText(".."+text.substring(text.lastIndexOf("\\", text.length()-20)));
```
Pablete1234, line 47: i would ensure the cutting point to always be a file separator, so just search for the file separator char that is closest to a target size (probably a longer on than 20 chars, that looks like a pretty small target, maybe 50 chars?)

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImagePicker.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public JComponent getEditorComponent(ConfigSetting<ImageWrapper> configSetting) {
+        setEditing(configSetting.getValue());
+        return this;
+    }
+
+    private void setEditing(ImageWrapper image){
+        this.image = image == null ? new ImageWrapper() : image; //do i need this, i dont think its needed
```
Pablete1234, line 62: you do need it, otherwise you won't be able to return a non-null image in getEditorValue if this was null from the start

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImagePicker.java
```diff
@@ -0,0 +1,76 @@
...
+                };
+                FileNameExtensionFilter filter = new FileNameExtensionFilter(
+                        "Images Files", "png", "jpeg", "jpg", "gif");
+                fc.setFileFilter(filter);
+                fc.setAcceptAllFileFilterUsed(false);
+            }
+
+            if (fc.showOpenDialog(null) == JFileChooser.APPROVE_OPTION) {
+                image.setPath(fc.getSelectedFile().getAbsolutePath());
+                setText(Strings.fileName(image.getPath()));
```
Pablete1234, line 39: you should probably just call `setEditing(image)` and let that handle the renaming

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImagePicker.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.swing.filechooser.FileNameExtensionFilter;
+import java.awt.*;
+import java.io.File;
+
+public class ImagePicker extends JButton implements OptionEditor<ImageWrapper> {
+    private JFileChooser fc;
+    private ImageWrapper image;
+
+    public ImagePicker(){
+        putClientProperty("JButton.buttonType", "square");
```
Pablete1234, line 19: i could be wrong but i *think* this may no longer be needed, we enforce square-ness on a higher level already, try to remove it

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImagePicker.java
anjeshshrestha: 50 char might be too much, shows the whole directory from c: if the image is in downloads

Pablete1234: just checked and yeah, 50 as a target may be too much, maybe within the range of 20-30 should be fine

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImagePicker.java
```diff
@@ -0,0 +1,75 @@
...
+                image.setPath(fc.getSelectedFile().getAbsolutePath());
+                setEditing(image);
+            }
+        });
+
+    }
+    @Override
+    public void setText(String text) {
+        if(text != null && text.contains(File.separator) && text.length() > 30){
+            super.setText(".."+text.substring(text.lastIndexOf(File.separator, text.length()-30)));
```
Pablete1234, line 46: lastIndexOf can return -1
lastIndexOf will also try to cut at the *last* file separator, instead of the *first* file separator (starting from length - 30)

The idea is to find the closest file separator to len-30 and cut there

Also the formatting, this methid is missing many spaces

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImagePicker.java
```diff
@@ -0,0 +1,80 @@
...
+                image.setPath(fc.getSelectedFile().getAbsolutePath());
+                setEditing(image);
+            }
+        });
+
+    }
+
+    @Override
+    public void setText(String text) {
+        int separatorIndex = text.lastIndexOf(File.separator, text.length() - 30);
```
Pablete1234, line 49: You need to search for indexOf, which returns the first occurrence (after len - 30). Using lastIndexOf you would always cut at the *last* separator, leaving just the filename

Also len - 30 can be negative, i don't think it's correct to call `String#indexOf(..., -10)`, so you need to ensure this is only called if string is above 30 chars (what you had before)

anjeshshrestha: there is two options i can use in the if statement after `text != null && text.length() > 30 &&`
`text.indexOf(File.separator, text.length() - 30) != -1`
or 
`text.substring(text.length() - 30).contains(File.separator)`

would either one of theses be a good idea?

Pablete1234: the implementation of contains is often implemented as indexOf != -1, at least in collections

I'd go for indexOf approach because contains requires you to substring, also, you can reuse the indexof result:
```
int sepIdx;
if (... && len > 30 && (sepIdx = text.indexOf(..., len - 30)) != -1) {
    substring(sepIdx, len)
}
```

Pablete1234:
The PR needs to be properly rebased, atm it's including changes all accross the codebase that belong to previous commits in master not this pr

src/main/java/com/github/manolo8/darkbot/config/ImageWrapper.java
```diff
@@ -0,0 +1,33 @@
...
+    public String getPath() {
+        if (this.path == null)
+            return "None";
+        return this.path;
+    }
+
+    public void setPath(String path) {
+        this.path = path;
+        this.loaded = false;
+    }
```
Pablete1234, line 20: i believe gson will serialize it by using these 2 methods, which means an empty image will serialize as the string "None"

I think getPath should return null if it is indeed null (handle displaying a "None" on the editor, not here)

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #120, opened 2022-01-07
Implement Npc api entity setters

pajtor:
Not sure about setAmmo and setFormation


src/main/java/com/github/manolo8/darkbot/core/entities/Npc.java
```diff
@@ -76,16 +76,31 @@ public NpcInfoImpl(HeroItemsAPI items, SettingsProxy settingsProxy) {
             this.items = items;
             this.settingsProxy = settingsProxy;
         }
+        
+        @Override
+        public boolean setShouldKill(boolean shouldKill) {
```
Pablete1234, line 81: setters should be void, not return the value being set
```suggestion
        public void setShouldKill(boolean shouldKill) {
```
if in the api it was defined like this, it was wrong and should be changed

Pablete1234:
This needs to be rebased ontop of api
```
git fetch origin
git rebase origin/api
```
then force-push to here

PR #122, opened 2022-01-08
Add option to use username as title

ODAKAB:
Reason: To make each instance easily distinguishable among others e.g. when task is unresponsive and you want to terminate process.

Concern: Weird usernames might conflict with Windows OS?


Pablete1234 (changes requested):
Override setTitle to avoiding setting the title each time, because most of the times it's going to be set to the same, and called every tick
```
@Override
void setTitle(String title) {
    if (!Objects.equals(getTitle(), title)) super.setTitle(title);
}
```

Pablete1234 (approved):

Pablete1234:
The pr has conflicts that need to be solved before being able to merge it

PR #123, opened 2022-01-09
implemented reset stats & is instance valid

Punikekk:


src/main/java/com/github/manolo8/darkbot/backpage/BackpageManager.java
```diff
@@ -256,6 +258,15 @@ private String sidStat() {
         }
     }
 
+    @Override
+    public boolean isInstanceValid() {
+        // in isInvalid() method, checking sid & instance from stats manager are not synchronised
+        // so even if #isInvalid() returns false the connection can throw exception anyway
+        // instead local sid & instance will be updated once before any task tick
+        // or even better, keep old vales if new are empty, where old ones can still be valid
```
Pablete1234, line 266:
```suggestion
        // Only check against local sid & instance variables, since stats manager ones are
        // updated in the main thread, while the variables here are updated on the background
        // thread every tick
```
I'd simplify this comment, it was phrased a bit weird

src/main/java/com/github/manolo8/darkbot/backpage/BackpageManager.java
```diff
@@ -174,7 +174,9 @@ public HttpURLConnection getConnection(String path, int minWait) throws Exceptio
     }
 
     public HttpURLConnection getConnection(String params) throws Exception {
-        if (isInvalid()) throw new UnsupportedOperationException("Can't connect when sid is invalid");
+        // basically there's no need for that check, if sid is updates once before any task tick anyway,
+        // until this method is called from another thread
```
Pablete1234, line 178:
```suggestion
```

i'm pretty sure its obvious enough, doesn't need the comment

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #127, opened 2022-01-16
Changed "-1" to '∞'

QR4SH3D:
Infinity symbol


QR4SH3D:
Was not able to test and after I was able to test it didn't work

ODAKAB:
Hi, you should modify the visual part on MapDrawer
You can take my rejected pull request as an example.

PR #132, opened 2022-02-02
add display of nearby players inside Players tab

fabio1999ita:


src/main/java/com/github/manolo8/darkbot/gui/nearplayers/NearPlayerEditor.java
```diff
@@ -0,0 +1,100 @@
...
+            //TODO: add save function to player tag list
+        });
+        refreshList();
+    }
+
+
+    public void refreshList() {
+        Thread t1 = new Thread(() -> {
+            while (true) {
+                if (main != null) {
```
Pablete1234, line 59: You're creating a few threading issues here
 - The entity updates must only be accessed on the main thread, or if accessed from other threads, they must use synchronized access, otherwise you'll get concurrent modification exceptions and your thread will crash.
  - You're modifying swing components from outside the swing ui thread, that can cause inconsistencies in the models and break or deadlock the UI
  
The solution is to make use of entity events (launched from the main thread) and using `SwingUtilities.invokeLater(() -> { your swing code });` for any ui-editing code (addElement, or remove, etc)

src/main/java/com/github/manolo8/darkbot/gui/nearplayers/NearPlayerRenderer.java
```diff
@@ -0,0 +1,87 @@
...
+
+import javax.swing.*;
+import javax.swing.border.Border;
+import javax.swing.border.CompoundBorder;
+import javax.swing.border.EmptyBorder;
+import java.awt.*;
+import java.util.HashMap;
+import java.util.Map;
+
+public class NearPlayerRenderer extends JPanel implements ListCellRenderer<PlayerInfo> {
```
Pablete1234, line 16: I'm pretty sure you could just extend the default renderer for this one, you're not displaying tags or anything too complex in this list, rewriting an impl is a bit overkill, you could simply display the username on this, no need for the id since it's all ephermal

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,71 @@
...
 
     private final PlayerManager playerManager;
     private final SearchField sf;
     private final PlayerTagManager tagEditor;
     protected Main main;
 
+    @EventHandler
+    public void onEntityPlayerAdd(EntitiesAPI.EntityCreateEvent event) {
+        if (event.getEntity() instanceof Player) {
+            if (playersModel != null && main.config.PLAYER_INFOS != null && main.config.PLAYER_INFOS.values()
```
Pablete1234, line 35: The 2 ifs can just be joined into 1

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,71 @@
...
+                    .stream().noneMatch(a -> a.toString()
+                            .contains(((Player) event.getEntity()).getEntityInfo().getUsername()))) {
```
Pablete1234, line 37: Never use a "to string" to match things like this.
PLAYER_INFOS is a map indexed by the id, just use that to check:
`main.config.PLAYER_INFOS.hasKey(event.getEntity().getId())`

the stream, the o(n) complexity, and the string memory allocations, are not needed, when you can just do it in o(1)

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,71 @@
...
+                nearbyPlayerList.add(new PlayerInfo(((Player) event.getEntity())
```
Pablete1234, line 38: You're manipulating the nearbyPlayerList from both the main thread, and the swing ui thread (in `refreshList`)

You are also creating 2 different player infos, one in the element list and another in the nearby player list.

Move this inside the invokeLater and just run it all in the swing event thread, and create a single instance of PlayerInfo

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,71 @@
...
+                        .getEntityInfo().getUsername(), event.getEntity().getId()));
+
+                SwingUtilities.invokeLater(() -> playersModel.
+                        addElement(new PlayerInfo(((Player) event.getEntity())
+                                .getEntityInfo().getUsername(), event.getEntity().getId())));
```
Pablete1234, line 43: Just add a constructor to PlayerInfo that takes in a Player as parameter to simplify this

```suggestion
                        addElement(new PlayerInfo((Player) event.getEntity()));
```

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,71 @@
...
+                System.out.println("added " + ((Player) event.getEntity()).getEntityInfo().getUsername());
+            } else {
+                System.out.println("in list yet " + ((Player) event.getEntity()).getEntityInfo().getUsername());
+            }
```
Pablete1234, line 47: These prints are not needed anymore

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,71 @@
...
+        }
+    }
+
+    @EventHandler
+    public void onEntityPlayerRemove(EntitiesAPI.EntityRemoveEvent event) {
+        if (event.getEntity() instanceof Player) {
+            if (playersModel != null && main.config.PLAYER_INFOS != null && main.config.PLAYER_INFOS.values()
```
Pablete1234, line 54: Similarly, the ifs can be merged

You could also simplify this by using a return if the instance isn't a player (both here and in the method above):
`if (!(event.getEntity() instanceof Player)) return;`
that allows you to un-indent everything else

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,71 @@
...
+                    .stream()
+                    .noneMatch(a -> a.toString()
+                            .contains(((Player) event.getEntity()).getEntityInfo().getUsername()))) {
```
Pablete1234, line 57: Again, you should just check against the id as the hashmap key, not using a string-check in a stream

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,71 @@
...
+                nearbyPlayerList.stream()
+                        .filter(a -> a.toString().contains(((Player) event.getEntity()).getEntityInfo().getUsername()))
+                        .collect(toList()).forEach(nearbyPlayerList::remove);
```
Pablete1234, line 60: You're iterating the player list, converting to string, checking for contains which could return multiple results, and removing all of them. The code is wrong in several ways:

 - If there's a player named `PlayerABCDE` and another player named `Player`, when `Player` gets out of sight, both are removed.
 - You're converting a PlayerInfo to string to check for the username, that to string contains even the tag because it's used for filtering, and regardless, you should never rely on toString of an object for your logic, as that is prone to changing.
 - You're collecting all results to a list, causing another list allocation
 
 You can just use removeIf:
 ```suggestion
                nearbyPlayerList.removeIf(pl -> pl.getId() == event.getEntity().getId());
```

Keep in mind however, none of this is relevant because you should be doing this in the swing event thread to avoid race conditions on nearbyPlayerList, as i stated in player add. See comment below for how to easily do it in the swing event thread.

File: src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java [hunk 11]
```diff
@@ -4,29 +4,71 @@
...
+                
+                SwingUtilities.invokeLater(() -> Arrays.stream(playersModel.toArray())
+                        .filter(a -> a.toString().contains(((Player) event.getEntity()).getEntityInfo().getUsername()))
+                        .forEach(playersModel::removeElement));
```
Pablete1234, line 64: Same issues as above, but you're also allocating a new array for the whole model, including both nearby and non-nearby players.

Iterate the model to find the object,, and call removeElement for just that, also, update the nearby player list in the swing thread to avoid race conditions:

```suggestion
                SwingUtilities.invokeLater(() -> {
                        for (int i = 0; i < playersModel.size(); i++) {
                            PlayerInfo info = playersModel.getElement(i);
                            if (info.getId() == event.getEntity().getId()) {
                                nearbyPlayerList.remove(info);
                                playersModel.removeElement(info);
                                break;
                            }
                        }
                });
```
Keep in mind this will only work if the player info you use in both lists is the same, which you SHOULD do (see the comment on player add)

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,71 @@
...
+                System.out.println("removed " + ((Player) event.getEntity()).getEntityInfo().getUsername());
+            } else {
+                System.out.println("in list yet " + ((Player) event.getEntity()).getEntityInfo().getUsername());
+            }
```
Pablete1234, line 68: These are no longer needed

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -64,6 +109,10 @@ public void refreshList(String filter) {
                 .filter(pi -> pi.filter(query))
                 .sorted(Comparator.comparing(pi -> pi.username))
                 .forEach(playersModel::addElement);
+        if (nearbyPlayerList != null)
```
Pablete1234, line 112: This can never be false. nearbyPlayerList is on the initializer and is final. you can just remove the if

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -46,6 +85,9 @@ public void setup(Main main) {
 
         tagEditor.setup(main);
 
+        EventBrokerAPI eventBroker = main.pluginAPI.requireAPI(EventBrokerAPI.class);
+        eventBroker.registerListener(this);
```
Pablete1234, line 89: doing this on the setup method, means it will be called every time this is re-setup (this is done when changing config profile).
this means you will be registering the listener multiple times with a new dispatcher each time

you must take care of this

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -39,13 +98,53 @@ public PlayerEditor() {
...
         add(new JScrollPane(playerInfoList,
                 JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED, JScrollPane.HORIZONTAL_SCROLLBAR_NEVER), "span, grow");
+
+        /*playerInfoList.addListSelectionListener(event -> {
+            if (!event.getValueIsAdjusting()) {
+
+            }
+        });*/
+
+        playerInfoList.addMouseListener(new MouseAdapter() {
```
Pablete1234, line 108: Mouse listener is NOT the tool for the task, you can select and unselect players using the keyboard, and all of this logic will essentially be broken

Please use a listener for the selection model on the list

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,86 @@
...
     private final PlayerManager playerManager;
     private final SearchField sf;
     private final PlayerTagManager tagEditor;
     protected Main main;
 
+    public int indexOfPlayersModel(int playerId) {
+        for (int i = 0; i < playersModel.size(); i++)
+            if (playerId == playersModel.get(i).getUserId()) return i;
+        return -1;
+    }
```
Pablete1234, line 38: This should be under the playeradd/remove events, generally you want to order methods from more relevant to less relevant, the "util" methods at the bottom (like this one is).

The constructor should also be above all these new methods

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -4,29 +4,86 @@
...
+                        if (!playerInfoList.isSelectedIndex(i)) {
+                            nearbyPlayerList.remove(info);
+                            playersModel.removeElement(info);
+                            break;
+                        } else {
+                            nearbyPlayerList.remove(info);
+                            break;
+                        }
+                    }
+                }
```
Pablete1234, line 82: What's wrong with the first (commented-out) version?
You shouldn't need to do a loop here, you have the indexOf version to find the element directly.

You will simply need to edit it a bit so that you have the logic of removing from nearbyPlayerList even if selected

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -39,13 +96,33 @@ public PlayerEditor() {
 
         add(new JScrollPane(playerInfoList,
                 JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED, JScrollPane.HORIZONTAL_SCROLLBAR_NEVER), "span, grow");
+
+        playerInfoList.addListSelectionListener(event -> SwingUtilities.invokeLater(() -> {
+            if (!event.getValueIsAdjusting()) {
```
Pablete1234, line 101:
```suggestion
            if (event.getValueIsAdjusting()) return;
```
You can simplify these all-encompasing ifs with just returning on the oposite case. That way you avoid indenting-in the whole method

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -39,13 +96,33 @@ public PlayerEditor() {
 
         add(new JScrollPane(playerInfoList,
                 JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED, JScrollPane.HORIZONTAL_SCROLLBAR_NEVER), "span, grow");
+
+        playerInfoList.addListSelectionListener(event -> SwingUtilities.invokeLater(() -> {
```
Pablete1234, line 100: The event occurs on the swing thread, you DO NOT and SHOULD NOT use an invoke later for this, you want to do this synchronously when the selection changes

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -39,13 +96,33 @@ public PlayerEditor() {
...
+                    if (!nearbyPlayerList.contains(playersModel.get(i + main.config.PLAYER_INFOS.size()))) {
+                        if (!playerInfoList.getSelectedValuesList().contains(playersModel.get(i + main.config.PLAYER_INFOS.size()))) {
+                            listToRemove.add(i + main.config.PLAYER_INFOS.size());
+                        }
+                    }
+                }
+                Collections.reverse(listToRemove);
+                for (int p : listToRemove) {
+                    playersModel.remove(p);
+                }
```
Pablete1234, line 113: This is one very convoluted way of removing elements. I think the list selection change listener has a start and end interval change, or something similar, you only need to iterate those elements, as long as they belong to the "nearby player list".

https://docs.oracle.com/javase/tutorial/uiswing/events/listselectionlistener.html

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerRenderer.java
```diff
@@ -15,7 +15,7 @@
 
 public class PlayerRenderer extends JPanel implements ListCellRenderer<PlayerInfo> {
 
-    private final JLabel playername = new JLabel();
+    private final JLabel playerName = new JLabel();
```
Pablete1234, line 18:
```suggestion
    private final JLabel playername = new JLabel();
```
keep this as-is, playername is used as a single word in this case, just like username

Pablete1234 (changes requested):

Pablete1234 (changes requested):

fabio1999ita:
video
problem was that

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerManager.java
```diff
@@ -58,6 +60,21 @@ private void addPlayerByUsername() {
             if (name == null) return;
             editor.main.config.UNRESOLVED.add(new UnresolvedPlayer(name));
         }
+
+        private void addSelectedUserId() {
+            assert false;
```
Pablete1234, line 65:
```suggestion
```

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -22,10 +22,10 @@
 import java.util.Set;
 
 public class PlayerEditor extends JPanel implements Listener {
-    private final JList<PlayerInfo> playerInfoList;
+    final JList<PlayerInfo> playerInfoList;
     private final DefaultListModel<PlayerInfo> playersModel;
     private final PlayerRenderer renderer;
-    private final Set<PlayerInfo> nearbyPlayers = new LinkedHashSet<>();
+    final Set<PlayerInfo> nearbyPlayers = new LinkedHashSet<>();
```
Pablete1234, line 28:
```suggestion
    private final Set<PlayerInfo> nearbyPlayers = new LinkedHashSet<>();
```
Keep it private, if you need to access it add a getter

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -115,6 +115,7 @@ public void onEntityPlayerAdd(EntitiesAPI.EntityCreateEvent event) {
                     PlayerInfo info = playersModel.get(idx);
                     nearbyPlayers.add(info);
                 }
+                refreshList();
```
Pablete1234, line 118: if you call refreshList here, there's no longer a point to calling addElement at all, since refreshList would already re-create the whole thing.

The point is you cannot do refreshList because that method does not calculate the difference (remove the uneeded, and add the new ones) instead it wipes the whole list and re-creates it, meaning any selection is lost.

One of my original points was you could rewrite refreshList to be able to smartly add/remove things from the list instead of wiping it and re-creating, and that would simplify the rest of your code as you would only need to add or remove from nearbyPlayers and call refreshList, without bothering about any other adding/removing logic and avoiding duplicates

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerManager.java
```diff
@@ -58,6 +60,20 @@ private void addPlayerByUsername() {
             if (name == null) return;
             editor.main.config.UNRESOLVED.add(new UnresolvedPlayer(name));
         }
+
+        private void addSelectedPlayer() {
+            if (editor.getPlayerInfoList().getSelectedValuesList().isEmpty()) {
```
Pablete1234, line 65:
```suggestion
            if (editor.getPlayerInfoList().isSelectionEmpty()) {
```

Avoids having to create the list of values when you can just check for empty

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerRenderer.java
```diff
@@ -18,13 +19,23 @@
...
+    private int addedPlayerCount = -1;
+
+    public PlayerRenderer(JList<? extends PlayerInfo> list) {
         super(new MigLayout("ins 4px 0px 4px 5px, fill, h 28px!", "[50px!]8px![120px!]8px:push[]8px!", "[]"));
         id.setFont(id.getFont().deriveFont(9f));
         id.setHorizontalAlignment(SwingConstants.RIGHT);
         id.setVerticalAlignment(SwingConstants.BOTTOM);
+
+        this.defaultBorder = this.getBorder();
+        this.titledBorder = new ListCellTitledBorder(list, "Nearby Players");
```
Pablete1234, line 34: This should come from I18n too

Pablete1234 (changes requested):
Almost good to merge, just a few minor changes

src/main/resources/lang/strings_en.properties
```diff
@@ -305,12 +305,16 @@ safety_places.cbs_mode.ally_neutral=Ally or empty
...
+players.add_player.selected=Selected player
 players.add_player.username=Username:
 players.add_player.user_id=User ID:
 players.add_player.invalid_id=Invalid user ID
 players.add_player.invalid_id.no_number={0} is not a valid number
+players.add_player.selected.invalid_selection=Invalid selection
+players.add_player.selected.invalid_selection.no_selected=You need to select nearby player(s) to add
 players.remove_player.title=Are you sure?
 players.remove_player.msg=Do you want to delete the selected {0} players?
+players.title_separator="Nearby Players"
```
Pablete1234, line 317:
```suggestion
players.title_separator=Nearby Players
```

Pablete1234 (approved):

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -35,17 +41,38 @@ public PlayerEditor() {
...
         playerInfoList = new JList<>(playersModel = new DefaultListModel<>());
-        playerInfoList.setCellRenderer(new PlayerRenderer());
+        playerInfoList.setCellRenderer(renderer = new PlayerRenderer(playerInfoList));
 
         add(new JScrollPane(playerInfoList,
                 JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED, JScrollPane.HORIZONTAL_SCROLLBAR_NEVER), "span, grow");
+
+        playerInfoList.addListSelectionListener((event) -> {
+            if (event.getValueIsAdjusting()) return;
+            int min = Math.max(event.getFirstIndex(), main.config.PLAYER_INFOS.size());
```
Pablete1234, line 51:
```suggestion
            int min = event.getFirstIndex();
```
We can't assume the first N entries (player info size) won't be temporal players, because you could be running a filtered search in which case they wouldn't count for that. They're already checked a bit lower down so it's not an issue

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -64,6 +91,59 @@ public void refreshList(String filter) {
...
+
+            SwingUtilities.invokeLater(() -> {
+                int idx = indexOfPlayersModel(event.getEntity().getId());
+                if (idx < 0) {
+                    PlayerInfo playerInfo = new PlayerInfo((Player) event.getEntity());
+                    nearbyPlayers.add(playerInfo);
+                    playersModel.addElement(playerInfo);
+                } else {
+                    PlayerInfo info = playersModel.get(idx);
+                    nearbyPlayers.add(info);
```
Pablete1234, line 115:
```suggestion
                    nearbyPlayers.add(playersModel.get(idx));
```
Slight cleanup, these can be merged

src/main/java/com/github/manolo8/darkbot/gui/players/PlayerEditor.java
```diff
@@ -64,6 +91,59 @@ public void refreshList(String filter) {
...
+                int idx = indexOfPlayersModel(event.getEntity().getId());
+                if (idx != -1) {
+                    PlayerInfo info = playersModel.getElementAt(idx);
+                    if (!playerInfoList.isSelectedIndex(idx)) {
+                        nearbyPlayers.remove(info);
+                        playersModel.removeElement(info);
+                    } else {
+                        nearbyPlayers.remove(info);
+                    }
+                }
```
Pablete1234, line 138:
```suggestion
                if (idx == -1) return;
                PlayerInfo info = playersModel.getElementAt(idx);
                if (!playerInfoList.isSelectedIndex(idx)) {
                    nearbyPlayers.remove(info);
                    playersModel.removeElement(info);
                } else {
                    nearbyPlayers.remove(info);
                }
```
Less indenting makes the code easier to follow

PR #137, opened 2022-02-17
add support for 3D mode

Punikekk:


Punikekk:
Seems like after reload `force2D` returns 0, but game is in 2D mode without enabled option `Force HW accel`
For some reason, game after reload, doesnt detect directX.
Need to fix it somehow

src/main/java/com/github/manolo8/darkbot/core/BotInstaller.java
```diff
@@ -93,7 +93,7 @@ private boolean tryInstall() {
         BotInstaller.SEP = API.readMemoryInt(mainApplicationAddress.get() + 4);
 
         if ((query = API.queryMemory(bytesToSettings, 1)).length != 1) return true;
-        this.settingsAddress.send(query[0] - 237);
+        this.settingsAddress.send(query[0] - 277);
```
Pablete1234, line 96: Is this change not going to break anything else (eg: in 2d mode)?

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -125,10 +125,12 @@ public void update() {
 
         if (targetPtr == 0) inGameTarget = null;
         else if (targetPtr == petAddress) inGameTarget = pet;
-        else inGameTarget = main.mapManager.entities.allEntities.stream()
-                .flatMap(Collection::stream)
-                .filter(entity -> entity.address == targetPtr)
-                .findAny().orElse(null);
+        else if (inGameTarget == null
+                 || inGameTarget.address != targetPtr)
```
Pablete1234, line 129:
```suggestion
        else if (inGameTarget == null || inGameTarget.address != targetPtr)
```

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -209,14 +211,23 @@ public boolean setMode(int con, Character form) {
 //        }
 
         Configuration conf = Configuration.of(con);
-        if (conf == Configuration.UNKNOWN) return false; // unknown config? return, as we can't select unknown configuration.
+        if (conf == Configuration.UNKNOWN)
+            return false; // unknown config? return, as we can't select unknown configuration.
```
Pablete1234, line 215:
```suggestion
        if (conf == Configuration.UNKNOWN) return false; // unknown config? return, as we can't select unknown configuration.
```

src/main/java/com/github/manolo8/darkbot/core/objects/Clickable.java
```diff
@@ -40,7 +47,8 @@ public void reset() {
      * @return prevent swf crash
      */
     private boolean isInvalid() {
-        return defRadius <= 0 || address == 0 || API.readMemoryLong(address) != confirm;
+        return address == 0
+               || API.readLong(address + 32) != owner.address; // confirm owner address
```
Pablete1234, line 51:
```suggestion
        return defRadius <= 0 || address == 0 || API.readLong(address + 32) != owner.address; // confirm owner address
```
In this case, even if redundant, i would rather keep the def radius check. I've seen too many weird things happen with this logic and i don't want random crashes to occur for something so easy to keep

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -460,9 +475,25 @@ private void drawHero(Graphics2D g2) {
...
+
+        g2.drawLine(translateX(view.rightBotX), translateY(view.rightBotY),
+                translateX(view.leftBotX), translateY(view.leftBotY));
+
+        g2.drawLine(translateX(view.leftBotX), translateY(view.leftBotY),
+                translateX(view.leftTopX), translateY(view.leftTopY));
+
+//        g2.drawRect(translateX(mapManager.boundX), translateY(mapManager.boundY),
+//                translateX(mapManager.boundMaxX - mapManager.boundX),
+//                translateY(mapManager.boundMaxY - mapManager.boundY));
```
Pablete1234, line 496:
```suggestion
```

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/BotInstaller.java
Punikekk: Just moved settings address class offset (-40) everywhere, now it points exactly to `Settings` class instance

src/main/java/com/github/manolo8/darkbot/core/objects/Clickable.java
Punikekk: Sometimes, no idea why, the entities(mostly boxes) have `defRadius == 0`, and bot is unable to select it cause this method returns true

Pablete1234: precisely, if the radius is 0 for odd reasons we don't want to mess with them, gotta find where the real bug is and fix it

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -284,7 +267,19 @@ public Configuration getConfiguration() {
 
     @Override
     public boolean isInMode(ShipMode mode) {
-        return mode.getConfiguration() == getConfiguration() && mode.getFormation() == getFormation();
+        return mode.getConfiguration() == getConfiguration() && (mode.getFormation() == getFormation()
+                || (mode.getFormation() == null && mode instanceof MutableShipMode
+                    && ((MutableShipMode) mode).toSelectChar == selectedChar));
```
Pablete1234, line 272: Either MutableShipMode, or Config.ShipMode, both of those have a char to check against

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -296,7 +291,13 @@ public boolean setMode(@NotNull ShipMode mode) {
     private void setConfigAndFormation(ShipMode mode) {
         if (mode.getConfiguration() != null &&
                 mode.getConfiguration() != getConfiguration()) toggleConfiguration();
-        setFormation(mode.getFormation());
+
+        SelectableItem.Formation formation = mode.getFormation();
+
+        if (formation == null && mode instanceof MutableShipMode) //todo remove me later
```
Pablete1234, line 297: both MutableShipMode, or Config.ShipMode, both have a formation to set the legacy way

also, the MutableShipMode should only do the legacy way *if* the set ship mode that it was set to is a legacy one, if it was a modern ship mode it should be null and this codepath should not be used

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -367,11 +370,13 @@ public Configuration getConfiguration() {
         public void set(ShipMode other) {
             this.configuration = other.getConfiguration();
             this.formation = other.getFormation();
+            this.toSelectChar = null;
```
Pablete1234, line 373: make it depend on `other instanceof Config.ShipMode`, if it is, then set the char to what it has to be, otherwise null

Pablete1234 (changes requested):
You may be better off creating a `interface LegacyShipMode extends ShipMode` (in darkbot directly) that both MutableShipMode and Config.ShipMode implement, and that adds a `Charater getLegacyFormation()` function.
If getLegacyFormation is not null, it will use that to compare what formation is set, and to set formation based on it.
If it is null then it will use the new codepaths

Pablete1234 (approved):

PR #138, opened 2022-03-05
Implement DrawableHandler

Punikekk:


src/main/java/com/github/manolo8/darkbot/gui/zones/safety/SafetiesDisplay.java
```diff
@@ -3,77 +3,123 @@
...
-            drawStaticEntities(g2);
-            drawMap(g2);
-            if (editor.safetyInfos == null) return;
-            drawCustomZones(g2);
-        }
+    public void setup(Main main) {
+        super.setup(main);
+
+        this.safetiesDisplay = new SafetiesDisplayDrawer<>(main.pluginAPI, this);
+        this.safetiesDisplay.setup(editor);
```
Pablete1234, line 47: This class was already intentionally a very dummy class, that only relied on what the editor told it, setup shouldn't need to create its own drawer or do anything like that, simply leave it as it was, override the draw method from parent and instead of just drawing all layers, draw only what we want (zones, static entities, map name, and custom zones) by calling the different drawables required, and that's it

i don't understand the need for a SafetiesDisplayDrawer, you can simply draw ontop yourself directly in the paintComponent method

src/main/java/com/github/manolo8/darkbot/gui/zones/ZoneEditor.java
```diff
@@ -85,88 +60,145 @@ public void mouseMoved(MouseEvent e) {
...
     }
 
     @Override
-    protected void paintComponent(Graphics g) {
-        Graphics2D g2 = setupDraw(g);
+    protected void onPaint() {
+        zonesDrawer.onDraw(mapGraphics);
+    }
+
+    public static class ZoneEditorDrawer extends SafetiesDisplay.SafetiesDisplayDrawer<ZoneEditor> implements Listener {
```
Pablete1234, line 71: I've explained it more in safeties display, but i don't understand the need to create an inner-class for these editor drawers, they simply act as a way to display the map on their own, they don't need to create a drawer because the order is hard-coded and we don't plan on supporting anything rendering in between, these editors stay as-is.

Also, it seems like a very weird way to do this for ZoneEditor to extend SafetiesDisplay, what does one have to do with the other semantically? they shouldn't inherit eachother

src/main/java/com/github/manolo8/darkbot/gui/zones/safety/SafetiesDisplay.java
```diff
@@ -3,77 +3,123 @@
...
+import eu.darkbot.api.managers.HeroAPI;
+import eu.darkbot.api.managers.StarSystemAPI;
 
-import java.awt.*;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.util.Comparator;
 
-class SafetiesDisplay extends MapDrawer {
+public class SafetiesDisplay extends MapDrawer {
```
Pablete1234, line 21: this was intentionally package-private, we don't want this to be exposed outside

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/drawables/StatsDrawer.java
```diff
@@ -0,0 +1,161 @@
...
+
+        for (T render : toRender) {
+            renderer.render(left + 4, top, width - 8, render);
+            top += lineHeight;
+        }
+    }
+
+    private boolean hasDisplayFlag(DisplayFlag displayFlag) {
+        return displayFlags.getValue().contains(displayFlag);
+    }
```
Pablete1234, line 156: this should be a util exposed as part of MapGraphics

src/main/java/com/github/manolo8/darkbot/gui/drawables/InfosDrawer.java
```diff
@@ -0,0 +1,195 @@
...
+    private void drawPetFuel(MapGraphics mg, Point pos, int width, int height, double fuelPercent) {
+        mg.setColor(mg.getColor("fuel").darker());
+        mg.drawRect(pos, width, height, true);
+        mg.setColor("fuel");
+        mg.drawRect(pos, (int) (fuelPercent * width), height, true);
+    }
+
+    private boolean hasDisplayFlag(DisplayFlag displayFlag) {
+        return displayFlags.getValue().contains(displayFlag);
+    }
```
Pablete1234, line 194: Should be part of map graphics

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/utils/pathfinder/PolygonImpl.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
+public class PolygonImpl extends AreaImpl implements Area.Polygon {
+
+    private final List<? extends Locatable> vertices;
+    private final RectangleImpl bounds = new RectangleImpl();
+
+    private boolean invalidate = true;
```
Pablete1234, line 17: `invalidate` is a verb, `invalid` or `dirty` are nouns

variables should be nouns, while methods should be verbs
```suggestion
    private boolean invalid = true;
```

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -122,613 +50,172 @@ public MapDrawer(Main main) {
...
-        mid = width / 2;
-
-        g.setColor(cs.BACKGROUND);
-        g.fillRect(0, 0, width, height);
-
-        Graphics2D g2 = (Graphics2D) g.create();
-
-        g2.setRenderingHints(hints);
-        return g2;
+        //mapGraphics.dispose();
```
Pablete1234, line 102:
```suggestion
        mapGraphics.dispose();
```

Keep the call to mapGraphics.dispose(), even if it ends up being an empty method that does nothing
i think it's reasonable to assume we may need to free some resources after drawing, so we can leave the call there

src/main/java/com/github/manolo8/darkbot/gui/drawables/ZonesDrawer.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public void drawCustomZonePath(MapGraphics mg, ZoneInfo zoneInfo) {
+        if (!(zoneInfo instanceof com.github.manolo8.darkbot.config.ZoneInfo)) return;
+
+        List<com.github.manolo8.darkbot.config.ZoneInfo.Zone> sortedZones =
+                ((com.github.manolo8.darkbot.config.ZoneInfo) zoneInfo).getSortedZones();
+
+        Area.Rectangle mapBounds = starSystem.getCurrentMapBounds();
+
+        //convert it to actual list of locatables?
```
Pablete1234, line 111: probably best to do that, yeah
just add them all to a list and draw it as a poly

src/main/java/com/github/manolo8/darkbot/gui/zones/ZoneEditor.java
```diff
@@ -85,60 +66,52 @@ public void mouseMoved(MouseEvent e) {
...
     void setup(Main main, java.util.Map<Integer, ZoneInfo> zonesByMap) {
         super.setup(main);
-        if (mapChange != null) main.mapManager.mapChange.remove(mapChange);
-        main.mapManager.mapChange.add(mapChange = map -> {
-            zoneInfo = zonesByMap.computeIfAbsent(map.id, id -> new ZoneInfo(config.BOT_SETTINGS.OTHER.ZONE_RESOLUTION));
-            ZoneEditor.this.repaint();
-        });
-        zoneInfo = zonesByMap.get(main.hero.map.id);
+
+        main.pluginAPI.requireAPI(EventBrokerAPI.class).registerListener(this);
```
Pablete1234, line 70: if setup is called multiple times, you will be registering to the event broker multiple times, you have to prevent that

src/main/java/com/github/manolo8/darkbot/gui/zones/ZoneEditor.java
```diff
@@ -85,60 +66,52 @@ public void mouseMoved(MouseEvent e) {
...
+        zonesDrawer.drawZones(mapGraphics);
 
-        synchronized (Main.UPDATE_LOCKER) {
-            drawZones(g2);
-            drawStaticEntities(g2);
-            drawMap(g2);
-        }
+        constantEntitiesDrawer.drawPortals(mapGraphics);
+        constantEntitiesDrawer.drawBattleStations(mapGraphics);
+        constantEntitiesDrawer.drawStations(mapGraphics);
```
Pablete1234, line 86:
```suggestion
        constantEntitiesDrawer.onDraw(mapGraphics);
```

You can draw all constant entities, it is equivalent to this since zones no longer are inside there (and they never should've been)

src/main/java/com/github/manolo8/darkbot/gui/zones/safety/SafetiesDisplay.java
```diff
@@ -3,77 +3,93 @@
...
-    protected void drawMap(Graphics2D g2) {
-        g2.setColor(cs.TEXT_DARK);
-        g2.setFont(cs.FONTS.BIG);
-        drawString(g2, hero.map.name, mid, (height / 2) + 12, Align.MID);
+    protected void onPaint() {
+        zonesDrawer.drawZones(mapGraphics);
+
+        constantEntitiesDrawer.drawPortals(mapGraphics);
+        constantEntitiesDrawer.drawBattleStations(mapGraphics);
+        constantEntitiesDrawer.drawStations(mapGraphics);
```
Pablete1234, line 59:
```suggestion
        constantEntitiesDrawer.onDraw(mapGraphics);
```

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/drawables/InfosDrawer.java
```diff
@@ -0,0 +1,187 @@
...
+
+        String status = i18n.get((main.isRunning() ? "gui.map.running" : "gui.map.waiting"),
+                Time.toString(stats.getRunningTime().toMillis()));
+
+        mg.drawString(mg.getWidthMiddle(), mg.getHeight() / 2 + 35, status, MapGraphics.StringAlign.MID);
+
+        mg.setFont("small");
+        String info = i18n.get("gui.map.info", main.getVersion().toString(),
+                (main.isRunning() || !resetRefresh.getValue() ? Time.toString(System.currentTimeMillis() - main.lastRefresh) : "00"), Time.toString(refreshTime.getValue() * 60 * 1000L));
+
```
Pablete1234, line 84:
```suggestion
        String info = i18n.get("gui.map.info", main.getVersion().toString(),
                (main.isRunning() || !resetRefresh.getValue() ? Time.toString(System.currentTimeMillis() - main.lastRefresh) : "00"),
                Time.toString(refreshTime.getValue() * 60 * 1000L));
```

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -122,613 +50,173 @@ public MapDrawer(Main main) {
...
-    protected void drawCustomZonePath(Graphics2D g2, ZoneInfo zoneInfo) {
-        if (zoneInfo == null) return;
-        List<ZoneInfo.Zone> zones = zoneInfo.getSortedZones();
-        for (int i = 0; i < zones.size(); i++) {
-            Location loc1 = zones.get(i).innerPoint(0.5, 0.5, MapManager.internalWidth, MapManager.internalHeight);
-            Location loc2 = zones.get((i + 1) % zones.size()).innerPoint(0.5, 0.5, MapManager.internalWidth, MapManager.internalHeight);
-            drawLine(g2, loc1, loc2);
+        @Override
+        public int toScreenPointY(double gameY) {
+            return (int) Math.round(((gameY / mapBounds.getHeight()) * getHeight()));
```
Pablete1234, line 209:
```suggestion
            return (int) Math.round((gameY / mapBounds.getHeight()) * getHeight());
```

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -122,613 +50,173 @@ public MapDrawer(Main main) {
...
-    }
 
-    protected void drawString(Graphics2D g2, String str, int x, int y, Align align) {
-        if (str == null || str.isEmpty()) return;
-        if (align != Align.LEFT) {
-            int strWidth = g2.getFontMetrics().stringWidth(str);
-            x -= strWidth >> (align == Align.MID ? 1 : 0);
+        @Override
+        public double toGameLocationY(int screenY) {
+            return ((screenY / (double) getHeight()) * mapBounds.getHeight());
```
Pablete1234, line 219:
```suggestion
            return (screenY / (double) getHeight()) * mapBounds.getHeight();
```

src/main/java/com/github/manolo8/darkbot/gui/drawables/DevStuffDrawer.java
```diff
@@ -0,0 +1,72 @@
...
+import eu.darkbot.api.game.entities.Ship;
+import eu.darkbot.api.game.other.Lockable;
+import eu.darkbot.api.game.other.Point;
+import eu.darkbot.api.managers.ConfigAPI;
+import eu.darkbot.api.managers.EntitiesAPI;
+import eu.darkbot.api.managers.HeroAPI;
+
+import java.util.Collection;
+
+@Feature(name = "DevStuff Drawer", description = "Draws dev infos")
```
Pablete1234, line 20:
```suggestion
@Feature(name = "DevStuff Drawer", description = "Draws dev infos (eg: unknown entities, pathfinding points, and entity metadata)")
```

src/main/java/com/github/manolo8/darkbot/gui/drawables/DynamicEntitiesDrawer.java
```diff
@@ -0,0 +1,147 @@
...
+import eu.darkbot.api.game.other.Locatable;
+import eu.darkbot.api.game.other.Lockable;
+import eu.darkbot.api.game.other.Movable;
+import eu.darkbot.api.managers.ConfigAPI;
+import eu.darkbot.api.managers.EntitiesAPI;
+import eu.darkbot.api.managers.HeroAPI;
+
+import java.util.Collection;
+
+@Feature(name = "Dynamic Entities Drawer", description = "Draws dynamic entities")
```
Pablete1234, line 22:
```suggestion
@Feature(name = "Dynamic Entities Drawer", description = "Draws dynamic entities (eg: npcs, boxes, or players)")
```

src/main/java/com/github/manolo8/darkbot/gui/drawables/ConstantEntitiesDrawer.java
```diff
@@ -0,0 +1,71 @@
...
+import eu.darkbot.api.extensions.Drawable;
+import eu.darkbot.api.extensions.MapGraphics;
+import eu.darkbot.api.game.entities.BattleStation;
+import eu.darkbot.api.game.entities.Portal;
+import eu.darkbot.api.game.entities.Station;
+import eu.darkbot.api.managers.EntitiesAPI;
+
+import java.util.Collection;
+
+@Feature(name = "Constant Entities Drawer", description = "Draws not-movable entities (eg: portals, CBS, and stations)")
```
Pablete1234, line 13:
```suggestion
@Feature(name = "Constant Entities Drawer", description = "Draws entities that are constantly on the map (eg: portals, CBS, and stations)")
```
refering to *non-movable* may be misleading. eg: mines don't move, but are not drawn by this. I think the more relevant "distinction" is that the entities in here are always on the map, while others may come and go as you get closer (usually)

src/main/java/com/github/manolo8/darkbot/gui/drawables/InfosDrawer.java
```diff
@@ -0,0 +1,189 @@
...
+import eu.darkbot.api.managers.I18nAPI;
+import eu.darkbot.api.managers.PetAPI;
+import eu.darkbot.api.managers.StatsAPI;
+import eu.darkbot.api.utils.Inject;
+
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+import java.text.NumberFormat;
+
+@Feature(name = "Infos Drawer", description = "Draws infos about map, hero, health etc.")
```
Pablete1234, line 27:
```suggestion
@Feature(name = "Infos Drawer", description = "Draws info about bot state, current map, hero & target health etc.")
```

src/main/java/com/github/manolo8/darkbot/gui/drawables/StatsDrawer.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.ToIntFunction;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Feature(name = "Stats Drawer", description = "Draws statistics")
+public class StatsDrawer implements Drawable {
```
Pablete1234, line 30:
```suggestion
@Feature(name = "Overlay Drawer", description = "Draws overlays (eg: stats, boosters, or group info)")
public class OverlayDrawer implements Drawable {
```
I'd rename this class to OverlayDrawer, i think it better represents what this is

Pablete1234 (changes requested):

PR #140, opened 2022-03-11
Add manual captcha solve

fabio1999ita:


src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -30,8 +33,10 @@
 public class LoginUtils {
     private static final Pattern LOGIN_PATTERN = Pattern.compile("\"bgcdw_login_form\" action=\"(.*)\"");
     private static final Pattern DATA_PATTERN = Pattern.compile("\"src\": \"([^\"]*)\".*}, \\{(.*)}");
-
     private static final Map<String, String> FORCED_PARAMS = new HashMap<>();
+    static CookieManager cookieManager;
+    static boolean captcha;
```
Pablete1234, line 38: you need to scope your changes better, just having these 2 random variables here floating is going to be a future pain in maintainability, keep the cookie manager scoped to where it was

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -194,6 +208,8 @@ private static String replaceParameters(String params) {
     }
 
     private static String getLoginUrl(String in) {
+        System.out.println("captcha: " + in.contains("recaptcha_image"));
+        captcha = in.contains("recaptcha_image");
```
Pablete1234, line 212: What is this doing here? Why does a function to **get the login url from the string page** cause a side-effect in the global "captcha" variable?
Remove this from here, and perform this check where it belongs, not by abusing something else, you have the "frontPage" variable pretty clearly available

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -158,10 +166,16 @@ private static void usernameLogin(LoginData loginData, String domain) throws IOE
 
         CookieHandler.setDefault(null);
 
-        HttpCookie cookie = cookieManager.getCookieStore().getCookies().stream()
+        HttpCookie cookie;
+
+        if (captcha && domain.equals("lp")) {
+            Application.launch(LoginBrowser.class, loginData.getUsername(), loginData.getPassword());
+        }
```
Pablete1234, line 173: You can know there's a captcha before trying to log in, why do we try to log in (knowing it will fail) before doing this?

You have the CaptchAPI a bit higher up, which is the *perfect* place to just add this handling code

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -158,10 +166,16 @@ private static void usernameLogin(LoginData loginData, String domain) throws IOE
...
+
+        cookie = cookieManager.getCookieStore().getCookies().stream()
                 .filter(c -> c.getName().equalsIgnoreCase("dosid"))
                 .filter(c -> c.getDomain().matches(".*\\d+.*"))
-                .findFirst().orElseThrow(() -> new WrongCredentialsException("Wrong credentials or unsolved reCaptcha"));
+                .findFirst().orElse(null);
```
Pablete1234, line 178: Don't touch this stuff, leave it be as it was, this is for logging in, which you shouldn't need to mess with. You should only need to mess with the captcha, which is an "alternative" to logging in, you don't need to mess with this code to implement your thing

Pablete1234 (changes requested):
Overall this is pretty hacky, and requires bundling all of javafx with darkbot past java 11, which i'm not a fan of.

If you still want to proceed with this, i just don't get the need for the browser to perform the login. You can just load the mainpage, remove all the input form and leave just the captcha box for the user to solve. Once they solve the captcha return the g-reponse thing back and call to login with our own thing, but using the recaptcha response. Overall, the only thing you need to implement this as, is a `CaptchaAPI` which darkbot already provides an injection part on. Make a default implementation of CaptchAPI which works if no other 3rd party captcha solver is used, which uses the browser.

fabio1999ita:
video

Pablete1234:
Because this is getting implemented as a captcha solver which the bot already supports, no changes directly on darkbot are required

PR #143, opened 2022-03-19
Removed Update in Handle Response in auction and dispatch manager

anjeshshrestha:


Pablete1234 (changes requested):
What i'm seeing  here is you're simply wanting to avoid calling update after getting a reponse

This approach is kinda flawed, since we *do* want to update when we get a response.
The main issue trying to be solved is concurrent updates as you're iterating, to solve that, just act on that.

A way to solve such things is to make a copy and act upon that copy:

```java
    public List<String> collectAll() {
        List<InProgress> toCollect = data.getInProgress().values().stream()
                .filter(ip -> !ip.getCollectable().equals("0")).collect(Collectors.toList());
        return toCollect.map(this::collect).map(InProgress::getId).collect(Collectors.toList());
    }
```
It requires allocating an extra list, but it solves the root issue and doesn't create the side-effect of things not being updated after handling a response

anjeshshrestha:
> 

this should definitely work for the dispatch update issue. 

for the auction, we dont really need to update for that case because it doesnt impact the values we bid and slows it down loading the page over again and parsing it

Pablete1234 (commented):
Run the formatter, there's a bunch of missing spaces around the pr

Pablete1234 (approved):

PR #144, opened 2022-03-24
using auction type id

anjeshshrestha:
some cases where it wouldnt allow bidding because of wrong item id

removed enum type too, didnt seem to be much different from string for use case. maybe if day/week auction was also supported

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/backpage/auction/AuctionData.java
+++ b/src/main/java/com/github/manolo8/darkbot/backpage/auction/AuctionData.java
@@ -62,13 +62,11 @@ public class AuctionData {
             r.setId("item_hour_" + id);
             r.setAuctionType(AuctionItems.Type.HOUR);
         } else if (string.contains("day")) {
-            r = auctionItems.get("item_day_" + id);
-            if (r == null) auctionItems.put(id, r = new AuctionItems()); //need to clean up this somehow
+            r = auctionItems.computeIfAbsent("item_day_" + id, i -> new AuctionItems());
             r.setId("item_day_" + id);
             r.setAuctionType(AuctionItems.Type.DAY);
         } else if (string.contains("week")) {
-            r = auctionItems.get("item_week_" + id);
-            if (r == null) auctionItems.put(id, r = new AuctionItems()); //need to clean up this somehow
+            r = auctionItems.computeIfAbsent("item_week_" + id, i -> new AuctionItems());
             r.setId("item_week_" + id);
             r.setAuctionType(AuctionItems.Type.WEEK);
         } else {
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/manolo8/darkbot/backpage/auction/AuctionItems.java
```diff
@@ -92,10 +91,4 @@ public String toString() {
                 "instant=" + instantBuy +
                 "}";
     }
-
-    protected enum AuctionType {
-        HOUR,
-        DAY,
-        WEEK
-    }
```
Pablete1234, line 100: Don't get rid of AuctionType, keep it, and when sending the request just have a method that turns it into the string you want to send (just add a `String getId()` implemented as `name().toLowerCase(Locale.ROOT)`

Also, because you have this inside AuctionItems, you probably can just name it Type, so it ends up being AuctionItems.Type

src/main/java/com/github/manolo8/darkbot/backpage/auction/AuctionData.java
```diff
@@ -58,7 +58,19 @@ private boolean buildAuctionItems(String string, Pattern type) {
         AuctionItems r = auctionItems.get(id);
         if (r == null) auctionItems.put(id, r = new AuctionItems());
 
-        r.setId(id);
+        if (string.contains("hour")) {
+            r.setId("item_hour_" + id);
+            r.setAuctionType("hour");
```
Pablete1234, line 63:
```suggestion
            r.setAuctionType(AuctionItems.Type.HOUR);
```
etc

Pablete1234 (changes requested):

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/backpage/auction/AuctionData.java
```diff
@@ -5,32 +5,43 @@
...
+    public Map<String, AuctionItems> getAuctionHourItems() {
+        return auctionItems.entrySet().stream().filter(a -> a.getValue().type == AuctionItems.Type.HOUR).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    public Map<String, AuctionItems> getAuctionDayItems() {
+        return auctionItems.entrySet().stream().filter(a -> a.getValue().type == AuctionItems.Type.DAY).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    public Map<String, AuctionItems> getAuctionWeekItems() {
+        return auctionItems.entrySet().stream().filter(a -> a.getValue().type == AuctionItems.Type.WEEK).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
Pablete1234, line 37: split up these long lines into multiple shorter lines

Pablete1234 (approved):

PR #145, opened 2022-03-25
implement DataReaders

Punikekk:


src/main/java/com/github/manolo8/darkbot/core/api/util/DataReader.java
```diff
@@ -0,0 +1,49 @@
+package com.github.manolo8.darkbot.core.api.util;
+
+import java.nio.ByteBuffer;
+
+public interface DataReader extends AutoCloseable {
+
+    int MAX_CHUNK_SIZE = 2048;
+
+    ByteBuffer getByteBuffer();
```
Pablete1234, line 9: Is this really required? i think the concept of data reader doesn't need to forcibly be backed by a byte buffer

the AbstractDataReader can add it as an abstract method and it should all work the same

src/main/java/com/github/manolo8/darkbot/core/api/util/AbstractDataReader.java
```diff
@@ -0,0 +1,141 @@
...
+import com.github.manolo8.darkbot.core.api.GameAPI;
+import com.github.manolo8.darkbot.core.utils.ByteUtils;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public abstract class AbstractDataReader implements DataReader {
+
+    private final GameAPI.StringReader reader;
+
+    protected AtomicBoolean inUse = new AtomicBoolean();
```
Pablete1234, line 12:
```suggestion
    protected final AtomicBoolean inUse = new AtomicBoolean();
```

src/main/java/com/github/manolo8/darkbot/core/api/GameAPIImpl.java
```diff
@@ -51,6 +53,8 @@
 
     protected long lastFailedLogin;
 
+    protected AbstractDataReader[] dataReaders = new AbstractDataReader[10];
```
Pablete1234, line 56:
```suggestion
    protected DataReader[] dataReaders = new DataReader[10];
```
Can't we simply rely in *whatever* data reader this happens to be? that's the point of the interface, not relying on the impl

src/main/java/com/github/manolo8/darkbot/core/api/GameAPIImpl.java
```diff
@@ -279,6 +283,42 @@ public void readMemory(long address, byte[] buffer, int length) {
...
+
+            if (result == null) {
+                reader.reset(0);
+                return reader;
+            }
+
+            if (result) return reader;
+        }
+
+        throw new RuntimeException("Every reader is busy");
```
Pablete1234, line 305:
```suggestion
        throw new RuntimeException("All DataReaders are in use. Some code is calling readData and not closing the resource!");
```
Be a bit more specific here, it's always good to give a bit more info

src/main/java/com/github/manolo8/darkbot/core/api/GameAPIImpl.java
```diff
@@ -279,6 +283,42 @@ public void readMemory(long address, byte[] buffer, int length) {
...
+    public boolean readData(long address, int length, Consumer<DataReader> reader) {
+        try (DataReader r = readData(address, length)) {
+            if (r.getLimit() != length) return false;
+
+            reader.accept(r);
+        }
+        return true;
+    }
+
+    protected AbstractDataReader createReader(int idx) {
```
Pablete1234, line 318:
```suggestion
    protected DataReader createReader(int idx) {
```

src/main/java/com/github/manolo8/darkbot/core/objects/slotbars/Item.java
```diff
@@ -16,8 +17,7 @@
 import static com.github.manolo8.darkbot.Main.API;
 
 public class Item extends UpdatableAuto implements eu.darkbot.api.game.items.Item {
-    private static final int START = 36, END = 128 + 8;
-    private static final byte[] BUFFER = new byte[END - START];
+    private static final int START = 36, END = 136 - START;
```
Pablete1234, line 20:
```suggestion
    private static final int START = 36, SIZE = 136 - START;
```
These are not start & end, they're start & length or size or however you want to call it

src/main/java/com/github/manolo8/darkbot/core/objects/slotbars/Item.java
```diff
@@ -57,16 +57,22 @@ public void update() {
...
-        API.readMemory(address + START, BUFFER);
 
-        buyable = BUFFER[0] == 1;
-        activatable = BUFFER[4] == 1;
-        selected = BUFFER[8] == 1;
-        available = BUFFER[12] == 1;
-        visible = BUFFER[16] == 1;
-        quantity = ByteUtils.getDouble(BUFFER, 92);
+        long timerAdr;
+        try (DataReader reader = API.readData(address + START, END)){
```
Pablete1234, line 62:
```suggestion
        try (DataReader reader = API.readData(address + START, END)) {
```

src/main/java/com/github/manolo8/darkbot/core/objects/swf/FlashDictionary.java
```diff
@@ -0,0 +1,161 @@
...
+        return 0;
+    }
+
+    private static class Entry<K, V> {
+        private final Map<K, Lazy<V>> listeners;
+
+        private K key;
+        private V value;
+
+        public Entry(Map<K, Lazy<V>> listeners) {
```
Pablete1234, line 136: If you need to pass down the listeners to the entry, it may be simpler to just make this class non-static and use them from the outer class directly

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/api/GameAPIImpl.java
```diff
@@ -279,6 +283,42 @@ public void readMemory(long address, byte[] buffer, int length) {
...
     }
 
+    @Override
+    public DataReader readData(long address, int length) throws RuntimeException {
+        if (length <= 0 || length > DataReader.MAX_CHUNK_SIZE)
+            throw new ArrayIndexOutOfBoundsException("Length is <= 0 or exceeds max chunk size: " + DataReader.MAX_CHUNK_SIZE);
+
+        for (int i = 0; i < dataReaders.length; i++) {
+            AbstractDataReader reader = (AbstractDataReader) dataReaders[i];
+            if (reader == null) reader = (AbstractDataReader) (dataReaders[i] = createReader(i));
```
Pablete1234, line 293:
```suggestion
            DataReader reader = dataReaders[i];
            if (reader == null) reader = (dataReaders[i] = createReader(i));
```
Any reason this can't rely only on the interface, and not the implementation?

PR #146, opened 2022-03-29
Implemented default revive handler

Punikekk:


Pablete1234 (commented):
The current state of `RepairAPI` is probably not ready for this to be properly adopted

The current `getAvailableRepairOptions` is int-based while the new thing is enum based

Should probably include a `RepairAPI#isAvailable(RespawnOption option)` and `Collection<RespawnOption> RepairAPI#getAvailable()` for this to make sense

The code is ok tho

src/main/java/com/github/manolo8/darkbot/core/utils/ByteUtils.java
```diff
@@ -222,5 +226,42 @@ public void resetCache() {
...
+        @Override
+        public long searchClassClosure(Predicate<Long> pattern) {
+            long mainAddress = botInstaller.mainApplicationAddress.get();
+            if (mainAddress == 0) return 0;
+
+            long table = Main.API.readMemoryLong(mainAddress, 0x10, 0x10, 0x18, 0x10, 0x28);
+            int capacity = Main.API.readMemoryInt(table + 8) * 8;
+
+            if (tableData.length < capacity) {
+                tableData = new byte[capacity];
```
Pablete1234, line 244: We're allocating a table of arbitrary size, this is generally a pretty bad idea

The mem-read could return a random value that we cannot fullfil, causing an out of mem exception directly
Also, this table is not cleared after 750ms, it is only *replaced* after 750ms if anything happens to re-request it, but if nothing asks for this table again it will just stay there forever taking up however much memory it required

Pablete1234 (changes requested):

PR #147, opened 2022-03-29
Fixes some after jump desyncs

Punikekk:


src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -141,6 +141,13 @@ public void update(long address) {
         pet.update(API.readMemoryLong(address + 176));
         clickable.setRadius(0);
         id = API.readMemoryInt(address + 56);
+
+        if (locationInfo.isInitialized()) {
+            locationInfo.update();
+
+            drive.directStop();
+            drive.directStop();
+        }
```
Pablete1234, line 150: This is atm assuming that an update to the address, means a map jump, which isn't always the case

i believe this code should run a few MS after map jump, *if* the ship is not being moved (if a plugin/feature is already moving the ship then this desync isn't relevant)

So just consider saving a lastMapChange time, and if it's been 500ms since last map change, reset it to -1 and if the ship isn't moving call direct stop (to make the ship fly back to the current location client-side)

Pablete1234:
I believe this has became irrelevant with the changes in #157

PR #149, opened 2022-04-06
fix to open in current darkbot folder & auction regex match too long

anjeshshrestha:


src/main/java/com/github/manolo8/darkbot/backpage/AuctionManager.java
```diff
@@ -64,8 +67,12 @@ public boolean bidItem(AuctionItems auctionItem, double amount) {
...
 
     private boolean handleResponse(String type, String id, String response) {
-        boolean failed = response.contains("icon = 'error';");
-        System.out.println(type + " (" + id + ") " + (failed ? "failed" : "succeeded") + " : " + (failed ? response.substring(0, response.indexOf("function showHelp()")) : ""));
-        return !failed;
+        Matcher m = AUCTION_ERROR_PATTERN.matcher(response);
+        boolean valid = false;
+        if (m.find()) {
+            valid = !m.group(2).contains("error");
+            System.out.println(type + " (" + id + ") " + (valid ? "succeeded" : "failed") + " : " + m.group(1));
```
Pablete1234, line 74:
```suggestion
            System.out.println("AuctionManager: " + type + " (" + id + ") " + (valid ? "succeeded" : "failed") + " : " + m.group(1));
```
If we're leaving logs, better say where they come from

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #151, opened 2022-04-10
Added add support for new PetAPI 

m9w:
Implementing event added in MR of API: https://github.com/darkbot-reloaded/DarkBotAPI/pull/31
**This MR can be applied only when API was updated and was built new Release with updated dependency.**

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
@@ -22,7 +22,7 @@ import eu.darkbot.api.game.other.Health;
 import eu.darkbot.api.game.other.Locatable;
 import eu.darkbot.api.game.other.Location;
 import eu.darkbot.api.game.other.LocationInfo;
-import eu.darkbot.api.game.other.Lockable;
+import eu.darkbot.api.managers.EventBrokerAPI;
 import eu.darkbot.api.managers.PetAPI;
 import eu.darkbot.api.utils.Inject;
 import eu.darkbot.api.utils.ItemNotEquippedException;
@@ -53,6 +53,7 @@ public class PetManager extends Gui implements PetAPI {
     private final List<Ship> ships;
     private final Pet pet;
     private final PetGearSelectorHandler gearSelectorHandler;
+    private final EventBrokerAPI eventBroker;
 
     private long togglePetTime, selectModuleTime;
     private long activeUntil;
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
```diff
@@ -383,7 +385,10 @@ private void updateNpcLocatorList(long gearsSprite) {
         if (locatorNpcList.getSize() < oldSize && validUntil > System.currentTimeMillis()) return;
 
         validUntil = System.currentTimeMillis() + 100;
+        int oldState = locatorList.hashCode();
         locatorNpcList.sync(locatorList, Gear::new, null);
+        if (oldState != locatorList.hashCode())
```
Pablete1234, line 390: Technically speaking hashCode and equals work differently.
Even if hash code is not different, they could be different lists, because hash collisions exist.

Even if it's not nearly as efficient, i think here you should either:
A) create a copy of locator list before the clone, and compare with `equals` to the list after the sync
B) modify sync to return if anything has changed (this may be complicated as sync is already used in other situations where the result needs to be the returned list

src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
```diff
@@ -427,6 +432,11 @@ public void update() {
             this.fuzzyName = Strings.fuzzyMatcher(name);
             this.check = API.readMemoryLong(address, 208, 152, 0x10);
         }
+
+        @Override
+        public int hashCode() {
+            return name.hashCode();
```
Pablete1234, line 438: pretty sure that while you're doing this intentionally, it is not correct
The hashcode should be using all the fields, and probably the most important one, is id

I'm not totally sure, but what does ID represent for npcs? is it a random number or does the gear id match the npc id or something else?

Pablete1234 (changes requested):
Most importantly, look at the comments in the DarkBotAPI PR, as getting the API done properly is more important than the later implementation

src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
Pablete1234, line 390: I've implemented the `class Updatable.Reporting` which can be used in a new `syncAndReport` method, that returns true if anything has changed, and false otherwise.

If you rebase this PR you should be able to use that instead:

```suggestion
        if (locatorNpcList.syncAndReport(locatorList, Gear::new))
```
(You will need to update Gear to support the new interface, but it shouldn't be a big deal, see https://github.com/darkbot-reloaded/DarkBot/blob/master/src/main/java/com/github/manolo8/darkbot/core/entities/Zone.java#L41 for an example)

Pablete1234 (changes requested):

Pablete1234:
Implemented in 2248343

PR #152, opened 2022-04-13
Added custom background from the game

Punikekk:


src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -224,4 +266,72 @@ public double toGameLocationY(int screenY) {
             return (screenY / (double) getHeight()) * mapBounds.getHeight();
         }
     }
+
+    private static class BackgroundImgDownloaderTask extends SwingWorker<Image, Void> {
```
Pablete1234, line 270: This should probably be part of GameResourcesAPI (the same that handles translations, should have the methods to search for images)

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -224,4 +266,72 @@ public double toGameLocationY(int screenY) {
...
+
+        public BackgroundImgDownloaderTask(MapDrawer mapDrawer, int mapId) {
+            this.mapDrawer = mapDrawer;
+            this.mapId = mapId;
+        }
+
+        @Override
+        protected void done() {
+            try {
+                mapDrawer.backgroundImage = get();
```
Pablete1234, line 285: I think we should cache these in a local folder, create a "maps" or "map-images" folder inside the bot folder, and search for the image there first, and if not available then download it

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -124,6 +152,20 @@ private void drawActionButton() {
...
+            backgroundImage = null;
+            return;
+        }
+
+        if (main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.ENABLED
+                && main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.USE_GAME_BACKGROUND) {
+            new BackgroundImgDownloaderTask(this, e.getNext().getId())
+                    .execute();
+        }
+    }
```
Pablete1234, line 167: I think this should just be resolved on rendering time, otherwise it will only work after jumping map when you select a map.

You can just store a "int lastMapId", and update it when the image is changed, and avoid the Listener in general

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -79,10 +97,20 @@ public void setup(Main main) {
...
         //background image is drawn first
-        if (main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.ENABLED && main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.IMAGE.getImage() != null) {
-            mapGraphics.g2.setComposite(AlphaComposite.getInstance(AlphaComposite.SRC_OVER, (float) main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.OPACITY));
-            mapGraphics.g2.drawImage(main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.IMAGE.getImage(), 0, 0, mapGraphics.width, mapGraphics.height, this);
-            mapGraphics.g2.setComposite(AlphaComposite.getInstance(AlphaComposite.SRC_OVER, 1f));
+        if (main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.ENABLED) {
+            Image img = null;
+
+            if (main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.USE_GAME_BACKGROUND)
+                img = backgroundImage;
```
Pablete1234, line 104: in here you can just check for `if hero.map.id != lastMapId` and update the image (or start a worker that downloads it for the future)

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -79,10 +87,27 @@ public void setup(Main main) {
...
-            mapGraphics.g2.setComposite(AlphaComposite.getInstance(AlphaComposite.SRC_OVER, 1f));
+        if (main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.ENABLED) {
+            Image img = null;
+
+            if (main.config.BOT_SETTINGS.CUSTOM_BACKGROUND.USE_GAME_BACKGROUND) {
+                GameMap currentMap = main.hero.getMap();
+
+                if (currentMap != lastMap) {
+                    minimapFuture = currentMap.getId() > 0 ? flashResManager.getBackgroundImage(currentMap) : null;
+                    lastMap = currentMap;
```
Pablete1234, line 98:
```suggestion
                    if (minimapFuture != null) future.cancel();
                    this.minimapImg = null;
                    lastMap = currentMap;
                    minimapFuture = currentMap.getId() <= 0 ? null :
                            flashResManager.getBackgroundImage(currentMap)
                                    .thenAccept(result -> this.minimapImg = result)
                                    .whenComplete((result, t) -> this.minimapFuture = null);
```
Cancel any previous future, to avoid them potentially overriding the new image (although in practice due to the pipelined nature of the background thread (no concurrency) this would never occur.
Cleanup the current "minimapImg" (old)
Update last map
Resolve the new future, and when it's done, just set minimapImg to the result. Also when completed (either normally or exceptionally) cleanup the future so only the img itself stays.

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -79,10 +87,27 @@ public void setup(Main main) {
...
+                }
+
+                if (minimapFuture != null)
+                    img = minimapFuture.getNow(null);
```
Pablete1234, line 102: `img = minimapImg`

src/main/java/com/github/manolo8/darkbot/utils/FileUtils.java
```diff
@@ -27,6 +28,19 @@ public static void ensureDirectoryExists(Path path) {
...
+            Files.createDirectories(path);
+            return Files.exists(path);
+        } catch (IOException e) {
+            System.err.println("Failed to create: " + path);
+            e.printStackTrace();
+        }
+
+        return false;
+    }
+
```
Pablete1234, line 43:
```suggestion
```

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/backpage/FlashResManager.java
```diff
@@ -72,6 +91,16 @@ public void onTickTask() {
         }
     }
 
+    @Override
+    public void onBackgroundTick() {
```
Pablete1234, line 95: implementing onBackgroundTick makes onTickTask not run (since you're overriding that default impl. You will be breaking the above method wich loads translations

src/main/java/com/github/manolo8/darkbot/backpage/FlashResManager.java
```diff
@@ -72,6 +91,16 @@ public void onTickTask() {
...
+        parseMinimapIds(); // should be lazy-loaded?
```
Pablete1234, line 96:
```suggestion
        loadTranslations();
        loadMinimapIds();
```
Yeah we can make those run always, (they each check if they need to refresh or not)

src/main/java/com/github/manolo8/darkbot/backpage/FlashResManager.java
```diff
@@ -1,32 +1,51 @@
...
  * To be able to undo translations (convert a translated text into it's original translation id) in an efficient
  * manner we will substring all resources to the first N characters to be able to do an initial fast search.
  */
 @Feature(name = "Flash resource manager", description = "Holds the resources for the in-game language")
 public class FlashResManager implements Task, GameResourcesAPI {
-
+    private static final Path MINIMAPS_PATH = OSUtil.getDataPath("Images", "minimaps");
     private static final String URL = "https://darkorbit-22.bpsecure.com/spacemap/templates/{lang}/flashres.xml";
 
+    private static Map<Integer, String> mapBackgroundIds = new HashMap<>();
```
Pablete1234, line 45: move this down next to all translations, also, don't make it static

src/main/java/com/github/manolo8/darkbot/utils/OSUtil.java
```diff
@@ -0,0 +1,54 @@
...
+
+        private final String pattern;
+        private final Path botDataPath;
+
+        OS(@Language("RegExp") String pattern, Path appDataPath) {
+            this.pattern = pattern;
+            this.botDataPath = appDataPath.resolve("DarkBot");
+        }
+
+        private static OS findCurrentOS() {
```
Pablete1234, line 40:
```suggestion
        private static OS getCurrent() {
```
given that this is namespaced "OS" you'd have to call it with `OS.findCurrentOS` which sounds a bit redundant, so remove the 'OS'

Pablete1234 (changes requested):
mostly looking good, just a few small changes

PR #157, opened 2022-06-15
Added KekkaPlayer

Punikekk:


src/main/java/com/github/manolo8/darkbot/core/IDarkBotAPI.java
```diff
@@ -118,6 +119,20 @@ default void setVisible(boolean visible, boolean fullyHideEnabled) {
...
+    void setLocalProxy(int port);
+    void setPosition(int x, int y);
+    void setFlashOcxPath(String path);
+    void setClientSize(int width, int height);
+    void setMinClientSize(int width, int height);
+    void setTransparency(int transparency);
+    void setVolume(int volume); // 0 - 100
+
+    // LOW = 0, MEDIUM = 1, HIGH = 2, BEST = 3, AUTO_LOW = 4, AUTO_HIGH = 5
+    void setQuality(int quality);
```
Pablete1234, line 134: Should use an enum, then internally resolve as val.ordinal(), makes more sense from the outside

src/main/java/com/github/manolo8/darkbot/core/api/GameAPIImpl.java
```diff
@@ -77,14 +79,30 @@ public GameAPIImpl(StartupParams params,
...
             maxFps.addListener(fpsLimitListener = this::setMaxFps);
             setMaxFps(maxFps.getValue());
         } else {
             this.fpsLimitListener = null;
         }
+
+        if (hasCapability(GameAPI.Capability.PROXY) && (Boolean) config.getConfigValue("bot_settings.api_config.use_proxy")) {
+            new Thread(() -> {
+                int proxy = 7777;
+                while (proxy < 9999) {
```
Pablete1234, line 94: This is a 2k port range!! if you've exhausted a whole 2k port range there's some serious trouble elsewhere! You shouldn't be needing to test more than 100 ports.

src/main/java/com/github/manolo8/darkbot/core/api/GameAPIImpl.java
```diff
@@ -77,14 +79,30 @@ public GameAPIImpl(StartupParams params,
...
+                    try {
+                        new KekkaPlayerProxyServer(handler, proxy);
+                    } catch (BindException e) {
+                        e.printStackTrace();
```
Pablete1234, line 98: this will spam console/logs if running multiple bots and you have to skip a few ports, just add a single-line log `"Skipping port X for proxy: " + e.getMessage()` without the stacktrace

src/main/java/com/github/manolo8/darkbot/core/api/GameAPIImpl.java
```diff
@@ -77,14 +79,30 @@ public GameAPIImpl(StartupParams params,
...
-            ConfigAPI config = HeroManager.instance.main.configHandler;
             ConfigSetting<Integer> maxFps = config.requireConfig("bot_settings.api_config.max_fps");
             maxFps.addListener(fpsLimitListener = this::setMaxFps);
             setMaxFps(maxFps.getValue());
         } else {
             this.fpsLimitListener = null;
         }
+
+        if (hasCapability(GameAPI.Capability.PROXY) && (Boolean) config.getConfigValue("bot_settings.api_config.use_proxy")) {
+            new Thread(() -> {
```
Pablete1234, line 92: Personally not a fan of in-line threads, consider making a ProxyServerThread that extends thread, and sets a name (that also helps when looking at running threads on the debugger, as it has a name)

src/main/java/com/github/manolo8/darkbot/core/api/GameAPIImpl.java
```diff
@@ -77,14 +79,30 @@ public GameAPIImpl(StartupParams params,
...
+                int proxy = 7777;
```
Pablete1234, line 93:
```suggestion
                int port = 7777;
```
pretty sure port is a better var name here

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -142,7 +142,7 @@ public void update(long address) {
         super.update(address);
 
         pet.update(API.readMemoryLong(address + 176));
-        clickable.setRadius(0);
+        clickable.setRadius(-45);
```
Pablete1234, line 145: uh... what happened here??

src/main/java/eu/darkbot/utils/KekkaPlayerProxyServer.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.concurrent.Executors;
+
+public class KekkaPlayerProxyServer {
+
+    private static final ExecutorService THREAD_POOL = Executors.newFixedThreadPool(8);
+
+    public KekkaPlayerProxyServer(GameAPI.Handler handler, int port) throws IOException {
+        try (ServerSocket serverSocket = new ServerSocket(port)) {
+            int id = 0;
+            handler.setLocalProxy(port);
```
Pablete1234, line 18: if this is set in here, maybe after the try set it back to 0?

src/main/java/eu/darkbot/utils/KekkaPlayerProxyServer.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.*;
+import java.net.*;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+public class KekkaPlayerProxyServer {
+
+    private static final ExecutorService THREAD_POOL = Executors.newFixedThreadPool(8);
+
+    public KekkaPlayerProxyServer(GameAPI.Handler handler, int port) throws IOException {
```
Pablete1234, line 15: What is really the point of an object in this case? the object is stateless and has no methods except the constructor!!

Consider simply making this a static method instead:
`KekkaPlayerProxyServer.runProxy(handler, port)`

src/main/resources/lang/strings_en.properties
```diff
@@ -228,6 +228,7 @@ config.bot_settings.api_config.force_game_language.desc=Set the in-game language
 config.bot_settings.api_config.enforce_hw_accel=Enforce hardware acceleration
 config.bot_settings.api_config.enforce_hw_accel.desc=This solves bugs in the dark orbit client on dark boat, but can have a negative impact in performance.
 config.bot_settings.api_config.use_3d=Use 3D in-game
+config.bot_settings.api_config.use_proxy=Use proxy(necessary on win7)
```
Pablete1234, line 231:
```suggestion
config.bot_settings.api_config.use_proxy=Use proxy (must on win7)
```

src/main/java/com/github/manolo8/darkbot/gui/components/ApiSettingsPanel.java
```diff
@@ -20,6 +23,49 @@ public ApiSettingsPanel(Config.BotSettings.APIConfig display, JComponent parent)
...
+            JButton b = new JButton("Clear Cache");
+            b.addActionListener(l -> Main.API.clearCache());
+            add(b, "span 3, grow");
+        }
+
+        if (Main.API.hasCapability(GameAPI.Capability.HANDLER_CLEAR_RAM)) {
+            JButton b = new JButton("Clear Ram");
+            b.addActionListener(l -> Main.API.emptyWorkingSet());
+            add(b, "span 3, grow");
+        }
```
Pablete1234, line 46: These probably belong in the diagnostics popup menu instead of the api visbility one, but i don't really mind it much

Pablete1234 (changes requested):
Overall pretty good, just some minor changes

Husstlee (changes requested):
.

Punikekk:
> .

?

dm94 (changes requested):
The previous option could also be used, I know it is more problematic but safer.

if (trySetShowing(value)) {
            if (value) {
                clickMinimized();
            } else {
                if (changeCloseClick || !clickMinimized())
                    click(5, 5);

                changeCloseClick = !changeCloseClick;
            }
            toggleVisibility();
            return false;
        }

src/main/java/com/github/manolo8/darkbot/gui/drawables/ConstantEntitiesDrawer.java
```diff
@@ -42,14 +43,21 @@ public void drawPortals(MapGraphics mg) {
...
+                mg.drawRectCentered(bs, 3, false);
+            } else {
                 mg.drawOvalCentered(bs, 11, 9, true);
 
-            else mg.drawRectCentered(bs, 3, false);
+
+//                if (bs.hasEffect(EffectManager.Effect.DEFLECTOR_SHIELD.getId())) {
+//                    mg.setColor("meteroid");
+//                    mg.drawOvalCentered(bs, 30, 25, false);
+//                }
```
Pablete1234, line 59: ?

src/main/java/com/github/manolo8/darkbot/modules/utils/AttackAPIImpl.java
```diff
@@ -1,100 +1,158 @@
...
+    }
 
-        @Override
-        public SelectableItem.Laser get() {
-            return useRsb ? SelectableItem.Laser.RSB_75
-                    : useSab ? SelectableItem.Laser.SAB_50
-                    : SelectableItem.Laser.LCB_10;
-            //todo change with config fields
-        }
+    // TODO: 11.07.2022 use offensive abilities on its own (venom/cyborg or diminisher)
```
Pablete1234, line 91: So is this a lost feature atm?

src/main/java/com/github/manolo8/darkbot/modules/utils/LegacyFlashPatcher.java
```diff
@@ -83,7 +83,8 @@ private static class FlashInstaller implements Fixer {
                 REG_SVR = Paths.get(System.getenv("WINDIR"), "SysWOW64", "regsvr32");
 
         public boolean needsFix() {
-            return LibSetup.downloadLib("DarkFlash.ocx", FLASH_OCX);
+            String ocxName = "DarkFlash" + (OSUtil.isWindows7OrLess() ? "-W7" : "") + ".ocx";
+            return LibSetup.downloadLib(ocxName, FLASH_OCX);
```
Pablete1234, line 87: this "-W7" logic exists in 2 places, here and in LibUtils#getFlashOcxPath, i think this should reuse the other

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -96,7 +96,12 @@ public PrintStreamWithDate(String logfile) throws FileNotFoundException, Unsuppo
 
         @Override
         public void println(String string) {
-            super.println("[" + LocalDateTime.now().format(LOG_DATE) + "] " + string);
+            StackTraceElement[] stack = Thread.currentThread().getStackTrace();
+            string = "[" + LocalDateTime.now().format(LOG_DATE)
+                    + (stack.length > 2 ? " | " + stack[2] : "")
```
Pablete1234, line 101: I think using the whole stacktrace is too long, most logs end up with things like:
`com.github.manolo8.darkbot.Bot.main(Bot.java:51)`
I think we should shorten the package if it's from the bot (ie: `com.github.manolo8.darkbot` -> `db` or `eu.darkbot.api` ->`dbapi`)

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/modules/utils/AttackAPIImpl.java
Punikekk: yes, but it is kinda replaced by features in plugins

src/main/java/com/github/manolo8/darkbot/modules/utils/LegacyFlashPatcher.java
Punikekk: actually this shouldn't be touched

src/main/java/com/github/manolo8/darkbot/core/IDarkBotAPI.java
```diff
@@ -130,6 +135,28 @@ default void setVisible(boolean visible, boolean fullyHideEnabled) {
     void refine(long refineUtilAddress, OreAPI.Ore ore, int amount);
 
     long callMethod(int index, long... arguments);
+    boolean callMethodChecked(boolean checkName, String signature, int index, long... arguments);
+    boolean callMethodAsync(int index, long... arguments);
+    boolean useItem(Item item);
+    boolean isUseItemSupported();
```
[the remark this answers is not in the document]
Punikekk: this method checks also if is able to use item currently

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -160,13 +164,18 @@ public void jumpPortal(com.github.manolo8.darkbot.core.entities.Portal portal) {
...
         if (!portal.isValid()) return;
         if (System.currentTimeMillis() - portalTime < 500) return; // Minimum delay
         if ((System.currentTimeMillis() - portalTime > 20000 || isNotJumping(portal)) &&
                 (portal.isSelectable() || portals.stream().noneMatch(p -> p != portal && p.isSelectable()))) {
-            API.keyboardClick(keybinds.getCharCode(JUMP_GATE));
+
+            if (clickPortal) portal.trySelect(false);
+            else keybinds.pressKeybind(JUMP_GATE);
+
+            clickPortal = !clickPortal;
```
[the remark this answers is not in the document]
Pablete1234: 1) What stuck at portal issue?
2) did it fix it?
3) shouldn't it just use the jump keybind, and only if it fails, then in a subsequent attempt then try to use click alternatively?

src/main/java/com/github/manolo8/darkbot/core/IDarkBotAPI.java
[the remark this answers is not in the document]
Pablete1234: Nothing prevents capabilities changing dynamically in hasCapability, darkboat with darkhook would return true/false to direct movement etc depending on if you enabled it or not in the settings. Essentially override the hasCapability method in kekka adapter, and if it's asking for direct item support do extra logic, otherwise, default to super.hasCapability

src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
```diff
@@ -73,8 +73,8 @@ public class GuiManager implements Manager, GameScreenAPI {
     private enum LoadStatus {
         WAITING(gm -> gm.main.hero.address != 0 && !gm.connecting.isVisible()),
         AFTER_LOGIN(gm -> {
-            API.directKeyClick(gm.main.config.LOOT.AMMO_KEY);
-            API.directKeyClick(gm.main.config.LOOT.AMMO_KEY);
+            API.rawKeyboardClick(gm.main.config.LOOT.AMMO_KEY, false);
+            API.rawKeyboardClick(gm.main.config.LOOT.AMMO_KEY, false);
```
Pablete1234, line 77: use `API.keyboardClick(Character, boolean)` (it doesn't exist, you'll have to add it, but `raw` keyboard clicks should be an internal thing, not needed by outside callers)

Pablete1234 (approved):

PR #158, opened 2022-06-15
parse dispatch collected info

anjeshshrestha:


src/main/java/com/github/manolo8/darkbot/backpage/DispatchManager.java
```diff
@@ -42,11 +52,19 @@ public boolean update(int expiryTime) {
         return false;
     }
 
+    public HashMap<String, Integer> getCollected(){
```
Pablete1234, line 55:
```suggestion
    public Map<String, Integer> getCollected() {
```

src/main/java/com/github/manolo8/darkbot/backpage/DispatchManager.java
```diff
@@ -42,11 +52,19 @@ public boolean update(int expiryTime) {
...
+        return collected;
+    }
+
+    public HashMap<String, Integer> getLastCollected(){
```
Pablete1234, line 59:
```suggestion
    public Map<String, Integer> getLastCollected() {
```

src/main/java/com/github/manolo8/darkbot/backpage/DispatchManager.java
```diff
@@ -18,10 +22,16 @@ public class DispatchManager {
     private final Main main;
     private final DispatchData data;
     private long lastDispatcherUpdate;
+    private final HashMap<String, Integer> collected;
+    private final HashMap<String, Integer> lastCollected;
```
Pablete1234, line 26:
```suggestion
    private final Map<String, Integer> collected;
    private final Map<String, Integer> lastCollected;
```

src/main/java/com/github/manolo8/darkbot/backpage/DispatchManager.java
```diff
@@ -105,6 +123,18 @@ public boolean collect(InProgress progress) {
 
     public boolean handleResponse(String type, String id, String response) {
         boolean failed = response.contains("\"result\":\"ERROR\"");
+        if(!failed){
```
Pablete1234, line 126:
```suggestion
        if (!failed) {
```

src/main/java/com/github/manolo8/darkbot/backpage/DispatchManager.java
```diff
@@ -105,6 +123,18 @@ public boolean collect(InProgress progress) {
...
+            this.lastCollected.clear();
+            JsonObject jsonObj = g.fromJson (response, JsonObject.class); //Converts the json string to JsonElement without POJO
+            for(JsonElement item : jsonObj.get("rewardsLog").getAsJsonArray()){
```
Pablete1234, line 129:
```suggestion
            for (JsonElement item : jsonObj.get("rewardsLog").getAsJsonArray()) {
```

Pablete1234 (changes requested):
Mostly minor formatting issues

Also keep in mind that you should use the general interface `Map` over specific implementations (eg: `HashMap`) in the code in general. Assign them to instances of `new HashMap` but don't restrict them being changed in the future if needed

PR #161, opened 2022-07-01
force only one instance for same .jar (need java 9+)

fabio1999ita:


dm94:
I understand the problems of running multiple instances of the same executable but I know a lot of people who do it this way these days.
Is this change necessary?

fabio1999ita:
> I understand the problems of running multiple instances of the same executable but I know a lot of people who do it this way these days. Is this change necessary?

running from same jar make plugin unusable (when refresh all plugin or update it) and increase risk of crash, so...

dm94:
> running from same jar make plugin unusable (when refresh all plugin or update it) and increase risk of crash (it can delete also useless support), so...

the truth is that it would reduce the problems generated by not knowing how to use the bot.

Pablete1234:
The main issue with this one is it takes away power from users, simply show a pop-up on startup that warns about this instead of straight up doing nothing. Doing nothing will not get you less support issues, it will get you the same people who would try to run it twice, instead of complaining that it broke for running twice, complaining that the bot doesn't work anymore.

Also i believe checking the log file is NOT the way to prevent multiple instances running, and that another program (that isn't darkbot) locking that file will make darkbot not load, which is absurd.

fabio1999ita:
> the truth is that it would reduce the problems generated by not knowing how to use the bot.

really a lot of program can run multiple instance, darkbot should not

> The main issue with this one is it takes away power from users, simply show a pop-up on startup that warns about this instead of straight up doing nothing. Doing nothing will not get you less support issues, it will get you the same people who would try to run it twice, instead of complaining that it broke for running twice, complaining that the bot doesn't work anymore.
> 
> Also i believe checking the log file is NOT the way to prevent multiple instances running, and that another program (that isn't darkbot) locking that file will make darkbot not load, which is absurd.

you can't make 1000 popup of warning, if you not like read log find another way

Pablete1234:
> you can't make 1000 popup of warning

Yes you can either make a warning or not prevent it from happening, but doing nothing when starting up is going to be more confusing than a pop up. It's not good behavior.

> if you not like read log find another way
A common solution that other software integrates is to create a "pid" file on startup, that saves the process id. When another instance of the program loads, it will check that pid file, and search in the OS if there's a process with that id running, if there is, then you show the warning that there's already another instance running. Also, add a shutdown hook on the bot that deletes the file.

fabio1999ita:
> Also, add a shutdown hook on the bot that deletes the file.

this should on startup, so when you launch bot it check for all file with pid name and delete the useless, because if bot crash he not delete pid file

Pablete1234:
If you delete the file on startup you never know if another is running. The point is that you take the process id stored in the file and check if such process is running. If the previous one crashed, you won't find a process by that id, and so, you know it's not 2 instances of the bot running.

fabio1999ita:
> If you delete the file on startup you never know if another is running.

not true, first you need to read file and check if pid exists, if not delete that file, all this on startup

Pablete1234:
If it does not exist, you simply overwrite it with the new PID, there's no point in deleting it

fabio1999ita:
image

TeYroXOfficial:
Would be nice if you could add possibility to translate this popup via strings for each language.

fabio1999ita:
> Would be nice if you could add possibility to translate this popup via strings for each language.

the problem is that has 3 line of text, so need 3 line of traduction or some strange system to go next line

Pablete1234:
several string on the bot are multiline either using \n or <html>line1<br>line2</html>, search for them in strings file and see

Regardless, while that's a nice to have, for this popup i wouln't be that concerned with localization, i'm more concerned with the use of weird libs, ,runtime, and windows-specific api to search process. Consider just making this check java9>, and use java9's process handle api,which is MUCH cleaner

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +38,43 @@ public static void main(String[] args) throws IOException {
...
+
+        Path filePath = Paths.get("curr.pid");
+        long currentPid = ProcessHandle.current().pid();
+
+        if (Files.exists(filePath)) {
+            List<String> fileContent = new ArrayList<>(Files.readAllLines(filePath));
+
+            if (fileContent.isEmpty()) {
+                Files.writeString(filePath, "\n" + currentPid, StandardOpenOption.APPEND);
+            } else {
```
Pablete1234, line 57:
```suggestion
            if (!fileContent.isEmpty()) {
```
Just go straight to checking

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +38,43 @@ public static void main(String[] args) throws IOException {
...
+                for (int i = 0; i < fileContent.size(); i++) {
```
Pablete1234, line 58: I would ignore anything that isn't the first line, we do not have a need to store a list, at all

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +38,43 @@ public static void main(String[] args) throws IOException {
...
+                    if (ProcessHandle.of(Long.parseLong(fileContent.get(i))).isPresent()) {
```
Pablete1234, line 59: If the file contains anything that isn't a number, this will throw an exception when parsing. If something like that occurs you should be ignoring it not throwing an exception

Additionally,  you should be checking if the process handle info matches that of the same process (in other words, it is also a darkbot instance running, not just any other random program). You can check this like:
ProcessHandle other = ProcessHandle.of(...);
if (other.isPresent() && other.info().command()...)

You can check that the command is the same (eg: java.exe), and the args match (eg: -jar darkbot.jar)

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +38,43 @@ public static void main(String[] args) throws IOException {
...
+                        Files.writeString(filePath, "\n" + currentPid, StandardOpenOption.APPEND);
```
Pablete1234, line 60: does this mean we only write *if* there's a collision?
or otherwise, have a completely separate if statement to handle the same thing (writing to file) ?

Just do this after the whole checking section, and only if the user decides to proceed (if they shutdown instead, no need). Also don't append, simply overwrite if it exists already, Create new if missing. i believe Files.writeString already has an option for it. It's simpler and we don't need to keep track of all processes that are created as long as we've warned them at least once.

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +38,43 @@ public static void main(String[] args) throws IOException {
...
+                        Popups.showMessageSync("Multiple bot from same folder", new JOptionPane(
+                                "You're currently running multiple bot from same folder" + "\n" +
+                                "This may can cause crash or unexpected problems.\n" +
+                                "Please create new folder for another instance of bot.",
+                                JOptionPane.WARNING_MESSAGE, JOptionPane.DEFAULT_OPTION));
```
Pablete1234, line 65: An option to either cancel start, or "continue anyways" should be here

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,86 @@ public static void main(String[] args) throws IOException {
...
+                        }
+                    } catch (java.lang.NumberFormatException e) {
+                        e.printStackTrace();
+                        Files.writeString(filePath, currentPid + " " + currentStartTime, StandardOpenOption.TRUNCATE_EXISTING);
+                    }
+                }
+            }
+        } else {
+            Files.writeString(filePath, currentPid + " " + currentStartTime, StandardOpenOption.CREATE);
+        }
```
Pablete1234, line 121: I would again insist into not having this write string 3 times, instead having it as a final step outside all the ifs

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,86 @@ public static void main(String[] args) throws IOException {
...
+                                            "This can cause crash, unexpected problems, and some broken functionality.\n" +
+                                            "Please create a separate folder to run multiple instances.",
+                                            JOptionPane.WARNING_MESSAGE)
+                                    .options(new Object[]{proceed, cancel})
+                                    .initialValue(cancel)
+                                    .showOptionSync();
+
+                            if (result.get() == 1) {
+                                System.out.println("Reject multiple bot warning, closing bot");
+                                System.exit(100);
```
Pablete1234, line 106:
```suggestion
                                System.exit(0);
```
This is closing the program in a normal manner, no need for an exceptional exit

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +38,43 @@ public static void main(String[] args) throws IOException {
...
         SwingUtilities.invokeLater(() -> new Main(params));
     }
 
+    private static void checkUniqueInstance(StartupParams params) throws IOException {
+        if (params.has(StartupParams.LaunchArg.NO_WARN)) return;
+
+        Path filePath = Paths.get("curr.pid");
+        long currentPid = ProcessHandle.current().pid();
+
+        if (Files.exists(filePath)) {
```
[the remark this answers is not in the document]
Pablete1234: i'd like to re-iterate on this one, it would simplify the code

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,86 @@ public static void main(String[] args) throws IOException {
...
+        Path filePath = Paths.get("curr.pid");
+        long currentPid = ProcessHandle.current().pid();
+        ProcessHandle processHandle = ProcessHandle.current();
+        ProcessHandle.Info processInfo = processHandle.info();
+        long currentStartTime = processInfo.startInstant().map(Instant::toEpochMilli).orElse(0L);
+
+        if (Files.exists(filePath)) {
+            List<String> fileContent = Files.readAllLines(filePath, StandardCharsets.UTF_8);
+
+            if (!fileContent.isEmpty()) {
```
Pablete1234, line 61: there is no need for this check, it will simply perform a loop with 0 iterations without it, which reduces identation as a whole

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,86 @@ public static void main(String[] args) throws IOException {
...
+                        }
+
+                        ProcessHandle externalProcessHandle = ProcessHandle.of(filePid).orElse(null);
+                        long externalStartTime = 0;
+                        if (externalProcessHandle != null) {
+                            ProcessHandle.Info ExternalProcessInfo = externalProcessHandle.info();
+                            externalStartTime = ExternalProcessInfo.startInstant().map(Instant::toEpochMilli).orElse(0L);
+                        }
+
+                        if (ProcessHandle.of(filePid).isPresent() && externalStartTime == fileStartTime) {
```
Pablete1234, line 79:
```suggestion
                        if (externalProcessHandle != null && externalStartTime == fileStartTime) {
```

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,86 @@ public static void main(String[] args) throws IOException {
...
+                            Files.writeString(filePath, "\n" + currentPid + " " + currentStartTime, StandardOpenOption.APPEND);
```
Pablete1234, line 80: just leave this one for the end, if you close the panel there's no need for the pid to be written at all

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,87 @@ public static void main(String[] args) throws IOException {
...
+
+        Path filePath = Paths.get("curr.pid");
+        long currentPid = ProcessHandle.current().pid();
+        ProcessHandle processHandle = ProcessHandle.current();
+        ProcessHandle.Info processInfo = processHandle.info();
+        long currentStartTime = processInfo.startInstant().map(Instant::toEpochMilli).orElse(0L);
+
+        List<String> fileContent;
+        if (Files.exists(filePath)) {
+            fileContent = Files.readAllLines(filePath, StandardCharsets.UTF_8);
```
Pablete1234, line 60:
```suggestion
        List<String> fileContent = Files.exists(filePath) ? Files.readAllLines(filePath, StandardCharsets.UTF_8) : Collections.emptyList();
```
You can remove the inlining in the whole block

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,87 @@ public static void main(String[] args) throws IOException {
...
+            int count = 0;
```
Pablete1234, line 61: this is not needed

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,87 @@ public static void main(String[] args) throws IOException {
...
+                            System.out.println("Reject multiple bot warning, closing bot");
+                            System.exit(0);
+                        } else {
+                            Files.writeString(filePath, "\n" + currentPid + " " + currentStartTime, StandardOpenOption.APPEND);
+                        }
+                        break;
+                    } else {
+                        if (count == fileContent.size() - 1)
+                            Files.writeString(filePath, currentPid + " " + currentStartTime, StandardOpenOption.TRUNCATE_EXISTING);
+                    }
```
Pablete1234, line 112:
```suggestion
                    }
```

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,87 @@ public static void main(String[] args) throws IOException {
...
+                } catch (java.lang.NumberFormatException e) {
+                    e.printStackTrace();
+                    Files.writeString(filePath, currentPid + " " + currentStartTime, StandardOpenOption.TRUNCATE_EXISTING);
+                }
+                count++;
```
Pablete1234, line 117:
```suggestion
                } catch (java.lang.NumberFormatException e) {
                    e.printStackTrace();
                }
```

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,87 @@ public static void main(String[] args) throws IOException {
...
+            }
+
+        } else {
+            Files.writeString(filePath, currentPid + " " + currentStartTime, StandardOpenOption.CREATE);
+        }
```
Pablete1234, line 122:
```suggestion
        Files.writeString(filePath, currentPid + " " + currentStartTime, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING);
```
You do not need the if-s at all, you can simply write at the very end, and always write if it happens to reach here, if you decide to close it will exit earlier

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,10 +41,87 @@ public static void main(String[] args) throws IOException {
...
+                                .options(new Object[]{proceed, cancel})
+                                .initialValue(cancel)
+                                .showOptionSync();
+
+                        if (result.get() == 1) {
+                            System.out.println("Reject multiple bot warning, closing bot");
+                            System.exit(0);
+                        } else {
+                            Files.writeString(filePath, "\n" + currentPid + " " + currentStartTime, StandardOpenOption.APPEND);
+                        }
```
Pablete1234, line 107:
```suggestion
                        }
```
Just let the opposite case continue, it will write the pid at the end.

Pablete1234 (changes requested):

fabio1999ita:
if everything is fine i will add the translations

there is another small problem where if you press the X to close the warning popup the program continues while it should close

fabio1999ita:
is not possible use `I18n.get` in main class? because i get:

```
Exception in thread "main" java.lang.NullPointerException
	at com.github.manolo8.darkbot.utils.I18n.get(I18n.java:56)
	at com.github.manolo8.darkbot.Bot.checkUniqueInstance(Bot.java:80)
	at com.github.manolo8.darkbot.Bot.main(Bot.java:47)
```

fabio1999ita:
need a `warning.png` icon

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,21 +43,83 @@ public static void main(String[] args) throws IOException {
...
+
+        Path filePath = Paths.get("curr.pid");
+        long currentPid = ProcessHandle.current().pid();
+        ProcessHandle processHandle = ProcessHandle.current();
+        ProcessHandle.Info processInfo = processHandle.info();
+        long currentStartTime = processInfo.startInstant().map(Instant::toEpochMilli).orElse(0L);
+
+        List<String> fileContent;
+        try {
+            fileContent = Files.exists(filePath) ? Files.readAllLines(filePath, StandardCharsets.UTF_8) : Collections.emptyList();
```
Pablete1234, line 60:
```suggestion
        try {
                    List<String> fileContent = Files.exists(filePath) ? 
                            Files.readAllLines(filePath, StandardCharsets.UTF_8) : Collections.emptyList();
```
doesn't seem like fileContent is used anywhere outside, so can likely just be put in the inner scope

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,21 +43,83 @@ public static void main(String[] args) throws IOException {
...
+
+                    ProcessHandle externalProcessHandle = ProcessHandle.of(filePid).orElse(null);
+                    long externalStartTime = 0;
+                    if (externalProcessHandle != null) {
+                        ProcessHandle.Info ExternalProcessInfo = externalProcessHandle.info();
+                        externalStartTime = ExternalProcessInfo.startInstant().map(Instant::toEpochMilli).orElse(0L);
+                    }
+
+                    if (externalProcessHandle != null && externalStartTime == fileStartTime) {
+                        JButton proceed = new JButton(I18n.get("start.same_folder_warn.button.proceed"), UIUtils.getIcon("warning")); //TODO add warning icon
```
Pablete1234, line 79:
```suggestion
                        JButton proceed = new JButton(I18n.get("start.same_folder_warn.button.proceed"), UIUtils.getIcon("warning"));
```

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -32,21 +43,83 @@ public static void main(String[] args) throws IOException {
         LibSetup.setupLibraries();
         StartupParams params = new StartupParams(args);
 
-        checkJavaVersion(params);
         SwingUtilities.invokeLater(() -> new Main(params));
     }
 
-    private static void checkJavaVersion(StartupParams params) {
+    static void checkUniqueInstance(StartupParams params) {
```
Pablete1234, line 49: Keep order consistent, checkJavaVersion was here first, is called first, and should be called first, so keep that as the first of the 2 methods in h ere

Pablete1234 (approved):
Will need to wait until we drop java8 support, but otherwise it's ready

PR #164, opened 2022-08-09
Added logging function and UDP log request

gabrielrossier:
### What was added ?

- Logging class that allow you to centralize logging functions
- UDP server that will send every log on port 7504

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 This version of darkbot has a built-in udp client that will send log on port 8015
 This also support multi-instance (It will add Process Id to make difference between instance)
 
-How to run:
+## How to run:
 - Clone the repository
 - Get the latest full-release on darkbot's discord https://discord.gg/uXHnZJ9
 - Unzip the release in a known folder outside of the project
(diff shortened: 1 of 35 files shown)


Pablete1234:
Hey, i'm not sure that i understand the need to have all logging calls replaced. You can overwrite what system out/err print to, which is why, we usually redirect it to a file. You can simply, on that level, implement support for managing the log in other ways, and then implement sending it via udp as part of a plugin since i don't believe that one is required behavior on the base bot for most people

gabrielrossier:
The problem I see here is that, plugins are not loaded until login success. I would like to be able to monitor logs what happen before login.

> Hey, i'm not sure that i understand the need to have all logging calls replaced. You can overwrite what system out/err print to, which is why, we usually redirect it to a file. You can simply, on that level, implement support for managing the log in other ways, and then implement sending it via udp as part of a plugin since i don't believe that one is required behavior on the base bot for most people

gabrielrossier:
I'm currently working on cleaner version

PR #165, opened 2022-08-10
Option to enable UDP logging in addition of file logging

gabrielrossier:
### UDP logging
Hello this is the cleaner version of my previous pull request that I've close.

The goal of this is to be able to read logs from another process using UDP on port 7504. By not making this as a plugins allow you to monitor your bot even before login to check if the bot has successfully started or if an error occured.

### What's new ? 
Possibility to enable UDP logging by setting "true" in the logConfig.conf file (Automatically created if missing with default configuration to false) in addition of file logging.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,8 +89,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.1</version>
                 <configuration>
-                    <source>8</source>
-                    <target>8</target>
+                    <source>9</source>
+                    <target>9</target>
                     <encoding>UTF-8</encoding>
                 </configuration>
             </plugin>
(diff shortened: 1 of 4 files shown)


Pablete1234:
Hey, i think what you should be looking into is maintaining stdout even if we're writing to a file, instead of sending the data via sockets.

A good place to start is how `FileDescriptor.out` stays there even if you overwrite System.setOut, (by default, System.out outputs to FileDescriptor.out, this logic can be seen in System.class's initialization)

I think the best solution to incorporate what you want here, is simply for the override-log behavior of the bot (where it logs to a file) makes it so system.out writes to both a file, and to a stream that points to FileDescriptor.out

This is roughly what java does when initializing system.out:
```
FileOutputStream fdOut = new FileOutputStream(FileDescriptor.out);
PrintStream stdout = new PrintStream(new BufferedOutputStream(fdOut, 128), true, "UTF-8"));

System.setOut(stout);
```
You can do similar, and call this stOut from our LogUtils ontop of calling to store on file

Pablete1234:
Closing due to lack of response, i feel this functionality should be part of reading std out instead of ports

PR #167, opened 2022-08-16
Add support for File class as config type

m9w:
Feature that replace JFileOpener
Based on ImagePicker.class


src/main/java/com/github/manolo8/darkbot/gui/tree/EditorProvider.java
```diff
@@ -4,21 +4,7 @@
...
-import com.github.manolo8.darkbot.gui.tree.editors.ImagePicker;
-import com.github.manolo8.darkbot.gui.tree.editors.MultiDropdownEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.NumberEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.PercentEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.PlayerTagEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.RangeEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.ShipModeEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.StringEditor;
-import com.github.manolo8.darkbot.gui.tree.editors.TableEditor;
+import com.github.manolo8.darkbot.gui.tree.editors.*;
```
Pablete1234, line 7: Disable wildcard imports in your IDE, keep them separated, it is generally bad to use wildcard imports in java because they can cause code to break for none of your fault (a class being added elsewhere, that now is imported and matches-up with a pre-existing name)

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/FileEditor.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.function.Consumer;
+
+public class FileEditor extends JButton implements OptionEditor<File> {
+    private JFileChooser fc;
+    private File file;
+    private final Consumer<JTree> stopEditing = JTree::stopEditing;
+
+    public FileEditor() {
+        setFocusable(false);
+        addActionListener(e -> stopEditing.accept((JTree) SwingUtilities.getAncestorOfClass(JTree.class, FileEditor.this)));
```
Pablete1234, line 21: These 2 are very questionable choices
I believe the button needs to remain focusable, at the very least for keyboard support, but also because otherwise i'm not sure what will even be focused by the tree node when editing
The second, assumes a tree is available, which isn't necessarily always available, also, when the button is pressed the pop-up to pick a file should be opened, not edition stopped, so it just feels pretty counter intuitive as to what's really going on here

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/FileEditor.java
```diff
@@ -0,0 +1,83 @@
...
+                fc = new JFileChooser(path) {
+                    @Override
+                    protected JDialog createDialog(Component parent) throws HeadlessException {
+                        JDialog dialog = super.createDialog(parent);
+                        dialog.setAlwaysOnTop(true);
+                        return dialog;
+                    }
+                };
+            }
+            if (fc.showOpenDialog(null) == JFileChooser.APPROVE_OPTION) setEditing(fc.getSelectedFile());
```
Pablete1234, line 34: since this is a straight copy from the other editor, it feels like it could probably be extracted to a separate class reused by both, maybe an `OnTopFileChooser` ?

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/FileEditor.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public void setText(String text) {
+        int sepIdx;
+        if (text != null && text.length() > 30 && (sepIdx = text.indexOf(File.separator, text.length() - 30)) != -1) {
+            super.setText(".." + text.substring(sepIdx));
+        } else {
+            super.setText(text);
+        }
+
+        setPreferredSize(new Dimension(getFontMetrics(getFont()).stringWidth(getText()) + 32, 0));
```
Pablete1234, line 48: This is again a straight copy, potentially make an util, like `Strings.shortFileName(String)`?

would make this get quite reduced to:
`super.setText(Strings.shortFileName(text));`

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/FileEditor.java
```diff
@@ -0,0 +1,83 @@
+package com.github.manolo8.darkbot.gui.tree.editors;
+
+import com.github.manolo8.darkbot.gui.AdvancedConfig;
+import com.google.gson.*;
```
Pablete1234, line 4: Same for this wildcard

Pablete1234 (changes requested):
Looking generally good, just a few comments

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/FileEditor.java
m9w: You right, it not work with keyboard(
I added this trick because new select files window triggered of releasing focus from button so as result not displayed new file name however saves to tree. It not reproduce if button setFocusable(false). 
This trick make possible to show new file name. I can't found another way to fix it. I will try to do it tomorrow.

Pablete1234: How does this work on the image picker? does it work there?
in theory when saving, either the tree editor should still be triggered, in which case, the name should just change by  your call to "setEditing" (which calls setText), if it's no longer focused, it's weird that it would even properly work to save the setting at all, as you need to reply to getEditorValue with the new value for it to be updated

m9w:
How I can see in repo exist a lot of wildcard for java.awt.* javax.swing.* com.google.gson.* so for this classes I leave it as is.

Pablete1234:
For specifically awt and swing, idea has a separate limit in its config, but the gson one probably slipped in and shouldn't be there

You can (and should) change this setting in idea under `Settings > Editor > Code Style > Java > imports > class count to use import with * = 999, Names count to use static import with * = 999`. Swing and awt are already added as exceptions by default:

image

src/main/java/com/github/manolo8/darkbot/gui/utils/window/FileChooserUtil.java
```diff
@@ -0,0 +1,28 @@
...
+import javax.swing.filechooser.FileNameExtensionFilter;
+import java.io.File;
+
+public class FileChooserUtil {
+    public static JFileChooser getChooser(String path) {
+        return getChooser(path, null, false);
+    }
+
+    public static JFileChooser getChooser(String path, FileNameExtensionFilter filter, boolean acceptAllFileFilter) {
+        if(path == null) path = new File(System.getProperty("user.dir")).getAbsolutePath();
```
Pablete1234, line 14:
```suggestion
        if (path == null) path = new File(System.getProperty("user.dir")).getAbsolutePath();
```

src/main/java/com/github/manolo8/darkbot/gui/utils/window/FileChooserUtil.java
```diff
@@ -0,0 +1,28 @@
...
+        JFileChooser fc = new JFileChooser(path) {
+            @Override
+            protected JDialog createDialog(Component parent) throws HeadlessException {
+                JDialog dialog = super.createDialog(parent);
+                dialog.setAlwaysOnTop(true);
+                return dialog;
+            }
+        };
+
+        if(filter != null) fc.setFileFilter(filter);
```
Pablete1234, line 24:
```suggestion
        if (filter != null) fc.setFileFilter(filter);
```

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ImagePicker.java
```diff
@@ -2,40 +2,26 @@
...
-                };
-                FileNameExtensionFilter filter = new FileNameExtensionFilter(
-                        "Images Files", "png", "jpeg", "jpg", "gif");
-                fc.setFileFilter(filter);
-                fc.setAcceptAllFileFilterUsed(false);
-            }
-
+            if (fc == null) fc = FileChooserUtil.getChooser(image.getPath(),
+                    new FileNameExtensionFilter("Images Files", "png", "jpeg", "jpg", "gif"),
+                    false);
```
Pablete1234, line 24:
```suggestion
            if (fc == null)
                fc = FileChooserUtil.getChooser(image.getPath(),
                        new FileNameExtensionFilter("Images Files", "png", "jpeg", "jpg", "gif"),
                        false);
```
unless it fits in a single line, i think it's best to always put the continuation of the if in a new line for readability

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/FileEditor.java
```diff
@@ -0,0 +1,70 @@
...
+    private File file;
+
+    public FileEditor() {
+        addActionListener(e -> {
+            if (fc == null) fc = FileChooserUtil.getChooser(file.getPath());
+            JTree tree = (JTree) SwingUtilities.getAncestorOfClass(JTree.class, FileEditor.this);
+            TreePath path = tree.getEditingPath();
+            if (fc.showOpenDialog(null) == JFileChooser.APPROVE_OPTION) {
+                setEditing(fc.getSelectedFile());
+                tree.getModel().valueForPathChanged(path, getEditorValue());
```
Pablete1234, line 27: When doing this, you're avoiding the user having the option to cancel the edition (eg: pressing ESC)
This works because the editor themselves don't force a new value, they simply provide the new value in `getEditorValue`, and when the user confirms edition (by either losing focus, hitting enter, or clicking the green tick) then stopEdition is called, which gets current editor's `getEditorValue` and saves it as `valueForPathChanged`, if instead the user hits ESC or the red X, editionCancelled is called instead which does not update the value.

If the value here is being lost and not updating, the issue is that focus is being lost at some point due to the opening of the file chooser? How does this work with the image chooser?

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/FileEditor.java
m9w: When opens new window then focus from component will be lost. It's no problem when we change the field of object (it's case of image chooser) but when we create new object that will be stored in properties we required do similar trick.

Please note that the cancellation is not work for image chooser because we patched original object that storing in config. Losing focus make not possible for this, I not found way to get back focus on button after closing window of selecting file.

Pablete1234: From what you're saying it seems like image chooser may be having faulty behavior then, it however, may be a pain to fix

I'll try to look into that, but otherwise for the time being i'm fine with this hard-coded value-change, *however* do leave a FIXME comment explaining the situation, along the lines of:
`// FIXME: opening the file picker makes the tree lose focus, so cell can't save, meaning we have to force it.`

That will hopefully make it clear this is not what should be done usually, it's just a bad workaround because of how the tree works

Pablete1234 (approved):

PR #168, opened 2022-08-16
Add autologin by SID

m9w:
Was customized logic of autologin.
If `sid` and `server` is present in prop file than for first attempt of login will used this credentials.
If previous step is unsuccessful and exist `login/password` for default authorization than this credential will used for attempts.
If default authorization was successful and `server` present in prop file than new SID will saved to prop file.

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
+++ b/src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
@@ -1,5 +1,6 @@
 package com.github.manolo8.darkbot.utils;
 
+import com.github.manolo8.darkbot.gui.utils.Strings;
 import eu.darkbot.api.API;
 import eu.darkbot.util.function.ThrowingFunction;
 
@@ -89,7 +90,7 @@ public class StartupParams implements API.Singleton {
         }
     }
 
-    public AutoLoginProps getAutologinInstance(){
+    public AutoLoginProps getAutoLoginProps(){
         return (AutoLoginProps) startupParams.getOrDefault(LaunchArg.LOGIN, null);
     }
 
@@ -125,34 +126,57 @@ public class StartupParams implements API.Singleton {
                 .collect(Collectors.joining(","));
     }
 
-    public static class AutoLoginProps {
+    public static class AutoLoginProps extends Properties {
         private final String path;
-        private final Properties props;
 
         private AutoLoginProps(String path) throws IOException {
             this.path = path;
-            props = new Properties();
             try (InputStreamReader reader = new InputStreamReader(new FileInputStream(path), StandardCharsets.UTF_8)) {
-                props.load(reader);
+                load(reader);
             }
             System.out.println("Loaded startup properties file");
         }
-        public String getAutoLoginValue(PropertyKey key) {
-            return props.getProperty(key.toString());
+
+        public String getUsername() {
+            return getProperty(PropertyKey.USERNAME.toString());
         }
 
-        public char[] getAutoLoginMasterPassword() {
-            String masterPassword = getAutoLoginValue(PropertyKey.MASTER_PASSWORD);
+        public String getPassword() {
+            return getProperty(PropertyKey.PASSWORD.toString());
+        }
+
+        public char[] getMasterPassword() {
+            String masterPassword = getProperty(PropertyKey.MASTER_PASSWORD.toString());
             return masterPassword == null ? null : masterPassword.toCharArray();
         }
 
-        public void setAutoLoginProp(PropertyKey key, String val) {
-            props.setProperty(key.toString(), val);
+        public String getServer() {
+            return getProperty(PropertyKey.SERVER.toString());
+        }
+
+        public String getSID() {
+            return getProperty(PropertyKey.SID.toString());
+        }
+
+        public boolean isAllowStoreSID() {
+            return Boolean.parseBoolean(getProperty(PropertyKey.ALLOW_STORE_SID.toString()));
+        }
+
+        public boolean isPossibleSIDLogin() {
+            return !Strings.isEmpty(getSID()) && !Strings.isEmpty(getServer());
+        }
+
+        public void setServer(String server) {
+            setProperty(PropertyKey.SERVER.toString(), server);
+        }
+
+        public void setSID(String sid) {
+            setProperty(PropertyKey.SID.toString(), sid);
         }
 
         public void updateLoginFile() {
             try (OutputStreamWriter writer = new OutputStreamWriter(new FileOutputStream(path), StandardCharsets.UTF_8)) {
-                props.store(writer, null);
+                store(writer, null);
             } catch (IOException e) {
                 e.printStackTrace();
             }
--- a/src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
+++ b/src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
@@ -63,7 +63,7 @@ public class StartupParams implements API.Singleton {
     }
 
     public enum PropertyKey {
-        USERNAME, PASSWORD, MASTER_PASSWORD, SERVER, SID;
+        USERNAME, PASSWORD, MASTER_PASSWORD, SERVER, SID, ALLOW_STORE_SID;
         @Override
         public String toString() {
             return this.name().toLowerCase(Locale.ROOT);
--- a/src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
+++ b/src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
@@ -126,57 +126,67 @@ public class StartupParams implements API.Singleton {
                 .collect(Collectors.joining(","));
     }
 
-    public static class AutoLoginProps extends Properties {
+    public static class AutoLoginProps {
+        private final Properties prop;
         private final String path;
 
         private AutoLoginProps(String path) throws IOException {
+            this.prop = new Properties();
             this.path = path;
             try (InputStreamReader reader = new InputStreamReader(new FileInputStream(path), StandardCharsets.UTF_8)) {
-                load(reader);
+                prop.load(reader);
             }
             System.out.println("Loaded startup properties file");
         }
 
+        private String getProperty(PropertyKey key) {
+            return prop.getProperty(key.toString());
+        }
+
+        private void setProperty(PropertyKey key, String val) {
+            prop.setProperty(key.toString(), val);
+        }
+
         public String getUsername() {
-            return getProperty(PropertyKey.USERNAME.toString());
+            return getProperty(PropertyKey.USERNAME);
         }
 
         public String getPassword() {
-            return getProperty(PropertyKey.PASSWORD.toString());
+            return getProperty(PropertyKey.PASSWORD);
         }
 
         public char[] getMasterPassword() {
-            String masterPassword = getProperty(PropertyKey.MASTER_PASSWORD.toString());
+            String masterPassword = getProperty(PropertyKey.MASTER_PASSWORD);
             return masterPassword == null ? null : masterPassword.toCharArray();
         }
 
         public String getServer() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -81,21 +84,32 @@ public static LoginData performAutoLogin(StartupParams params) {
         loginData.setCredentials(username, password);
```
Pablete1234, line 85:
```suggestion
        if (isSidLogin) loginData.setSid(sid, server + ".darkorbit.com");
```
i'd just set it from the start, since this cannot produce an exception

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -81,21 +84,32 @@ public static LoginData performAutoLogin(StartupParams params) {
...
 
         try {
-            System.out.println("Auto logging in (1/2)");
-            usernameLogin(loginData);
+            if(isSidAutoLogin) loginData.setSid(sid, server + ".darkorbit.com");
+            else {
+                System.out.println("Auto logging in (1/2)");
+                usernameLogin(loginData);
+            }
```
Pablete1234, line 91:
```suggestion
            if (!isSidLogin) {
                System.out.println("Auto logging in (1/2)");
                usernameLogin(loginData);
            }
```

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -81,21 +84,32 @@ public static LoginData performAutoLogin(StartupParams params) {
...
             System.out.println("Loading spacemap (2/2)");
             findPreloader(loginData);
         } catch (IOException e) {
             System.err.println("IOException trying to perform auto login, servers may be down");
             e.printStackTrace();
         } catch (WrongCredentialsException e) {
+            if(isSidAutoLogin) {
```
Pablete1234, line 98:
```suggestion
            if (isSidAutoLogin) {
```
Small formatting change

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -81,21 +84,32 @@ public static LoginData performAutoLogin(StartupParams params) {
...
+                System.err.println("Wrong SID");
```
Pablete1234, line 99:
```suggestion
                System.err.println("Expired SID in login properties file, attempting re-connect with user & pass");
```
descriptive errors will help users when they find themselves fighting this, which will eventually occur

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -81,21 +84,32 @@ public static LoginData performAutoLogin(StartupParams params) {
...
+            }
             System.err.println("Wrong credentials, check your username and password");
         }
 
         if (loginData.getPreloaderUrl() == null || loginData.getParams() == null) {
             System.err.println("Could not find preloader url or parameters, exiting bot.");
             System.exit(-1);
         }
+
+        if(!Strings.isEmpty(server) && !isSidAutoLogin) params.updateSidAndServer(loginData.getSid(), loginData.getUrl().split("\\.")[0]);
```
Pablete1234, line 111: I think this should be a prop in its own, like `UPDATE_SID=true` in the file, you're assuming here that just because we have a server we want to print the sid which seems a bit odd
```suggestion
        if (!isSidLogin && Boolean.parseBoolean(params.getAutoLoginValue(StartupParams.PropertyKey.UPDATE_SID)))
            params.updateSidAndServer(loginData.getSid(), loginData.getUrl().split("\\.")[0]);
```
additionally, you could probably do this right after calling `usernameLogin` as if that did not throw an exception, you there know *for sure* the sid will be fresh

Pablete1234 (changes requested):
Just a few comments to improve the feature, but looking pretty good! keep up the good work

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
m9w: I think flag is not necessary because if user want use only sid than properties should not contain login/password.
I not see scenario when user want use sid and login and not want store actual sid iin case present is wrong

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
m9w: On this line we can not grantee that attempt of re-login will happens, for example login/password is absent

Pablete1234: That's true, we cannot guarantee it'll happen
however, we just saying we're re-attempting, the next error you'll get after this, will be the one that is like "For auto login you need either user & pass, or sid & server", so you will be able to see what happened

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
Pablete1234: You may not have SID atm, but setup user & pass with the intention that they will save, and get re-used in future uses

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -135,4 +124,38 @@ public String toString() {
                 .map(e -> e.getKey().toString() + "= " + e.getValue().toString())
                 .collect(Collectors.joining(","));
     }
+
+    public static class AutoLoginProps {
```
Pablete1234, line 128: I'd keep this an internal class, not public and then keep the getAutoLoginValue, and other methods accessible directly via StartupParams
```suggestion
    private static class AutoLoginProps {
```

but again, this is mostly personal preference

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
```diff
@@ -84,24 +89,8 @@ public StartupParams(String[] args) throws IOException {
...
-
-    public String getAutoLoginValue(PropertyKey key) {
-        Properties p = (Properties) startupParams.get(LaunchArg.LOGIN);
-        return p.getProperty(key.name().toLowerCase(Locale.ROOT));
-    }
-
-    public char[] getAutoLoginMasterPassword() {
-        String masterPassword = getAutoLoginValue(PropertyKey.MASTER_PASSWORD);
-        return masterPassword == null ? null : masterPassword.toCharArray();
+    public AutoLoginProps getAutologinInstance(){
```
Pablete1234, line 92:
```suggestion
    public AutoLoginProps getAutoLogin() {
```
formatting

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -61,41 +62,57 @@ public static LoginData performUserLogin(StartupParams params) {
...
+            System.err.println("Credentials file requires username & either a password or a master password, or/and server & sid");
             System.exit(-1);
         }
 
         LoginData loginData = new LoginData();
         loginData.setCredentials(username, password);
 
+        System.out.println("Attempt of auto logging using " + (isSidLogin?"server/SID":"login/password") + " in (1/2)");
+        if (isSidLogin) loginData.setSid(sid, server + ".darkorbit.com");
+        else usernameLogin(loginData);
```
Pablete1234, line 88:
```suggestion
```
this needs to occur inside the try

the setSid is what can be here without an issue because it makes no calls that could lead to an exception, the usernameLogin **can** lead to exceptions and should be inside the try

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -61,41 +62,57 @@ public static LoginData performUserLogin(StartupParams params) {
...
+
         try {
-            System.out.println("Auto logging in (1/2)");
-            usernameLogin(loginData);
```
Pablete1234, line 85: `ìf (!isSidLogin) usernameLogin(loginData);`

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -61,41 +62,57 @@ public static LoginData performUserLogin(StartupParams params) {
...
-            System.err.println("Credentials file requires username and either a password or a master password");
+        if (!isSidLogin && (username == null || Strings.isEmpty(password))) {
+            System.err.println("Credentials file requires username & either a password or a master password, or/and server & sid");
             System.exit(-1);
         }
 
         LoginData loginData = new LoginData();
         loginData.setCredentials(username, password);
 
+        System.out.println("Attempt of auto logging using " + (isSidLogin?"server/SID":"login/password") + " in (1/2)");
```
Pablete1234, line 86:
```suggestion
        System.out.println("Auto logging in using " + (isSidLogin ? "server & SID" : "user & password") + " (1/2)");
```

Pablete1234 (changes requested):
I insist on the need for a separate flag about updating user/sid, but otherise looks good :+1:

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
m9w: Now role of this flag keep the `server` field.
BUT
I think may join `server` and `sid` to single line like prop file will contain `sid=eu1:yyyySiDIIIIiiiiiiiiiiiiiiiiiiiiiiiiiiii`, and existing of this field will be flag for store new sid. I mean if user not configure sid in props than props file never update.
Because now it look like that autologin by user/password will switch to server that declared in `server` field.

If sid wrong and we gets new than why we should keep expired sid in props.

m9w:
Want to receive confirmation regarding https://github.com/darkbot-reloaded/DarkBot/pull/168#discussion_r948893663

src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
Pablete1234: I think joining server & sid on the same prop is a bad idea, generally you want data to be atomic
it also makes it more intuitive, server and sid are basically never combined, and shouldn't be here either

how would it be if we had `user=username:pass`, it'd look pretty weird (ignoring that : could be part of user or pass so it'd be unreliable, of course)

I think they should be kept separate

Pablete1234:
I answered that, but i'll re-iterate in general what i believe the behavior that makes the most sense is:

- If you include both SID and SERVER, try to login with SID
  - If login is ok, proceed
  - If login fails, erase the SID from the file (optional, i don't think it's a *must* but it makes sense), fallback to user & pass login
- Attempt login with user & pass
  - If login is ok, and you have a prop to save sid set, update server & sid
  - If login fails, try again later?

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
m9w: image

src/main/java/com/github/manolo8/darkbot/utils/StartupParams.java
m9w: Method with same name already exist
```
public boolean getAutoLogin() {
    return has(LaunchArg.LOGIN);
}
```

Pablete1234: I was mostly concerned about capitalization and the space, you could name it `getAutoLoginProps()` too

m9w:
As for me access via getter look better than using enum constants.

Pablete1234 (changes requested):
Just a couple last formatting changes

Pablete1234 (approved):

PR #169, opened 2022-08-17
Allow auto feature class search in plugins

m9w:
Related to https://github.com/darkbot-reloaded/DarkBotAPI/pull/39

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureRegistry.java
+++ b/src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureRegistry.java
@@ -7,28 +7,20 @@ import com.github.manolo8.darkbot.extensions.plugins.Plugin;
 import com.github.manolo8.darkbot.extensions.plugins.PluginHandler;
 import com.github.manolo8.darkbot.extensions.plugins.PluginListener;
 import com.github.manolo8.darkbot.utils.I18n;
+import eu.darkbot.api.config.ConfigSetting;
 import eu.darkbot.api.extensions.FeatureInfo;
 import eu.darkbot.api.extensions.PluginInfo;
-import eu.darkbot.api.extensions.RegisterFeature;
 import eu.darkbot.api.managers.ExtensionsAPI;
 import eu.darkbot.api.utils.Inject;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.io.IOException;
-import java.net.URL;
-import java.net.URLClassLoader;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Enumeration;
 import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.jar.JarEntry;
-import java.util.jar.JarFile;
 import java.util.stream.Stream;
 
 public class FeatureRegistry implements PluginListener, ExtensionsAPI {
@@ -73,30 +65,10 @@ public class FeatureRegistry implements PluginListener, ExtensionsAPI {
     @Override
     public void afterLoad() {
         pluginHandler.LOADED_PLUGINS.forEach(pl ->
-                getFeaturesList(pl)
+                Arrays.stream(pl.getDefinition().features)
                         .forEach(feature -> registerPluginFeature(pl, feature)));
         pluginHandler.LOADED_PLUGINS.sort((p1, p2) -> Boolean.compare(isDisabled(p1), isDisabled(p2)));
         registryHandler.update();
-    }
-
-    private List<String> getFeaturesList(Plugin plugin) {
-        if (plugin.getDefinition().features.length != 0) return Arrays.asList(plugin.getDefinition().features);
-        URLClassLoader loader = new URLClassLoader(new URL[]{plugin.getJar()});
-        List<String> features = new LinkedList<>();
-        try (JarFile jarFile = new JarFile(plugin.getFile())) {
-            Enumeration<JarEntry> e = jarFile.entries();
-            while (e.hasMoreElements()) {
-                JarEntry jarEntry = e.nextElement();
-                if (jarEntry.getName().endsWith(".class")) {
-                    String className = jarEntry.getName().replace("/", ".").replace(".class", "");
-                    if(loader.loadClass(className).isAnnotationPresent(RegisterFeature.class))
-                        features.add(className);
-                }
-            }
-        } catch (IOException | ClassNotFoundException e) {
-            e.printStackTrace();
-        }
-        return features;
     }
 
     private boolean isDisabled(Plugin plugin) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureRegistry.java
```diff
@@ -65,12 +73,32 @@ public void beforeLoad() {
...
-                Arrays.stream(pl.getDefinition().features)
+                getFeaturesList(pl)
                         .forEach(feature -> registerPluginFeature(pl, feature)));
         pluginHandler.LOADED_PLUGINS.sort((p1, p2) -> Boolean.compare(isDisabled(p1), isDisabled(p2)));
         registryHandler.update();
     }
 
+    private List<String> getFeaturesList(Plugin plugin) {
+        if (plugin.getDefinition().features.length != 0) return Arrays.asList(plugin.getDefinition().features);
+        URLClassLoader loader = new URLClassLoader(new URL[]{plugin.getJar()});
```
Pablete1234, line 84: Plugins are already in a loader, why are we loading them a second time?

src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureRegistry.java
```diff
@@ -65,12 +73,32 @@ public void beforeLoad() {
...
+        List<String> features = new LinkedList<>();
+        try (JarFile jarFile = new JarFile(plugin.getFile())) {
+            Enumeration<JarEntry> e = jarFile.entries();
+            while (e.hasMoreElements()) {
+                JarEntry jarEntry = e.nextElement();
+                if (jarEntry.getName().endsWith(".class")) {
+                    String className = jarEntry.getName().replace("/", ".").replace(".class", "");
+                    if(loader.loadClass(className).isAnnotationPresent(RegisterFeature.class))
```
Pablete1234, line 92:
```suggestion
                    if (loader.loadClass(className).isAnnotationPresent(RegisterFeature.class))
```
small formatting change

src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureRegistry.java
```diff
@@ -65,12 +73,32 @@ public void beforeLoad() {
...
+                getFeaturesList(pl)
                         .forEach(feature -> registerPluginFeature(pl, feature)));
         pluginHandler.LOADED_PLUGINS.sort((p1, p2) -> Boolean.compare(isDisabled(p1), isDisabled(p2)));
         registryHandler.update();
     }
 
+    private List<String> getFeaturesList(Plugin plugin) {
+        if (plugin.getDefinition().features.length != 0) return Arrays.asList(plugin.getDefinition().features);
+        URLClassLoader loader = new URLClassLoader(new URL[]{plugin.getJar()});
+        List<String> features = new LinkedList<>();
```
Pablete1234, line 85:
```suggestion
        List<String> features = new ArrayList<>();
```
just use an array list, it's not like a linked list provides any benefit here

src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureRegistry.java
```diff
@@ -65,12 +73,32 @@ public void beforeLoad() {
     @Override
     public void afterLoad() {
         pluginHandler.LOADED_PLUGINS.forEach(pl ->
-                Arrays.stream(pl.getDefinition().features)
+                getFeaturesList(pl)
```
Pablete1234, line 76: I'd make this an either-or, so you either use `pl.getDefinition().features`, or, if it's null (the author doesn't define them in the list) then you search for the classes annotated with `@Feature` instead
That way the author may define them in either the "old way" (for backwards compat reasons), or simply clear the list, and then we'll search for all features ourselves

Pablete1234 (changes requested):
I already commented in the API pr, i like the idea but i wouldn't do it on all plugins from the get-go, i'd do it only if they do not define a features list, and i wouldn't use a new annotation i'd just reuse the current one.

Additionally, i'd suggest we make use of `basePackage` which exists on `plugin.json`, and only scan for classes within that. Otherwise uber-jars containing lots of classes from dependencies will take forever scanning lots of classes that are irrelevant to us.

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginHandler.java
```diff
@@ -260,5 +266,22 @@ private void testSignature(Plugin plugin, JarFile jar) throws IOException {
         }
     }
 
+    private String[] fetchFeatureList(Plugin plugin, JarFile jar) {
+        URLClassLoader loader = new URLClassLoader(new URL[]{ plugin.getJar() });
+        List<String> features = new ArrayList<>();
+        Enumeration<JarEntry> entries = jar.entries();
+        while (entries.hasMoreElements())
```
Pablete1234, line 273:
```suggestion
        while (entries.hasMoreElements()) {
```
This is a code block, put some braces around it for readability

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginHandler.java
```diff
@@ -260,5 +266,22 @@ private void testSignature(Plugin plugin, JarFile jar) throws IOException {
...
+            try {
+                JarEntry jarEntry = entries.nextElement();
+                if (jarEntry.getName().endsWith(".class")) {
+                    String className = jarEntry.getName().replace("/", ".").replace(".class", "");
+                    if(loader.loadClass(className).isAnnotationPresent(Feature.class))
```
Pablete1234, line 278:
```suggestion
                    if (loader.loadClass(className).isAnnotationPresent(Feature.class))
```
Formatting

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginHandler.java
```diff
@@ -260,5 +266,22 @@ private void testSignature(Plugin plugin, JarFile jar) throws IOException {
...
+                        features.add(className);
+                }
```
Pablete1234, line 280: This is essentially going to load **every single class**, in the jar, regardless of if it's needed or not, just to search for features, we have a basePackage in the plugin.json (the plugin definition) and we should use it. Search/load only classes which are inside that base package, avoid iterating or loading all dependencies a plugin includes, they're not gonna be features

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginHandler.java
```diff
@@ -187,9 +193,19 @@ private void loadPlugins(File[] pluginFiles, List<Plugin> previousPlugins) {
...
+        pluginEntryNames.entrySet().stream()
+                .filter(plugin -> LOADED_PLUGINS.contains(plugin.getKey()))
+                .forEach(plugin ->
+                        plugin.getKey().getDefinition().features = plugin.getValue().stream()
+                        .filter(file -> file.endsWith(".class"))
+                        .map(class_ -> class_.replace("/", ".").replace(".class", ""))
+                        .filter(class_ -> plugin.getKey().getBasePackage() == null || class_.startsWith(plugin.getKey().getBasePackage()))
+                        .filter(this::isFeature)
+                        .toArray(String[]::new)
+                );
```
Pablete1234, line 205:
```suggestion
        pluginClasses.forEach((plDef, cls) ->
                        plDef.features = cls.stream().filter(this::isFeature).toArray(String[]::new));
```
This should be significantly easier with the other changes

Pablete1234 (changes requested):
Looking much better, just a few overall cleanup changes and it's good to go :+1:

Pablete1234 (approved):

PR #170, opened 2022-08-17
Add support multiline status

m9w:


src/main/java/com/github/manolo8/darkbot/gui/drawables/InfosDrawer.java
```diff
@@ -92,8 +92,9 @@ private void drawInfos(MapGraphics mg) {
             String s = (main.isRunning() && main.repairManager.isDestroyed())
                     ? main.repairManager.getStatus()
                     : (main.tickingModule ? main.getModule().getStatus() : main.getModule().getStoppedStatus());
-
-            mg.drawString(5, 26, s, MapGraphics.StringAlign.LEFT);
+            int i = 12;
+            if(s != null) for (String line : s.split("\n"))
+                mg.drawString(5, i+=14, line, MapGraphics.StringAlign.LEFT);
```
Pablete1234, line 97:
```suggestion
            if (s != null) {
                int i = 12;
                for (String line : s.split("\n"))
                    mg.drawString(5, i+=14, line, MapGraphics.StringAlign.LEFT);    
            }
```
I'd rather not keep an if and a for on the same line for readability

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #176, opened 2022-09-15
add monthly member and retunee login closing

anjeshshrestha:
need to test still


anjeshshrestha:
tested with monthly calendar, works fine

anjeshshrestha:
cant find a way to test with returnee login window

Pablete1234 (approved):

PR #181, opened 2022-10-07
Show next map in bot GUI

Punikekk:


src/main/java/com/github/manolo8/darkbot/core/manager/SettingsManager.java
```diff
@@ -36,6 +36,8 @@ public void tick() {
         this.config = API.readMemoryInt(address + 92);
 
         this.nextMap = API.readMemoryInt(address + 244);
+        this.main.mapManager.checkNextMap(nextMap);
+
```
Pablete1234, line 40:
```suggestion
```
Instead of settings manager having to reach into map manager, make map manager check this instead

Pablete1234 (changes requested):
Make it so MapManager#tick is what checks for settings manager value for next map, that makes settings manager stay a dummy without extra logic, and makes map manager not rely on external calls and manage its own state

Pablete1234 (approved):

PR #182, opened 2022-10-10
Show next map from Jump CPU

Punikekk:


src/main/java/com/github/manolo8/darkbot/core/objects/facades/SpaceMapWindowProxy.java
```diff
@@ -0,0 +1,30 @@
...
+
+        @Override
+        public void update() {
+            if (address != 0) {
+                duration = Main.API.readInt(address + 32);
+                mapId = Main.API.readInt(address + 36);
+            } else {
+                duration = 0;
+                mapId = 0;
+            }
```
Pablete1234, line 27: You should set these when update(address) is called with an address of 0, instead of every tick that update is called.
also simplifies so this logic can `if (addr == 0) return;` as first line

Pablete1234 (approved):

PR #185, opened 2022-10-14
Implement FlashMap

Punikekk:


src/main/java/com/github/manolo8/darkbot/core/objects/swf/FlashMap.java
```diff
@@ -0,0 +1,379 @@
...
+            clear();
+        }
+
+        this.address = address;
+    }
+
+    private int getCapacity(int logCapacity, @SuppressWarnings("unused") boolean hasIterIndexes) {
+        if (logCapacity <= 0) return 0;
+
+        logCapacity = 1 << (logCapacity - 1);
```
Pablete1234, line 107: this looks roughly equivalent to 2 ^ (logCapacity - 1) right? probably should point out in a comment what the more human-readable form of this is

src/main/java/com/github/manolo8/darkbot/core/objects/swf/FlashMap.java
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    private void readHashTable(long table) {
+        long atomsAndFlags = API.readLong(table);
+        long atoms = (atomsAndFlags & ByteUtils.ATOM_MASK) + 8;
+
+        int size = API.readMemoryInt(table + 8); // includes deleted items
+        int capacity = getCapacity(API.readMemoryInt(table + 12), (atomsAndFlags & kHasIterIndex) != 0);
+
+        if (size <= 0 || size > MAX_SIZE || capacity <= 0 || capacity > MAX_CAPACITY) {
```
Pablete1234, line 120: is this max size (of 1024) an actual hard-limit? i know in the past we've had to raise it on arrays, since players got over 512 (old limit) entities on the map and it would stop updating all together. we raised it to 8*1024 in that case, unsure about this use-case tho

src/main/java/com/github/manolo8/darkbot/core/objects/swf/FlashMap.java
```diff
@@ -0,0 +1,379 @@
...
+            this.size = 0;
+            return;
+        }
+        if (entries.length < size) entries = Arrays.copyOf(entries, size);
```
Pablete1234, line 124: This is bad for performance, gotta allocate extra space for new entries.

This is generally something ArrayList does, and so does our other array type, instead of allocating for size, you allocate for size *1.25 and that way you can keep growing without needing to create a new array and copy everything every time a new one is added.

src/main/java/com/github/manolo8/darkbot/core/objects/swf/FlashMap.java
```diff
@@ -0,0 +1,379 @@
...
+                System.out.println("Invalid keyKind! expected: " + this.keyKind + ", read: " + keyKind);
+                break;
+            }
+
+            long valueAtom = ByteUtils.getLong(BUFFER, (offset += 8));
+            AtomKind valueKind = AtomKind.of(valueAtom);
+            if (valueKind != this.valueKind) {
+                System.out.println("Invalid valueKind! expected: " + this.valueKind + ", read: " + valueKind);
+                break;
+            }
```
Pablete1234, line 149: i guess this is here just for validation but doesn't actually occur, right?

src/main/java/com/github/manolo8/darkbot/core/objects/swf/FlashMap.java
```diff
@@ -0,0 +1,379 @@
...
+
+    @Override
+    public V get(Object key) {
+        int i = indexOf(key);
+        return i == -1 ? null : entries[i].value;
+    }
+
+    @Override
+    public V put(K key, V value) {
+        if (!valueUpdatable) throw new IllegalStateException("V Type must be Updatable!");
```
Pablete1234, line 200: so PUT is only valid when the type is updatable? the error here seems not quite intuitive enough.
I think it should be an UnsupportedOperationException, not illegal state, and say put is not supported for maps which do not use an updatable as value type

src/main/java/com/github/manolo8/darkbot/core/objects/swf/FlashMap.java
```diff
@@ -0,0 +1,379 @@
...
+        updatables.put(key, new ValueWrapper(value));
+
+        return value;
+    }
+
+    @Override
+    public V remove(Object key) {
+        if (updatables == null) return null;
+        ValueWrapper v = updatables.remove(key);
+        return v == null ? null : v.value;
```
Pablete1234, line 211: This may violate some contract, you can call `get(a)` and get a response, but call `remove(a)` and get a null returned, if this is not a map of updatables

src/main/java/com/github/manolo8/darkbot/core/objects/swf/FlashMap.java
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    private int indexOf(Object key) {
+        for (int i = 0; i < size(); i++) {
+            Entry entry = entries[i];
+            if (entry.getKey() == null) continue;
+            if (entry.key.equals(key)) return i;
```
Pablete1234, line 218: `.getKey()`, or `.key`? pick one and stick to it

src/main/java/com/github/manolo8/darkbot/core/objects/swf/FlashMap.java
```diff
@@ -0,0 +1,379 @@
...
+
+        private @Nullable ValueWrapper removeReference() {
+            references--;
+            return null;
+        }
+
+        private @NotNull ValueWrapper addReference() {
+            references++;
+            return this;
+        }
```
Pablete1234, line 250: these methods should just be void. there's no reason to have a method that literally always returns null, nor for it to return this

PR #187, opened 2022-10-25
Added open hangar & open gg

Punikekk:


src/main/java/com/github/manolo8/darkbot/gui/titlebar/FlashRunnerTask.java
```diff
@@ -0,0 +1,81 @@
...
+    private static final Path RUNNER_PATH = Paths.get("lib", "KekkaRunner.exe");
+    private static final Pattern PARAMS_PATTERN = Pattern.compile("src\":.\"([^\"]+).*width\":.(\\d+).*height\":.(\\d+).*(cdn[^}]+)");
+
+    private static boolean LIB_CHECKED = false;
+
+    private final BackpageManager backpageManager;
+    private final JMenuItem menuItem;
+    private final String name;
+
+    public FlashRunnerTask(BackpageManager backpageManager, JMenuItem menuItem, String name) {
```
Pablete1234, line 25: Instead of passing a JMenuItem, i would pass a `Consumer<Boolean> callback`, and call with true/false for ok or fail. This unties the logic of the runner to the actuall trigger (the menu item/gui), and allows it to be called from more than just the menu.

Additionally, the default callback should be a `result -> SwingUtilities.invokeLater(() -> menu.setEnable(true))` because we need that enabled state change to happen on the swing thread

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #191, opened 2022-11-11
adding `MDEBUG` arg for get console in in IDE on plugin debug

fabio1999ita:
adding a manual debug way to plugin for get console from adding an arg


Pablete1234:
Superceeded by #196

PR #192, opened 2022-11-16
Method for get SID status

dm94:
This method is required and added in the API. I thought that isInstanceValid returned false if the SID was KO but not

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/backpage/BackpageManager.java
+++ b/src/main/java/com/github/manolo8/darkbot/backpage/BackpageManager.java
@@ -275,6 +275,11 @@ public class BackpageManager extends Thread implements BackpageAPI {
     }
 
     @Override
+    public int getSidStatus() {
+        return sidStatus;
+    }
+
+    @Override
     public String getSid() {
         return sid;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/manolo8/darkbot/backpage/BackpageManager.java
```diff
@@ -274,6 +274,11 @@ public boolean isInstanceValid() {
         return sid != null && instance != null && !sid.isEmpty() && !instance.isEmpty();
     }
 
+    @Override
```
Pablete1234, line 277: What is this overriding from? does this have to come alongside a DarkBotAPI change i'm missing?

dm94:
you're right, I'm adding it to the API now

src/main/java/com/github/manolo8/darkbot/backpage/BackpageManager.java
dm94: you are right then I ask for the change in the api

Pablete1234: The change in API, makes this sid status a String, i believe string is better, just return the result of `sidStat()`, which is either "--", "ERR", "OK" or "KO"

dm94: updated

Pablete1234 (approved):

PR #195, opened 2022-12-09
fix some guis not being toggled

Punikekk:


src/main/java/com/github/manolo8/darkbot/core/objects/LogoutGui.java
```diff
@@ -19,6 +27,18 @@ public void update() {
     @Override
     public boolean show(boolean value) {
         if (value) lastShown = System.currentTimeMillis();
+        if (value) {
```
Pablete1234, line 30: these can be merged

src/main/java/com/github/manolo8/darkbot/core/objects/Gui.java
```diff
@@ -102,12 +105,25 @@ public void moveWindow(int newX, int newY) {
...
+
             return false;
         }
         return value == visible && isAnimationDone();
     }
 
+    protected void legacyToggle(boolean show) {
+        if (show) clickMinimized();
+        else {
+            if (changeCloseClick || !clickMinimized()) {
```
Pablete1234, line 120:
```suggestion
        else if (changeCloseClick || !clickMinimized()) {
```

Pablete1234 (approved):

PR #196, opened 2022-12-10
Move general output in both direction, console and file

fabio1999ita:
new #191

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/Bot.java
+++ b/src/main/java/com/github/manolo8/darkbot/Bot.java
@@ -29,7 +29,7 @@ public class Bot {
             e.printStackTrace();
         }
 
-        LogUtils.setOutputToFile();
+        LogUtils.setOutputToFileAndConsole();
 
         try {
             UIManager.getFont("Label.font"); // Prevents a linux crash
--- a/src/main/java/com/github/manolo8/darkbot/Bot.java
+++ b/src/main/java/com/github/manolo8/darkbot/Bot.java
@@ -29,10 +29,10 @@ public class Bot {
             e.printStackTrace();
         }
 
-        //String path = Bot.class.getProtectionDomain().getCodeSource().getLocation().getPath();
-        //if (System.console() == null && (path.endsWith(".jar") || path.endsWith(".exe"))) {
-        LogUtils.setOutputToFile();
-        //}
+        String path = Bot.class.getProtectionDomain().getCodeSource().getLocation().getPath();
+        if (System.console() == null && (path.endsWith(".jar") || path.endsWith(".exe"))) {
+            LogUtils.setOutputToFile();
+        }
         try {
             UIManager.getFont("Label.font"); // Prevents a linux crash
             UIManager.setLookAndFeel(new FlatDarkLaf());
--- a/src/main/java/com/github/manolo8/darkbot/Bot.java
+++ b/src/main/java/com/github/manolo8/darkbot/Bot.java
@@ -29,7 +29,7 @@ public class Bot {
             e.printStackTrace();
         }
 
-        LogUtils.setOutputToFileAndConsole();
+        LogUtils.setupLogOutput();
 
         try {
             UIManager.getFont("Label.font"); // Prevents a linux crash
--- a/src/main/java/com/github/manolo8/darkbot/Bot.java
+++ b/src/main/java/com/github/manolo8/darkbot/Bot.java
@@ -29,10 +29,10 @@ public class Bot {
             e.printStackTrace();
         }
 
-        String path = Bot.class.getProtectionDomain().getCodeSource().getLocation().getPath();
-        if (System.console() == null && (path.endsWith(".jar") || path.endsWith(".exe"))) {
+        //String path = Bot.class.getProtectionDomain().getCodeSource().getLocation().getPath();
+        //if (System.console() == null && (path.endsWith(".jar") || path.endsWith(".exe"))) {
             LogUtils.setOutputToFile();
-        }
+        //}
         try {
             UIManager.getFont("Label.font"); // Prevents a linux crash
             UIManager.setLookAndFeel(new FlatDarkLaf());
--- a/src/main/java/com/github/manolo8/darkbot/Bot.java
+++ b/src/main/java/com/github/manolo8/darkbot/Bot.java
@@ -29,10 +29,10 @@ public class Bot {
             e.printStackTrace();
         }
 
-        String path = Bot.class.getProtectionDomain().getCodeSource().getLocation().getPath();
-        if (System.console() == null && (path.endsWith(".jar") || path.endsWith(".exe"))) {
-            LogUtils.setOutputToFile();
-        }
+        //String path = Bot.class.getProtectionDomain().getCodeSource().getLocation().getPath();
+        //if (System.console() == null && (path.endsWith(".jar") || path.endsWith(".exe"))) {
+        LogUtils.setOutputToFile();
+        //}
         try {
             UIManager.getFont("Label.font"); // Prevents a linux crash
             UIManager.setLookAndFeel(new FlatDarkLaf());
--- a/src/main/java/com/github/manolo8/darkbot/Bot.java
+++ b/src/main/java/com/github/manolo8/darkbot/Bot.java
@@ -29,10 +29,8 @@ public class Bot {
             e.printStackTrace();
         }
 
-        //String path = Bot.class.getProtectionDomain().getCodeSource().getLocation().getPath();
-        //if (System.console() == null && (path.endsWith(".jar") || path.endsWith(".exe"))) {
-            LogUtils.setOutputToFile();
-        //}
+        LogUtils.setOutputToFile();
+
         try {
             UIManager.getFont("Label.font"); // Prevents a linux crash
             UIManager.setLookAndFeel(new FlatDarkLaf());
--- a/src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
+++ b/src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
@@ -1,5 +1,6 @@
 package com.github.manolo8.darkbot.utils;
 
+import com.github.manolo8.darkbot.utils.itf.ThrowingConsumer;
 import org.jetbrains.annotations.NotNull;
 
 import java.io.BufferedOutputStream;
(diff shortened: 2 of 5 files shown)


src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -24,13 +26,27 @@ public static void setOutputToFile() {
...
 
-        try {
+        /*try {
             PrintStream output = getLogger();
             System.setOut(output);
             System.setErr(output);
         } catch (FileNotFoundException | UnsupportedEncodingException e) {
             System.out.println("Failed to redirect logs, file not found:");
             e.printStackTrace();
+        }*/
```
Pablete1234, line 36: Cleanup after yourself, don't leave these around

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -24,13 +26,27 @@ public static void setOutputToFile() {
...
+        try {
+            FileOutputStream fileOut = new FileOutputStream("logs/" + START_TIME + ".log");
```
Pablete1234, line 38: This is not equivalent to the current behavior, log modifications aren't being appiled here for file output, see getLogger method

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -29,10 +29,10 @@ public static void main(String[] args) throws IOException {
             e.printStackTrace();
         }
 
-        String path = Bot.class.getProtectionDomain().getCodeSource().getLocation().getPath();
-        if (System.console() == null && (path.endsWith(".jar") || path.endsWith(".exe"))) {
+        //String path = Bot.class.getProtectionDomain().getCodeSource().getLocation().getPath();
+        //if (System.console() == null && (path.endsWith(".jar") || path.endsWith(".exe"))) {
```
Pablete1234, line 33: same as before, cleanup after yourself, don't leave commented code around if it's not a temporal thing

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -111,4 +128,42 @@ public void println(String string) {
...
+        }
+
+        @Override
+        public void write(int b) throws IOException {
+            for (OutputStream out : outputStreams)
+                out.write(b);
+        }
+
+        @Override
+        public void write(byte @NotNull [] b) throws IOException {
```
Pablete1234, line 146:
```suggestion
        public void write(@NotNull byte[] b) throws IOException {
```

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -111,4 +128,42 @@ public void println(String string) {
             super.println(string);
         }
     }
+
+    public static class MultiOutputStream extends OutputStream {
+        OutputStream[] outputStreams;
```
Pablete1234, line 133:
```suggestion
        private final OutputStream[] outputStreams;
```

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -111,4 +128,42 @@ public void println(String string) {
...
+        }
+
+        @Override
+        public void write(byte @NotNull [] b) throws IOException {
+            for (OutputStream out : outputStreams)
+                out.write(b);
+        }
+
+        @Override
+        public void write(byte @NotNull [] b, int off, int len) throws IOException {
```
Pablete1234, line 152:
```suggestion
        public void write(@NotNull byte[] b, int off, int len) throws IOException {
```

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -111,4 +128,42 @@ public void println(String string) {
             super.println(string);
         }
     }
+
+    public static class MultiOutputStream extends OutputStream {
```
Pablete1234, line 132: While the impl is generally fine, there's one critical issue, if any output stream fails all others will not get the data at all.

So if the 1st output throws an IOException, no other one gets the data.

You should try-catch all calls in the loop, and then maybe either bubble-up the exception, or wrap multiple exceptions around another one to bubble them up, unsure

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -20,16 +24,27 @@ public class LogUtils {
...
-            System.setOut(output);
-            System.setErr(output);
+            OutputStream consoleOut = new FileOutputStream(FileDescriptor.out);
+            OutputStream consoleErr = new FileOutputStream(FileDescriptor.err);
+
+            OutputStream file = new FileOutputStream("logs/" + START_TIME + ".log");
+
+            OutputStream multiOut = new MultiOutputStream(consoleOut, file);
+            OutputStream bufferedOut = new BufferedOutputStream(multiOut, 128);
+            PrintStream printStreamOut = new PrintStreamWithDate(bufferedOut);
```
Pablete1234, line 39: This set of 3 lines is repeated, as well as the consoleOut/consoleErr, i'd put them inside a function, so that we can simplify this method to:
```java
OutputStream fileLogger = new FileOutputStream("logs/" + START_TIME + ".log");

System.setOut(createPrintStream(FileDescriptor.out, file));
System.setErr(createPrintStream(FileDescriptor.err, file));
```

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -20,16 +24,27 @@ public class LogUtils {
...
+
+            OutputStream multiErr = new MultiOutputStream(consoleErr, file);
+            OutputStream bufferedErr = new BufferedOutputStream(multiErr, 128);
+            PrintStream printStreamErr = new PrintStreamWithDate(bufferedErr);
+
+            System.setOut(printStreamOut);
+            System.setErr(printStreamErr);
         } catch (FileNotFoundException | UnsupportedEncodingException e) {
-            System.out.println("Failed to redirect logs, file not found:");
```
Pablete1234, line 32: Keep this

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -90,8 +101,9 @@ private static void tryDelete(Path path) {
     }
 
     private static class PrintStreamWithDate extends PrintStream {
-        public PrintStreamWithDate(String logfile) throws FileNotFoundException, UnsupportedEncodingException {
-            super(logfile, "UTF-8");
+
+        public PrintStreamWithDate(OutputStream downstream) throws FileNotFoundException, UnsupportedEncodingException {
+            super(downstream, true, "UTF-8");
```
Pablete1234, line 106: You're assuming everyone's system works in UTF-8, avoid such assumption.

Java internally uses:
```
        setOut0(newPrintStream(fdOut, props.getProperty("sun.stdout.encoding")));
        setErr0(newPrintStream(fdErr, props.getProperty("sun.stderr.encoding")));
```

those `props` you can get via `System.getProperties().getProperty("sun.stdout.encoding")`.
If they're non-null, call with that encoding. If they're null, call a constructor without that encoding.

You can double-check all the java code in the System class:
image
image

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -111,4 +123,43 @@ public void println(String string) {
             super.println(string);
         }
     }
+
+    public static class MultiOutputStream extends OutputStream {
+        private final OutputStream[] outputStreams;
+        int count = 0;
```
Pablete1234, line 129:
```suggestion
```
You're not using this anymore

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
Pablete1234: This continues being a relevant point

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -20,20 +25,36 @@ public class LogUtils {
...
-            System.setOut(output);
-            System.setErr(output);
-        } catch (FileNotFoundException | UnsupportedEncodingException e) {
-            System.out.println("Failed to redirect logs, file not found:");
+            OutputStream fileLogger = new FileOutputStream("logs/" + START_TIME + ".log");
+
+            System.setOut(createPrintStream(new FileOutputStream(FileDescriptor.out), fileLogger));
+            System.setErr(createPrintStream(new FileOutputStream(FileDescriptor.err), fileLogger));
+        } catch (FileNotFoundException e) {
+            System.out.println("Failed to redirect logs, file not found: " + "logs/" + START_TIME + ".log");
```
Pablete1234, line 38: File not found would be for the parent folder, not the log. Regardless, you shouldn't assume the exception source, just either use e.getMessage, or let the following printStackTrace explain the *actusl* error

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -20,20 +25,36 @@ public class LogUtils {
...
             e.printStackTrace();
         }
     }
 
+    private static PrintStream createPrintStream(FileOutputStream multi, OutputStream fileLogger) {
+        OutputStream multiOut = new MultiOutputStream(multi, fileLogger);
+        OutputStream bufferedOut = new BufferedOutputStream(multiOut, 128);
+
+        try {
+            if (System.getProperty("file.encoding").equals(Charset.defaultCharset().toString()))
```
Pablete1234, line 48: You're ignoring `sun.stout.encoding` and the equivalent one for sterr.
Additionally, this check would always be true as default charset comes from that property

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -20,20 +25,36 @@ public class LogUtils {
...
+                return new PrintStreamWithDate(bufferedOut, "UTF-8");
```
Pablete1234, line 49: Don't hard-code utf-8, respect the OS's encoding.
In windows the default isnt utf8 becsuse the windows console doesnt support it, and it would break with this change.

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/utils/LogUtils.java
```diff
@@ -111,4 +133,62 @@ public void println(String string) {
...
+        }
+
+        @Override
+        public void write(int b) throws IOException {
+            try {
+                for (OutputStream out : outputStreams)
+                    out.write(b);
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
```
Pablete1234, line 151: these try-catch + for need to be reversed to for + try

Otherwise one output stream failing means the other ones never get the data

Pablete1234 (approved):

PR #199, opened 2022-12-12
add nova manager

anjeshshrestha:
Need to add unEquip agent, testing in progress


src/main/java/com/github/manolo8/darkbot/backpage/nova/Agent.java
```diff
@@ -0,0 +1,290 @@
...
+        public int getAmount() {
+            return amount;
+        }
+
+        public void setAmount(int amount) {
+            this.amount = amount;
+        }
+    }
+
+    public class Perk {
```
dm94, line 146: Separate this class in another file

src/main/java/com/github/manolo8/darkbot/backpage/nova/Agent.java
```diff
@@ -0,0 +1,290 @@
...
+            return upgrades;
+        }
+
+        public void setUpgrades(List<Upgrade> upgrades) {
+            this.upgrades = upgrades;
+        }
+
+    }
+
+    public class Cost {
```
dm94, line 240: Separate this class in another file

src/main/java/com/github/manolo8/darkbot/backpage/nova/Agent.java
```diff
@@ -0,0 +1,290 @@
...
+        public String getItem() {
+            return item;
+        }
+
+        public void setItem(String item) {
+            this.item = item;
+        }
+    }
+
+    public class Upgrade {
```
dm94, line 261: Separate this class in another file

dm94 (changes requested):
Separate these classes into different files

dm94 (approved):

src/main/java/com/github/manolo8/darkbot/backpage/NovaManager.java
```diff
@@ -0,0 +1,205 @@
...
+public class NovaManager {
+    private final BackpageManager backpageManager;
+    private final NovaData data;
+    private final Gson g;
+    private long lastNovaUpdate;
+
+    public NovaManager(BackpageManager backpageManager) {
+        this.backpageManager = backpageManager;
+        this.data = new NovaData();
+        this.g = new Gson();
```
Pablete1234, line 21: any reason for using a new gson instance here?
we have a few scattered around the codebase by now, probably should try to reuse them more

src/main/java/com/github/manolo8/darkbot/backpage/NovaManager.java
```diff
@@ -0,0 +1,205 @@
...
+            this.data.setResourceAmount(jsonObj.get("item").getAsJsonObject().get("amount").getAsInt());
+        } else {
+            System.out.println("NovaManager: " + response);
+        }
+    }
+
+    /**
+     * Method to get list of Nova Agent/Captain
+     * @throws IOException
+     */
```
Pablete1234, line 91: these comments are pretty meaningless with such vague descriptions
also, it's straight up wrong as this doesn't *get a list*, it doesn't return anything, it just updates them

src/main/java/com/github/manolo8/darkbot/backpage/NovaManager.java
```diff
@@ -0,0 +1,205 @@
...
+        return this.data;
+    }
+
+
+    /**
+     * This method is used to unequip, equip and get current captain
+     * @param id
+     * @throws IOException
+     */
+    public boolean updateActiveCaptain(int id) throws IOException {
```
Pablete1234, line 54: id? what id?

Captain id? if so name it properly, and if you're doing javadocs explain what the id is in the `@param` don't just leave it empty

src/main/java/com/github/manolo8/darkbot/backpage/NovaManager.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public boolean buyNova(int amount) throws IOException {
+        String response = backpageManager.getConnection("ajax/shop.php", Method.POST)
+                .setRawParam("action", "purchase")
+                .setRawParam("category", "special")
+                .setRawParam("itemId", "captain_captain-generic")
+                .setRawParam("amount", amount)
+                .setRawParam("level", "")
+                .setRawParam("selectedName", "")
+                .getContent();
```
Pablete1234, line 150: use setParam over setRawParam unless you know what you're doing.

setParam will convert things like spaces into the url-encoded versions, which is what you'd want if you had spaces

src/main/java/com/github/manolo8/darkbot/backpage/NovaManager.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public boolean upgradeAgentPerk(Agent agent, Perk perk, int upgradeLevel) throws IOException {
+        String response = backpageManager.getConnection("ajax/captain.php", Method.POST)
+                .setRawParam("command", "upgradePerk")
+                .setRawParam("captainId", agent.getCaptainId())
+                .setRawParam("perkId", perk.getPerkId())
+                .setRawParam("upgradeLevel", upgradeLevel)
+                .getContent();
+        this.update(0);
+        JsonObject jsonObj = g.fromJson(response, JsonObject.class); //Converts the json string to JsonElement without POJO
```
Pablete1234, line 202:
```suggestion
        JsonObject jsonObj = g.fromJson(response, JsonObject.class);
```
those comments are somewhat redundant

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/backpage/BackpageManager.java
```diff
@@ -60,14 +62,18 @@ protected static class SidStatus {
     private int userId;
     private Optional<LoginData> loginData;
 
+    private Gson gson;
```
Pablete1234, line 65: Consider initializing it directly here instead of the constructor, makes it so the constructors on all the other apis automatically get it, and you can remove the null-check from getGson method
```suggestion
    private final Gson gson = new Gson();
```
I'd also move it to the top of the class.

src/main/java/com/github/manolo8/darkbot/backpage/NovaManager.java
```diff
@@ -0,0 +1,163 @@
...
+import com.github.manolo8.darkbot.utils.http.Method;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+
+import java.io.IOException;
+
+public class NovaManager {
+    private final BackpageManager backpageManager;
+    private final NovaData data;
+    private final Gson g;
```
Pablete1234, line 15: avoid one-letter names in fields, they're fine in lambdas or small-scoped things, but i'd use the more descriptive ` `gson` here

Pablete1234 (approved):

PR #201, opened 2022-12-13
run from enemy tag

Punikekk:


dm94 (changes requested):
Where is the logic in avoiding these players?

Pablete1234:
> Where is the logic in avoiding these players?

in https://github.com/darkbot-reloaded/DarkBotAPI/pull/52

Pablete1234 (approved):

PR #202, opened 2022-12-16
check last internet read time & TemporalModule fix attempt

Punikekk:


Pablete1234:
For the temp module fix i believe this is a better approach: https://github.com/darkbot-reloaded/DarkBotAPI/pull/54/

Pablete1234 (approved):

PR #203, opened 2022-12-17
Use sid & instance uri from LoginData

Punikekk:


src/main/java/com/github/manolo8/darkbot/backpage/BackpageManager.java
```diff
@@ -298,4 +311,12 @@ public Instant getLastRequestTime() {
...
         return Optional.ofNullable(getReloadToken(body));
     }
+
+    private Optional<LoginData> loginData;
+    public void setLoginData(LoginData loginData) {
+        if (this.loginData != null)
+            throw new IllegalStateException("LoginData can be assigned only once!");
+
+        this.loginData = Optional.ofNullable(loginData);
+    }
```
Pablete1234, line 321: move the field to alongside the rest of the fields, and i'd put the setter quite higher too, probs right below the constructor

Pablete1234 (approved):

PR #206, opened 2022-12-23
avoid plugin already in use by other bugged java process

fabio1999ita:
Sometimes when the java process closes, it does not actually release/unlock the files for the plugins, so later when a new bot instance runs it is prevented from replacing those plugin files (eg: when trying to update plugins). If that happens, you're left with needing to restart your PC (or killing the process) before being able to update the plugins.

This pr attempts to fix it by making sure the class loader is closed on shutdown


fabio1999ita:
sometimes java bot process not full close and when you start a new bot the older bot block plugin, so in plugin section you get failed to load or something like this because they are already used in the bugged old process

Pablete1234 (approved):

PR #207, opened 2022-12-25
Nove update

anjeshshrestha:
when a error occured calling get rosterlist, it would remove all nova agents, thus also deleting them on plugin side

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/backpage/NovaManager.java
+++ b/src/main/java/com/github/manolo8/darkbot/backpage/NovaManager.java
@@ -24,18 +24,16 @@ public class NovaManager {
     public boolean update(long expiryTime) {
         try {
             if (System.currentTimeMillis() <= lastNovaUpdate + expiryTime) return false;
-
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/manolo8/darkbot/backpage/NovaManager.java
```diff
@@ -84,13 +85,14 @@ private void updateRosterList() throws IOException {
         if (jsonObj.get("result").getAsString().equalsIgnoreCase("OK")) {
             String resp = jsonObj.getAsJsonArray("rosterList").toString();
             Agent[] rosterList = gson.fromJson(resp, Agent[].class);
-
             for (Agent agent : rosterList) {
                 this.data.addAgent(agent);
             }
+            return false;
```
Pablete1234, line 91: shouldn't it be true here for "properly updated" ? what does true/false mean in this method?

anjeshshrestha: haha, i think too fast!!!

Pablete1234 (approved):

PR #208, opened 2022-12-25
Optimize trail generation to avoid garbage collection

Pablete1234:
Currently trails on the bot are generated by saving a TreeMap of Lines, they're added as needed, and the entries older than X time (configured) are pruned. This generates an unnecessarily large amount of allocations and deletions.

This PR implements a new data structure, a "recycling queue". It consists of a singly-linked list, with two heads and one tail, which works strictly in FIFO mode. You may only add elements at the back, and you may only retrieve/remove elements from the front.

When adding elements (at the back) the tail node gets a new next node, and tail starts pointing to the new node
When removing elements (from the front) the head pointer starts pointing to the next node
The "original head" pointer is used to keep strong a reference to the "removed" elements, as what essentially is a second queue, both sharing the same chain. When a new element is requested (to be added at the back), the "original head" is checked first, if it has leftover elements, then original head moves forwards (removes the element from that logical second queue) and the node is added at the back. If the "original head" reaches the head, it cannot advance anymore and instead allocates a new node to supply the demand for a new node.

Aditonally, when removing elements if the "overall size" (sum of all allocated nodes) far exceeds the amount of nodes actually in use, then the "original head" will be moved over to deallocate some of the objects. There's a ratio of a maximum of 50% extra buffer (eg: if size is 1000, you may only keep up to 500 extra as buffer).

This accomplishes a solution similar to one of a circular buffer, but of dynamic size and implemented as a single-linked list instead of an array (which is more expensive to grow/shrink).

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/gui/trail/Line.java
+++ b/src/main/java/com/github/manolo8/darkbot/gui/trail/Line.java
@@ -60,8 +60,8 @@ public class Line {
         Line last = null;
 
         for (Line line : lines) {
-            if (last == null || last.to != line.from) {
-                paths.add(current = new ArrayList<>(lines.size()));
+            if (last == null || last.to.equals(line.from)) {
+                paths.add(current = new ArrayList<>(lines.size() + 1));
                 current.add(line.getFrom());
             }
             last = line;
--- a/src/main/java/com/github/manolo8/darkbot/gui/trail/Line.java
+++ b/src/main/java/com/github/manolo8/darkbot/gui/trail/Line.java
@@ -1,45 +1,59 @@
 package com.github.manolo8.darkbot.gui.trail;
 
+import com.github.manolo8.darkbot.utils.data.SizedIterable;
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/manolo8/darkbot/gui/trail/Line.java
```diff
@@ -1,53 +1,67 @@
...
      */
-    private static List<List<Location>> getPaths(Collection<Line> lines) {
+    private static List<List<Location>> getPaths(SizedIterable<Line> lines) {
         List<List<Location>> paths = new ArrayList<>();
 
         List<Location> current = null;
         Line last = null;
 
         for (Line line : lines) {
             if (last == null || last.to != line.from) {
```
Punikekk, line 63: You cannot compare Location instances anymore

Pablete1234: true, forgot to push that change

PR #210, opened 2022-12-27
disable `checkInvalid` method when -no-op

fabio1999ita:
avoid annoying output of `checkInvalid` method when -no-op is on

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/BotInstaller.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/BotInstaller.java
@@ -129,22 +129,24 @@ public class BotInstaller implements API.Singleton {
 
     private long lastInternetRead;
     private void checkInvalid() {
-        if (API.hasCapability(GameAPI.Capability.HANDLER_INTERNET_READ_TIME)) {
-            long lastRead = API.lastInternetReadTime();
-            if (lastInternetRead != lastRead) {
-                lastInternetRead = lastRead;
-                invalidTimer.activate(20_000);
-            } else if (!invalidTimer.isArmed()) invalidTimer.activate(20_000);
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/manolo8/darkbot/core/BotInstaller.java
```diff
@@ -129,22 +129,24 @@ private boolean settingsPattern(long address) {
...
-        } else if (!invalidTimer.isArmed()) invalidTimer.activate(150_000); // 2.5 min
-
-        // timer is disarmed on refresh and on valid tick
-        if (invalidTimer.tryDisarm()) {
-            if (API.hasCapability(GameAPI.Capability.HANDLER_CLEAR_CACHE))
-                API.clearCache(".*");
-
-            API.handleRefresh();
-            System.out.println("Triggering refresh: stuck at loading screen for too long!");
+        if (API.hasCapability(GameAPI.Capability.LOGIN)) {
```
Pablete1234, line 132: `LOGIN` is the wrong capability to check here.

You may support logging in, but not support reloading (eg: backpage-only api)
You may not support logging in, but support reloading (eg: original browser api, where you logged-in the browser)

So this is definitely NOT the capability that correctly maps if you're able to reload or not.
There's a `BACKGROUND_ONLY` capability that has already been used elsewhere in the codebase precisely to gate-out these "invalid" checks.

```suggestion
        if (API.hasCapability(GameAPI.Capability.BACKGROUND_ONLY)) return;
```
Also i recommend you use a one-liner return to avoid indenting-in the rest of the method

Pablete1234 (changes requested):

Pablete1234:
Implemented it as part of c0f281d

PR #218, opened 2023-01-05
update configs every time that combobox is opened

fabio1999ita:
that is usefull when you make a manually copy of a config in folder of configs and you get it in the bot just opening the combobox of configs without make trick (like add a fake config) or restart bot

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/gui/titlebar/ConfigPicker.java
+++ b/src/main/java/com/github/manolo8/darkbot/gui/titlebar/ConfigPicker.java
@@ -9,6 +9,8 @@ import com.github.manolo8.darkbot.utils.I18n;
 import net.miginfocom.swing.MigLayout;
 
 import javax.swing.*;
+import javax.swing.event.PopupMenuEvent;
+import javax.swing.event.PopupMenuListener;
 import javax.swing.plaf.basic.ComboPopup;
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/manolo8/darkbot/gui/titlebar/ConfigPicker.java
```diff
@@ -40,6 +42,19 @@ public ConfigPicker() {
         list.addMouseListener(renderer);
         list.addMouseMotionListener(renderer);
 
+        addPopupMenuListener(new PopupMenuListener() {
+            @Override
+            public void popupMenuWillBecomeVisible(PopupMenuEvent e) {
+                setup(main);
```
Pablete1234, line 48: setup is a function that may be potentially blocking, as it's listing the files in a directory, and that may take an indeterminate amount of time (eg: imagine you run off of a remote disk). That's an issue as the GUI may freeze every time you use the dropdown.

Since the issue is likely to never actually be of much affect i'm going to ignore it, but it'd be nice to have it fixed so that configManager#getAvailableConfigs is called using a swing worker, and then, when done, on the swing thread, you can just re-create the list. That way you open the drop down, and as soon as the update is ready you'll see the new entries appear in the list

Pablete1234 (approved):

PR #222, opened 2023-01-07
add auto updater inside bot

fabio1999ita:


Pablete1234:
How is this going to interact with anyone who renames their bot, or has multiple of them? ie: DarkBot1.jar, DarkBot2.jar or similar?

PR #223, opened 2023-01-07
Added backpage browser & BackpageButton

Punikekk:


src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,264 @@
...
+                    new ProcessBuilder(BACKPAGE_PATH.resolve(EXECUTABLE_NAME).toAbsolutePath().toString(),
+                            "--sid", main.backpage.getSid(),
+                            "--url", main.backpage.getInstanceURI().toString())
+                            .start().waitFor();
+                }
+            } catch (InterruptedException | IOException e) {
+                e.printStackTrace();
+            }
+        } // inform user that instance/sid is not valid?
+        SwingUtilities.invokeLater(() -> button.setEnabled(true));
```
Pablete1234, line 68: given the amount of invoke-later's i'm seeing, consider using a SwingWorker instead

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,264 @@
...
+        Popups.Builder builder = Popups.of("Backpage browser", message)
+                .messageType(JOptionPane.QUESTION_MESSAGE)
+                .options(download, I18n.get("gui.backpage_button.cancel"));
+
+        JOptionPane optionPane = builder.build();
+        download.addActionListener(l -> optionPane.setValue(download));
+
+        if (download == builder.showSync()) {
+            if (!clearDirectory())
+                throw new DirectoryClearException(BACKPAGE_PATH);
```
Pablete1234, line 125: This just seems a bit odd, can't just clear directory throw such exception if it needs to?

Creating an exception that gets thrown in one single case and it handled in a generic block is just a bit odd to me, i get the reason here is so that you can show the user something relevant, but couldn't you just use a plain IOException and use the "message" parameter and show that in the pop-up?

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,264 @@
...
+            }
+        } // inform user that instance/sid is not valid?
+        SwingUtilities.invokeLater(() -> button.setEnabled(true));
+    }
+
+    private boolean canRun() {
+        try {
+            Version currentVersion = readVersionFile();
+
+            // have downloaded backpage and version was checked recently
```
Pablete1234, line 75: I think some of the comments are too obvious and maybe shouldn't be there

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,264 @@
...
+        if (Files.notExists(BACKPAGE_PATH)) return true;
+
+        try (Stream<Path> walk = Files.walk(BACKPAGE_PATH)) {
+            boolean backapgeFileAccessible = walk.sorted(Comparator.reverseOrder())
+                    .filter(path -> path.toString().contains(EXECUTABLE_NAME))
+                    .map(Path::toFile)
+                    .map(File::delete)
+                    .findFirst().orElse(true);
+
+            if (!backapgeFileAccessible) return false; // probably used by another process - cannot delete
```
Pablete1234, line 146:
```suggestion
            // Failed to delete, likely being used by another process
            if (!backapgeFileAccessible) return false;
```
Avoid trailing comments when possible, prefer placing them above the statement. This comment in particular is probably the only one that explains something non-obvious, make sure to keep it

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,264 @@
...
+        Version remoteVersion = releaseInfo.getVersion();
+        ReleaseInfo.Asset asset = releaseInfo.getValidAsset();
+        if (remoteVersion == null || asset == null) return false;
+
+        VERSION_CHECK_TIMER.activate(); // activate the timer here - version will be successfully checked
+        return current != null && current.compareTo(remoteVersion) >= 0
+                || askUserToDownload(progressBar, current, remoteVersion, asset);
+    }
+
+    //return true if is possible to run a backpage
```
Pablete1234, line 108: If you're looking to document what a return is... may as well do it with javadoc format

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,264 @@
...
+        });
+
+        try (ZipInputStream zis = new ZipInputStream(new BufferedInputStream(asset.openStream()))) {
+            int downloadedBytes = 0;
+            for (ZipEntry entry = zis.getNextEntry(); entry != null; entry = zis.getNextEntry()) {
+                Path to = BACKPAGE_PATH.resolve(entry.getName());
+
+                if (entry.isDirectory())
+                    Files.createDirectories(to);
+                else {
```
Pablete1234, line 170:
```suggestion
                if (entry.isDirectory()) {
                    Files.createDirectories(to);
                    continue;
                }
```
always add braces if it's not saving any lines at all
Also i think it'll be clearer if you un-indent the rest of this else statement

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,264 @@
...
+        JOptionPane optionPane = builder.build();
+        download.addActionListener(l -> optionPane.setValue(download));
+
+        if (download == builder.showSync()) {
+            if (!clearDirectory())
+                throw new DirectoryClearException(BACKPAGE_PATH);
+
+            FileUtils.createDirectories(BACKPAGE_PATH);
+            downloadBackpage(progressBar, asset);
+            writeVersionFile(remote);
```
Pablete1234, line 129: given this is a one-line method that is called in exactly one place and isn't particularily long, i'd say it could be inlined

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        return readVersionFile() != null;
```
Pablete1234, line 132: reading the file here seems pretty unnecessary. If the above if got to writeFileVersion, you can just return true directly.
If you get here via any other code-path, you can just `return current != null`

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,264 @@
...
+                        break;
+                    case MACOS:
+                        if (downloadUrl.contains("mac-x64.zip")) return true;
+                        break;
+                    case LINUX:
+                        if (downloadUrl.contains("linux-x64.zip")) return true;
+                        break;
+                }
+
+                return false;
```
Pablete1234, line 246: Consider adding an OSUtil.getShortName() that returns win, mac, or linux, then this can be simplified:
```suggestion
                return downloadUrl.contains(OSUtil.getShortName() + "-x64.zip");
```

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,206 @@
...
+        if (Files.notExists(VERSION_PATH)) return null;
+        return new Version(Files.readString(VERSION_PATH));
+    }
+
+    public static class ReleaseInfo {
+        public String name;
+        public List<Asset> assets;
+
+        public static ReleaseInfo get() throws IOException, JsonParseException {
+            return new Gson().fromJson(Http.create(RELEASE_URL).getContent(), ReleaseInfo.class);
```
Pablete1234, line 177: Reuse one of the gson instances we've already created, probably the one in Main is the best one to reuse.

Additionally, although not required since this is something we're already doing wrong in so many places, it'd be nice if you used the reader-based version of gson reading, as it doesn't need to allocate the full string (as getContent does):
```
return Http.create(RELEASE_URL)
        .consumeInputStream(is -> Main.GSON.fromJson(new InputStreamReader(is), ReleaseInfo.class));
```
probs would be nice if we had a consumeReader in http, or even built-in `toJson(Type/Class)` directly implemented in Http

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,206 @@
...
+        SwingUtilities.invokeLater(() -> progressBar.setValue(progress));
+    }
+
+    private @Nullable Version readVersionFile() throws IOException {
+        if (Files.notExists(VERSION_PATH)) return null;
+        return new Version(Files.readString(VERSION_PATH));
+    }
+
+    public static class ReleaseInfo {
+        public String name;
```
Pablete1234, line 173: Given that it seems like this is just turned into a Version right underneath, why not deserialize it as one?
```suggestion
        @SerializedName("name")
        public Version version;
```

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,206 @@
...
+            return new Version(name);
+        }
+
+        public Asset getValidAsset() {
+            return assets.stream()
+                    .filter(ReleaseInfo.Asset::isSupported)
+                    .findFirst().orElse(null);
+        }
+
+        public static class Asset {
```
Pablete1234, line 190: i'd advise against nesting this inside release info, BackpageTask.ReleaseInfo.Asset is just... not ideal, it can be just BackpageTask.Asset

src/main/java/com/github/manolo8/darkbot/gui/titlebar/BackpageTask.java
```diff
@@ -0,0 +1,206 @@
...
+    private @Nullable Version readVersionFile() throws IOException {
+        if (Files.notExists(VERSION_PATH)) return null;
+        return new Version(Files.readString(VERSION_PATH));
+    }
+
+    public static class ReleaseInfo {
+        public String name;
+        public List<Asset> assets;
+
+        public static ReleaseInfo get() throws IOException, JsonParseException {
```
Pablete1234, line 176: I think i would rather have releaseInfo as a just-data object, and have a `static ReleaseInfo getLatestRelease()` in BackpageTask directly

Pablete1234 (changes requested):
Almost ready to merge, just a few minor changes.
Also please rebase the branch and you'll get the new home.svg icon automatically, the home.png shouldn't be required anymore

Pablete1234 (approved):

PR #226, opened 2023-01-12
add plutus status

anjeshshrestha:
Add plutus proxy, has 1 field to get status of plutus


src/main/java/com/github/manolo8/darkbot/core/objects/facades/PlutusProxy.java
```diff
@@ -0,0 +1,18 @@
+package com.github.manolo8.darkbot.core.objects.facades;
+
+import com.github.manolo8.darkbot.core.itf.Updatable;
+import com.github.manolo8.darkbot.core.utils.ByteUtils;
+
+import static com.github.manolo8.darkbot.Main.API;
+
+public class PlutusProxy extends Updatable implements eu.darkbot.api.API.Singleton {
```
Pablete1234, line 8: Would be nice if instead of just api.singleton, this implemented an actual interface on the api so that it's available in the plugin api

Pablete1234 (changes requested):
Please separate changes for auction to a different pr

Pablete1234 (approved):
It's still missing an api update in order to compile, but i'll do that in a follow-up commit

PR #228, opened 2023-01-14
Added option to roam to radiation zone

pajtor:


src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -118,6 +118,7 @@ public static class Roaming {
             public @Option boolean SEQUENTIAL = false;
             public @Option boolean ONLY_KILL_PREFERRED = false;
             public @Option boolean ENEMY_CBS_INVISIBLE = false;
+            public @Option boolean RADIATION = false;
```
Pablete1234, line 121: This has nothing to do with *roaming*, roaming is for the bot just randomly moving around the map, not for traveling with a specific destination.

Put this setting in the misc category alongside Avoid mines, as "AVOID_RADIATION", and rename it accoordingly (default to true)

src/main/java/com/github/manolo8/darkbot/core/manager/MapManager.java
```diff
@@ -469,6 +472,7 @@ public boolean setTarget(long entity) {
     }
 
     public boolean isOutOfMap(double x, double y) {
+        if (main.config.GENERAL.ROAMING.RADIATION || radiation.getValue()) return false;
         return x < 0 || y < 0 || x > internalWidth || y > internalHeight;
```
Pablete1234, line 476: This code is checking for the value of the config... twice

```suggestion
        return avoidRadiation.getValue() && (x < 0 || y < 0 || x > internalWidth || y > internalHeight);
```

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/manager/MapManager.java
pajtor: the reason is that I dont know which one should be prioritized. I mean which way of getting the value

Pablete1234: Inside bot code, for just reading, either is fine. If read & write are used then it should be done via the ConfigSetting api. Regardless, i think new api is cleaner as you define it once then use it as a field of your own class.

src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -196,6 +196,8 @@ public static class Miscellaneous {
         public @Option @Visibility(Level.INTERMEDIATE) boolean LOG_CHAT = false;
         public @Option @Visibility(Level.INTERMEDIATE) boolean LOG_DEATHS = false;
         public @Option @Visibility(Level.INTERMEDIATE) boolean AVOID_MINES = true;
+
```
Pablete1234, line 199:
```suggestion
```

Pablete1234 (changes requested):
Have you tested the change? I believe this in itself will not re-create path points meaning if you toggle the setting on/off it won't take effect until an obstacle is added/removed or you jump map.
Consider PathFinder registering a change listener on the setting to invalidate the current path points.

pajtor:
> Have you tested the change? I believe this in itself will not re-create path points meaning if you toggle the setting on/off it won't take effect until an obstacle is added/removed or you jump map. Consider PathFinder registering a change listener on the setting to invalidate the current path points.

I checked the code, and if a module calls method getClosestDistance() then the change will be registered every tick

Pablete1234:
I'm not talking about if it will be able to enter radiation, i'm talking about if it's able to re-route thru radiation
enable dev mode and set a map zoom small enough that barrier corners outside radiation are shown, and see if path points are added/removed when just changing the setting and moving, or if it requires you adding/removing some avoided zone for the cache to update

Pablete1234 (approved):

PR #232, opened 2023-01-21
fix kicking user from group

anjeshshrestha:
this change will add kick user to pending task, thus keeping the group window open.

The offset for header was not added to kickUser click, it has been added now


src/main/java/com/github/manolo8/darkbot/core/manager/GroupManager.java
```diff
@@ -159,28 +161,81 @@ public void sendInvite(String username, long wait) {
...
+        GroupMember member = group.getMember(id);
+        kick(member);
     }
 
     public void kick(GroupMember member) {
-        if (member != null && canKick()) kickUser(group.indexOf(member));
+        if (member == null) return;
+        if (pending != null || !canKick()) return;
+        int idx = group.indexOf(member);
+        if (idx < 0) return;
```
Pablete1234, line 180: needs to be inside the task itself (the pending)

src/main/java/com/github/manolo8/darkbot/core/manager/GroupManager.java
```diff
@@ -159,28 +161,81 @@ public void sendInvite(String username, long wait) {
...
     public void kick(int id) {
-        kick(group.getMember(id));
+        GroupMember member = group.getMember(id);
+        kick(member);
     }
 
     public void kick(GroupMember member) {
-        if (member != null && canKick()) kickUser(group.indexOf(member));
+        if (member == null) return;
+        if (pending != null || !canKick()) return;
```
Pablete1234, line 178:
```suggestion
        if (member == null || pending != null || !canKick()) return;
```

Pablete1234 (changes requested):
Just a couple changes, and obviously testing, but looks good

Pablete1234 (approved):

PR #234, opened 2023-01-23
Implement methods & entities from API

Punikekk:


src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -282,7 +282,11 @@ private void tickLogic(boolean running) {
                 setRunning(false);
             } catch (Throwable e) {
                 FeatureDefinition<Module> modDef = featureRegistry.getFeatureDefinition(newModule);
-                if (modDef != null) modDef.getIssues().addWarning("bot.issue.feature.failed_to_tick", IssueHandler.createDescription(e));
+                if (modDef != null &&
+                        modDef.addIssue(e, "bot.issue.feature.failed_to_tick", PluginIssue.Level.WARNING)) {
+                    setModule(new DummyModule());
+                    setRunning(false);
```
Pablete1234, line 288: Just because an exception has happened doesn't mean the whole bot should halt, that's a terrible way to handle the exception.
Simply log it, add it to the list of issues, and hope it's a one-time thing and doesn't keep happening. Back in the day we also made it so the bot tick rate decreased (only tick once per second, delay for 1000ms every time there's an error) in hopes that it fixes itself just due to time passing and in-game state changing, but we no longer do that

src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureDefinition.java
```diff
@@ -162,4 +163,26 @@ public boolean canLoad() {
     public void addIssue(@NotNull String message, @NotNull String description, @NotNull Level level) {
         issues.add(message, description, PluginIssue.Level.values()[level.ordinal()]);
     }
+
+    /**
+     * @return true if exception cause was {@link FeatureException}
```
Pablete1234, line 168: I think this makes no sense.

This addIssue simply should add it normally, like any other issue being added, with the particularity that it will indeed handle feature exception in a different way when creating the error message, which is an implementation detail

src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureDefinition.java
```diff
@@ -162,4 +163,26 @@ public boolean canLoad() {
...
+     */
+    public boolean addIssue(Throwable e, String message, PluginIssue.Level level) {
```
Pablete1234, line 170:
```suggestion
    public boolean addIssue(String message, Level level, Throwable e) {
```
It's convention for throwable to be the last param in logging libraries and such

src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureDefinition.java
```diff
@@ -162,4 +163,26 @@ public boolean canLoad() {
...
+                    ? featureException.getDescription() : IssueHandler.createDescription(featureException.getCause());
+            issues.add(featureException.getMessage(), description, PluginIssue.Level.ERROR);
+            return true;
+        }
+
+        issues.add(message, IssueHandler.createDescription(e), level);
+        return false;
+    }
+
+    private FeatureException findFeatureException(Throwable e) {
```
Pablete1234, line 183: this seems pretty out of place, maybe a `FeatureException.find(Throwable t)` static method in the api itself?

src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureDefinition.java
```diff
@@ -162,4 +163,26 @@ public boolean canLoad() {
...
     }
+
+    /**
+     * @return true if exception cause was {@link FeatureException}
+     */
+    public boolean addIssue(Throwable e, String message, PluginIssue.Level level) {
+        FeatureException featureException = findFeatureException(e);
+        if (featureException != null) {
+            String description = featureException.getCause() == null
+                    ? featureException.getDescription() : IssueHandler.createDescription(featureException.getCause());
```
Pablete1234, line 174:
```suggestion
            String description = IssueHandler.createDescription(featureException);
```
Make IssueHandler.createDescription itself be the one who can provide a different/simpler description if the passed exception instanceof FeatureException.

src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureDefinition.java
```diff
@@ -162,4 +163,26 @@ public boolean canLoad() {
...
+            issues.add(featureException.getMessage(), description, PluginIssue.Level.ERROR);
```
Pablete1234, line 175:
```suggestion
            issues.add(featureException.getMessage(), description, level);
```
You've been given a level, respect it. You may add a severity level FeatureException if you want to make it configurable there, but otherwise you should respect what's passed in the method.

src/main/java/com/github/manolo8/darkbot/extensions/features/FeatureRegistry.java
```diff
@@ -119,8 +118,8 @@ private <T> Optional<T> getFeature(String id) {
                 feature.setInstance(instance = featureLoader.loadFeature(feature));
                 return Optional.of(instance);
             } catch (Throwable e) {
-                feature.getIssues().addFailure("bot.issue.feature.failed_to_load", IssueHandler.createDescription(e));
-                e.printStackTrace();
+                if (!feature.addIssue(e, "bot.issue.feature.failed_to_load", PluginIssue.Level.ERROR))
+                    e.printStackTrace();
```
Pablete1234, line 122: you shouldn't make printstacktrace conditional, always make the console show the error, it HAS happened, don't hide it away

src/main/java/com/github/manolo8/darkbot/gui/drawables/DynamicEntitiesDrawer.java
```diff
@@ -62,6 +76,28 @@ public void onDraw(MapGraphics mg) {
         drawHeroTarget(mg);
     }
 
+    // todo create colors for those three methods
```
Pablete1234, line 79: yeah, this is needed

src/main/java/com/github/manolo8/darkbot/Main.java
```diff
@@ -293,8 +297,7 @@ private void tickLogic(boolean running) {
                     setRunning(false);
                 } catch (Throwable e) {
                     featureRegistry.getFeatureDefinition(behaviour)
-                            .getIssues()
-                            .addFailure("bot.issue.feature.failed_to_tick", IssueHandler.createDescription(e));
+                            .addIssue(e, "bot.issue.feature.failed_to_tick", PluginIssue.Level.ERROR);
```
Pablete1234, line 300: i generally dislike that we're using addIssue with a level everywhere, i think addFailure/addWarning/addInfo are better for readability.

Now that we have lombok it should be pretty easy to `@Delegate` the issues in FeatureDefinition, and make FeatureInfo interface extend Issues interface, on an API level.
What that essentially means is all "getIssues().*" methods can now be called directly on FeatureDefinition

Pablete1234 (changes requested):

src/main/resources/lang/strings_en.properties
```diff
@@ -589,6 +589,9 @@ colors.mines=Mines
 colors.allies=Allies
 colors.enemies=Enemies
 colors.npcs=NPCs
+colors.low_relays=LoW Relays
+colors.space_balls=SpaceBalls
+colors.static_entities=Static Entities
```
Pablete1234, line 594: I think a total of 0 users are going to understand this category
maybe just call it "other entities" and use it as a generic fallback color?

Pablete1234 (approved):

PR #235, opened 2023-01-24
group size reset

anjeshshrestha:
I think, need to rework resetting `members` array.

I detected error on server restart, the group is not sent back invite. 

I need to test, but we should not upload current group changes for next release until this is fixed


Pablete1234:
What's the state of this, have you been able to check that the behavior has all been corrected?

anjeshshrestha:
> What's the state of this, have you been able to check that the behavior has all been corrected?

I believe all has been corrected. Was not able to reprocude the bugs I faced

Pablete1234 (approved):

PR #243, opened 2023-02-05
Changing font

devskoo:
changed big and small font to consolas bold instead of consolas so it's easier to read and clearer


Pablete1234:
Hey, for visual changes could you append screenshots of before & after? thanks

devskoo:
> Hey, for visual changes could you append screenshots of before & after? thanks

https://discord.com/channels/523159099870019584/524630584920899586/1071795476548030484

Pablete1234:
Those are not screenshots, that's a link to discord. It'd be nice for the images to be linked here on the issue for future reference

devskoo:
Default: [https://imgur.com/Os7eoIh](url)
Bold: [https://imgur.com/wunCN7z](url)

Pablete1234:
Updated your comment for them to *actually* be embedded.
I can understand the personal preference of wanting them bold, but editing the default for the bot means you would need  to write code to manage config versioning to make it so older configs get the update, which is hard to justify for this.
The change is minimal (and i actually think it makes the font a bit harder on the eyes for the map name). If you prefer it bold feel free to edit it via the color scheme settings (after all that's what they're for!) but i'm going to have to decline this as a change for the defaults.

PR #244, opened 2023-02-05
return null on auction, dispatch, nova within expiry time

anjeshshrestha:
Backpage Auction, Dispatch and Nova will return `null` on function call`update(int expiry)` if it has already updateded within the expiry time


anjeshshrestha:
This would be breaking with older plugin version, Im not sure how to keep compatability with older version without depricating the function and creating new function with different name.

Pablete1234 (approved):

Pablete1234:
While this breaks binary compatibility, it is reduced enough in scope that i believe only your own plugin is affected, and i'm fine with accepting that breakage

PR #245, opened 2023-02-06
Turning AVOID_RADIATION off by default

devskoo:
This should be off by default as the description already says... ''use only if you are well aware of what you are doing''


Pablete1234:
The default is correct as on, you should avoid radiation by default.
What's wrong here is the description, which should be "Disable only if you are well aware of what you are doing"

PR #247, opened 2023-02-07
update lang, 3d settings on every reload

Punikekk:
Update language & 3D settings on every API reload, not only on relogin
Changed default encoding while running bot from IDE

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/utils/login/LoginData.java
+++ b/src/main/java/com/github/manolo8/darkbot/utils/login/LoginData.java
@@ -1,28 +1,15 @@
 package com.github.manolo8.darkbot.utils.login;
 
 import com.github.manolo8.darkbot.backpage.BackpageManager;
-import com.github.manolo8.darkbot.config.Config;
-import com.github.manolo8.darkbot.config.ConfigEntity;
-import com.github.manolo8.darkbot.utils.I18n;
 import com.google.gson.reflect.TypeToken;
+import eu.darkbot.api.managers.ConfigAPI;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/github/manolo8/darkbot/utils/login/LoginData.java
```diff
@@ -1,13 +1,32 @@
...
-    private static final Pattern USER_ID_PATTERN = Pattern.compile("userID=(\\d+)");
+    private static final Type PARAMS_TYPE = new TypeToken<Map<String, String>>() {}.getType();
+    private static final Map<String, BiFunction<Config, String, String>> REPLACEMENTS = new HashMap<>() {{
+        put("display2d", (config, defValue) -> config.BOT_SETTINGS.API_CONFIG.USE_3D ? "1" : "2");
+        put("autoStartEnabled", (config, defValue) -> "1");
+        put("lang", (config, defValue) -> Optional.of(I18n.getLocale().getLanguage())
+                .filter(l -> !l.isEmpty())
+                .filter(l -> config.BOT_SETTINGS.API_CONFIG.FORCE_GAME_LANGUAGE)
+                .orElse(defValue));
+    }};
```
Pablete1234, line 25: i feel like this has no place in login data

login data is supposed to be a simple object for storing the user's login credentials for use in the bot, this is just logic that should be handled elsewhere.

Also i think we should make better use of ConfigSetting, i think we could declare these in a cleaner way

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #248, opened 2023-02-10
update captcha redirect url

anjeshshrestha:
Previously, it would make 2 request in CaptchaHandler to get the redirected page url and the content. 
Using HttpURLConnection, the url will be automatically updated when the request is made, allowing it to make 1 less request. 

This could be further optimized, if darkbot http client stored the HttpURLConnection it created and reused it to retrieive the url

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/backpage/AuctionManager.java
+++ b/src/main/java/com/github/manolo8/darkbot/backpage/AuctionManager.java
@@ -33,7 +33,6 @@ public class AuctionManager {
     }
 
     /**
-     *
      * @param expiryTime only update if within
      * @return null if update wasn't required (non-expired), true if updated ok, false if update failed
      */
@@ -44,7 +43,8 @@ public class AuctionManager {
             String page = backpageManager.getHttp("indexInternal.es?action=internalAuction").getContent();
             if (this.captchaHandler.needsCaptchaSolve(page)) {
                 System.out.println("AuctionManager: Captcha Detected");
-                return captchaHandler.solveCaptcha();
+                captchaHandler.solveCaptcha();
+                return false;
             }
             lastAuctionUpdate = System.currentTimeMillis();
             return data.parse(page);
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/manolo8/darkbot/backpage/AuctionManager.java
```diff
@@ -44,7 +43,8 @@ public Boolean update(long expiryTime) {
             String page = backpageManager.getHttp("indexInternal.es?action=internalAuction").getContent();
             if (this.captchaHandler.needsCaptchaSolve(page)) {
                 System.out.println("AuctionManager: Captcha Detected");
-                return captchaHandler.solveCaptcha();
+                captchaHandler.solveCaptcha();
+                return false;
```
anjeshshrestha, line 47: This should have been returning false, because it did not update anything

src/main/java/com/github/manolo8/darkbot/backpage/DispatchManager.java
```diff
@@ -63,9 +62,9 @@ public Boolean update(long expiryTime) {
             String page = backpageManager.getHttp("indexInternal.es?action=internalDispatch").getContent();
             if (captchaHandler.needsCaptchaSolve(page)) {
                 System.out.println("DispatchManager: Captcha Detected");
-                return captchaHandler.solveCaptcha();
+                captchaHandler.solveCaptcha();
+                return false;
```
anjeshshrestha, line 66: same with this

src/main/java/com/github/manolo8/darkbot/utils/CaptchaHandler.java
```diff
@@ -32,16 +33,16 @@ public boolean needsCaptchaSolve(String page) {
...
-        System.out.println("CaptchaHandler: Solving Captcha for " + this.action);
-        HttpURLConnection conn = backpage.getHttp(base).getConnection();
-        conn.setInstanceFollowRedirects(false);
-        if (!conn.getHeaderField("Location").isEmpty()) {
-            HttpURLConnection connection = backpage.getHttp(conn.getHeaderField("Location")).getConnection();
-            captchaResponseFuture = CaptchaAPI.getInstance().solveCaptchaFuture(connection.getURL(), IOUtils.read(connection.getInputStream(), true))
+        HttpURLConnection connection = backpage.getHttp(base).getConnection();
+        String page = IOUtils.read(connection.getInputStream());
+
+        if (needsCaptchaSolve(page)) {
```
anjeshshrestha, line 43: this check can be considered redundant? but makes sure that the url we got has captcha in it.

Pablete1234 (approved):
Changes are fine as they are, you can always look into further improvements later

PR #249, opened 2023-02-12
Make use of FeatureException

Punikekk:
Implement use of FeatureException to show messages & decription of FeatureException next to features in plugins tab,
also reset working module if failure was critical to stop further ticking

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/extensions/plugins/IssueHandler.java
+++ b/src/main/java/com/github/manolo8/darkbot/extensions/plugins/IssueHandler.java
@@ -1,11 +1,9 @@
 package com.github.manolo8.darkbot.extensions.plugins;
 
 import com.github.manolo8.darkbot.core.utils.Lazy;
-import com.github.manolo8.darkbot.extensions.features.FeatureDefinition;
 import eu.darkbot.api.extensions.FeatureException;
 
 import java.util.Arrays;
-import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.function.Consumer;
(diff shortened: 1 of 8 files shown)


src/main/java/com/github/manolo8/darkbot/extensions/plugins/IssueHandler.java
```diff
@@ -25,6 +28,56 @@ public IssueHandler(IssueHandler parent) {
...
+    private void addIssue(FeatureException fe) {
+        String message = fe.getMessage();
+        String description = fe.getDescription() == null ? createDescription(fe.getCause()) : fe.getDescription();
+        PluginIssue.Level level = PluginIssue.Level.values()[fe.getLevel().ordinal()];
+
+        PluginIssue issue = new PluginIssue(message, description, level, fe);
+        addIssue(issue);
+    }
+
+    private void addIssue(PluginIssue pluginIssue) {
```
Pablete1234, line 66: a method called `add(PluginIssue pluginIssue)` already exists, which does basically this exact functionality (add it, and send an update to the listener if it's new)

You should update that method instead of creating a copy-cat one, as that method should also do increase and print logic

src/main/java/com/github/manolo8/darkbot/extensions/plugins/IssueHandler.java
```diff
@@ -25,6 +28,56 @@ public IssueHandler(IssueHandler parent) {
...
+                if (issue.equals(pluginIssue)) {
+                    issue.increaseAndPrint();
+                    return;
+                }
+            }
+        } else {
+            issues.add(pluginIssue);
+            pluginIssue.increaseAndPrint();
+            listener.send(this);
+        }
```
Pablete1234, line 78:
```suggestion
        if (issues.add(pluginIssue)) {
            pluginIssue.increaseAndPrint();
            listener.send(this);
        } else {
            for (PluginIssue issue : issues) {
                if (issue.equals(pluginIssue)) {
                    issue.increaseAndPrint();
                    break;
                }
            }
        }
```

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/modules/DummyExceptionModule.java
```diff
@@ -0,0 +1,20 @@
...
+
+    private final String feature;
+
+    public DummyExceptionModule(String feature) {
+        this.feature = feature;
+    }
+
+    @Override
+    public String getStatus() {
+        return "Module: " + feature + ", occurred with critical error!";
```
Pablete1234, line 13:
```suggestion
        return "Module '" + feature + "' failed with critical error!\nSee plugins tab for more info";
```
I'd add a 2nd line to remind them to just go look at that

Pablete1234 (commented):
Lgtm, just that small tweak

Pablete1234 (approved):

PR #250, opened 2023-02-14
add Event Npc Proxy

anjeshshrestha:
This adds the generic Event proxy window, can use the update to provide status similar to Labyrinth and GOP. also add api implementation

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/manager/FacadeManager.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/manager/FacadeManager.java
@@ -11,13 +11,14 @@ import com.github.manolo8.darkbot.core.objects.facades.EscortProxy;
 import com.github.manolo8.darkbot.core.objects.facades.EternalBlacklightProxy;
 import com.github.manolo8.darkbot.core.objects.facades.EternalGateProxy;
 import com.github.manolo8.darkbot.core.objects.facades.FrozenLabyrinthProxy;
+import com.github.manolo8.darkbot.core.objects.facades.GauntletPlutusProxy;
 import com.github.manolo8.darkbot.core.objects.facades.HighlightProxy;
 import com.github.manolo8.darkbot.core.objects.facades.LogMediator;
+import com.github.manolo8.darkbot.core.objects.facades.NpcEventProxy;
 import com.github.manolo8.darkbot.core.objects.facades.SettingsProxy;
 import com.github.manolo8.darkbot.core.objects.facades.SlotBarsProxy;
 import com.github.manolo8.darkbot.core.objects.facades.SpaceMapWindowProxy;
 import com.github.manolo8.darkbot.core.objects.facades.StatsProxy;
-import com.github.manolo8.darkbot.core.objects.facades.GauntletPlutusProxy;
 import com.github.manolo8.darkbot.core.objects.swf.PairArray;
 import eu.darkbot.api.PluginAPI;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/manolo8/darkbot/core/manager/FacadeManager.java
```diff
@@ -3,21 +3,7 @@
...
-import com.github.manolo8.darkbot.core.objects.facades.EternalGateProxy;
-import com.github.manolo8.darkbot.core.objects.facades.FrozenLabyrinthProxy;
-import com.github.manolo8.darkbot.core.objects.facades.HighlightProxy;
-import com.github.manolo8.darkbot.core.objects.facades.LogMediator;
-import com.github.manolo8.darkbot.core.objects.facades.SettingsProxy;
-import com.github.manolo8.darkbot.core.objects.facades.SlotBarsProxy;
-import com.github.manolo8.darkbot.core.objects.facades.SpaceMapWindowProxy;
-import com.github.manolo8.darkbot.core.objects.facades.StatsProxy;
-import com.github.manolo8.darkbot.core.objects.facades.GauntletPlutusProxy;
+import com.github.manolo8.darkbot.core.objects.facades.*;
```
Pablete1234, line 6: disable wildcard imports in your IDE

anjeshshrestha: ahhh not again 😭

src/main/java/com/github/manolo8/darkbot/core/objects/facades/NpcEventProxy.java
```diff
@@ -0,0 +1,47 @@
...
+
+import static com.github.manolo8.darkbot.Main.API;
+
+public class NpcEventProxy extends Updatable implements NpcEventAPI {
+    public double time;
+    public String eventID;
+    public String eventName;
+    public String eventDescriptionID;
+    public String eventDescription;
+    public String npc_leftStatus;
```
Pablete1234, line 15:
```suggestion
    public String npcLeftStatus;
```

src/main/java/com/github/manolo8/darkbot/core/objects/facades/NpcEventProxy.java
```diff
@@ -0,0 +1,47 @@
...
+        return time;
+    }
+
+    @Override
+    public NpcEventAPI.Status getStatus() {
+        if(this.eventActive){
+            return Status.ACTIVE;
+        }else{
+            return Status.INACTIVE;
+        } //need to detect when event is over;
```
Pablete1234, line 31:
```suggestion
        if (this.eventActive) {
            return Status.ACTIVE;
        } else {
            return Status.INACTIVE;
        } //need to detect when event is over;
```
please fix spacing, also consider if we can only detect active/inactive potentially making it a boolean (ideally tho, we'd like to be able to distinguish that)

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/objects/facades/NpcEventProxy.java
anjeshshrestha: This was to keep consistency with other proxies

Pablete1234 (approved):

Pablete1234:
Feel free to update the api version to 0.7.2 which should make the build pass

Pablete1234 (approved):

PR #251, opened 2023-02-14
New method for get the current hangar id

dm94:


src/main/java/com/github/manolo8/darkbot/backpage/HangarManager.java
```diff
@@ -72,6 +72,18 @@ public HangarResponse getCurrentHangar() {
         return currentHangar;
     }
 
+    /**
+     * @return The ID of the current hangar. A value of 0 indicates that the
+     *         currentHangar has not been initialized.
+     */
+    public int getCurrentHangarId() {
+        return getHangarList().getData().getRet().getHangars().stream()
```
Pablete1234, line 80: this can NPE in several places, hangar list can be null, data can be null, ret can be null, etc
While it's painful to write, you need to check for nullability every step of the way

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/backpage/HangarManager.java
dm94: Solved as I would like to have at the moment "?."

Pablete1234 (approved):

PR #253, opened 2023-02-16
Pair array replace

anjeshshrestha:
Add option to remove and replace proxy

Allowing for quick implementation of proxy on plugin side without needed full darkbot upgrade


Pablete1234:
I'm going to deny this one, plugins should not ever have to be responsible for implementing proxies, or interacting with memory in any way, we want plugins to migrate to using APIs, and this is going completely in the opposite direction to that

PR #255, opened 2023-02-18
Added upgraded lab values from RefinementGui

anjeshshrestha:
Added upgraded values of laser, shield, speed and  rocket
Big thanks to Demetrio@


src/main/java/com/github/manolo8/darkbot/core/objects/RefinementGui.java
```diff
@@ -71,12 +74,17 @@ public void update() {
...
 
                 if (name != null && !name.isEmpty()) {
-                    String processedName = name.replace("ore_", "");
-                    fuzzyName = processedName.substring(0, 1).toUpperCase(Locale.ROOT) + processedName.substring(1);
+                    if(name.contains("ore_")) {
+                        String processedName = name.replace("ore_", "");
+                        fuzzyName = processedName.substring(0, 1).toUpperCase(Locale.ROOT) + processedName.substring(1);
+                    }
+                    if(name.contains("lab_")){
+                        fuzzyName = API.readMemoryString(address, 0x108);
```
Pablete1234, line 86: the previous one gets the fuzzyName out of name, but this one doesn't
is there a reason for it? 
if so, could you leave it as a comment?
otherwise, would this work if the logic was the same for both?
Maybe document what kind of strings these are as part of a comment

Also watch the formatting `if (...) {`, with spaces between if and before {

dm94: You can also use `if {} else if {}` so that it will not do the second check if it has already entered the first `if`

Pablete1234 (approved):

PR #264, opened 2023-03-19
Fix of PetManager being stuck selecting npcs with same ICON infinitely

Punikekk:
Attempt to fix an issue where few aliens in PET locator tab have samme icon, so darkbot finds first npc as selected one - which is not always correct.

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
@@ -365,16 +365,18 @@ public class HeroManager extends Player implements Manager, HeroAPI {
     public SelectableItem.Laser getLaser() {
         return items.getItems(ItemCategory.LASERS).stream()
                 .filter(Item::isSelected)
-                .map(item -> SelectableItem.Laser.of(item.getId()))
-                .findFirst().orElse(null);
+                .findFirst()
+                .map(item -> item.getAs(SelectableItem.Laser.class))
+                .orElse(null);
     }
 
     @Override
     public SelectableItem.Rocket getRocket() {
         return items.getItems(ItemCategory.ROCKETS).stream()
                 .filter(Item::isSelected)
-                .map(item -> SelectableItem.Rocket.of(item.getId()))
-                .findFirst().orElse(null);
+                .findFirst()
+                .map(item -> item.getAs(SelectableItem.Rocket.class))
+                .orElse(null);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -365,16 +365,18 @@ public boolean launchRocket() {
...
     }
 
     @Override
     public SelectableItem.Rocket getRocket() {
         return items.getItems(ItemCategory.ROCKETS).stream()
                 .filter(Item::isSelected)
-                .map(item -> SelectableItem.Rocket.of(item.getId()))
-                .findFirst().orElse(null);
+                .findFirst()
+                .map(item -> item.getAs(SelectableItem.Rocket.class))
```
Pablete1234, line 378: same here

src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
```diff
@@ -28,18 +28,12 @@
...
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.EnumMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
Pablete1234, line 36: avoid wildcard imports, edit your IDE settings to not collapse imports into wildcards and revert these

src/main/java/com/github/manolo8/darkbot/core/objects/Gui.java
```diff
@@ -45,8 +45,10 @@ public void update() {
         width = (int) Math.round(size.x);
         height = (int) Math.round(size.y);
         // Set pos relative to window size
-        x = (int) Math.round((MapManager.clientWidth - size.x) * 0.01 * pos.x);
-        y = (int) Math.round((MapManager.clientHeight - size.y) * 0.01 * pos.y);
+        //x = (int) Math.round((MapManager.clientWidth - size.x) * 0.01 * pos.x);
+        //y = (int) Math.round((MapManager.clientHeight - size.y) * 0.01 * pos.y);
+        x = x();
+        y = y();
```
Pablete1234, line 51: i wonder, could these simply be super.x, super.y (given that they're made protected in sprite object) ?
i think x and y in gui should be marked deprecated (with the intent that they should be obtained via getX/getY method instead

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #267, opened 2023-03-30
Added option to use Instant Repair after revive, plus some minor changes

Punikekk:
Option to use Instant Repairs after revive while bot is running
update locktype of ship every tick

Created list of all entities, so every entity is updated/removed only once. Also getAll() method returns updatable list now


src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -97,6 +97,7 @@ public static class Safety {
             public @Option("config.general.safety.revive_location") @Dropdown ReviveLocation REVIVE = ReviveLocation.BASE;
             public @Option @Visibility(Level.INTERMEDIATE) @Number(min = 5, max = 60, step = 10) int WAIT_BEFORE_REVIVE = 5;
             public @Option @Visibility(Level.INTERMEDIATE) @Number(min = 3, max = 15 * 60, step = 10) int WAIT_AFTER_REVIVE = 90;
+            public @Option @Visibility(Level.INTERMEDIATE) @Number(min = 0, max = Integer.MAX_VALUE, step = 10) int INSTANT_REPAIR = 0;
```
Pablete1234, line 100: I'd use a `@Number.Disabled` here, makes sense that this is a check-box style thing (with, additionally, a count).
Also, IMO avoid Integer.MAX_VALUE, that's a number users will rarely want to set, do 9999 or whatever seems like a reasonable number, that should be the maximum. Keep in mind the size of the spinner is set according to the maximum number (so a max-value will make a huge jspinner for no reason)

src/main/java/com/github/manolo8/darkbot/core/manager/RepairManager.java
```diff
@@ -80,18 +81,45 @@ public boolean setBeforeReviveTime() {
         return System.currentTimeMillis() - beforeReviveTime < (main.config.GENERAL.SAFETY.WAIT_BEFORE_REVIVE * 1000L);
     }
 
+    private boolean instantRepaired = true;
```
Pablete1234, line 84: place all the state of the class at the top whenever possible

also, i'd consider swapping this one around, so it's "shouldInstaRepair = false", set to true once you die (if config allows), and re-set back to false after setting

right now, you're setting instantRepaired to true to "disable" instnatrepairing, even if you *didn't* really instant repair, so what the variable represents and its name doesn't quite align

src/main/java/com/github/manolo8/darkbot/core/utils/EntityList.java
```diff
@@ -82,11 +83,12 @@ public EntityList(Main main, EventBrokerAPI eventBroker) {
...
         this.entityRegistry.addToAll(entity -> {
             if (entity instanceof Obstacle) obstacles.add((Obstacle) entity);
-            eventBroker.sendEvent(new EntityCreateEvent(entity));
         });
-        this.entityRegistry.addDefault(e -> {
-            unknown.add(e);
-            eventBroker.sendEvent(new EntityCreateEvent(e));
+        this.entityRegistry.addDefault(unknown::add);
+
+        this.entityRegistry.onAnyEntityCreate(entity -> {
```
Pablete1234, line 89: is there any reason entityRegistry isn't created with all these directly? i'm not sure why we're doing these .setMain, .addToAll etc
can't we just `new EntityRegistry(main, this::onEntityCreate, unknown::add);`
and have an onEntityCreate method here that adds to all, adds to obstacle (if needed), and sends the event?

src/main/java/com/github/manolo8/darkbot/core/utils/EntityList.java
```diff
@@ -61,13 +62,13 @@ public class EntityList extends Updatable implements EntitiesAPI {
     public final List<Pet> pets                     = register(PET);
     public final List<BattleStation> battleStations = register(CBS_ASTEROID, CBS_MODULE, CBS_STATION, CBS_MODULE_CON, CBS_CONSTRUCTION);
     public final List<BasePoint> basePoints         = register(BASE_HANGAR, BASE_STATION, HEADQUARTER, QUEST_GIVER, BASE_TURRET, REPAIR_STATION, REFINERY);
-    public final List<Entity> unknown               = register();
+    public final List<Entity> unknown               = register(UNKNOWN);
```
Pablete1234, line 65: this change is pretty useless, no entity is ever created as UNKNOWN, as the predicate for it always returns false!!!

I'm actually not even sure why unknown is part of entity factory, as it can never actually occur (there's a fallback in EntityFactory#find which returns an in-line EntityBuilder `id, addr -> new unknown` but these will never actually be part of the UNKNOWN factory). Additionally there's extra logic in EntityFactory#sendEntity, where if it's not a factory (ie, it's not any of the ones in the list) it goes to the fallback listener, which is what does get stuff added here.

TLDR: should we just get rid of UNKNOWN entity factory type? i don't think it serves a purpose (the list already gets populated independently of that)

src/main/java/com/github/manolo8/darkbot/modules/utils/AttackAPIImpl.java
```diff
@@ -142,7 +144,11 @@ public double modifyRadius(double radius) {
...
             radius = Math.min(600, radius);
 
-        return radius + heroItems.getItem(SelectableItem.Ability.ZEPHYR_MMT, ItemFlag.AVAILABLE)
+        Optional<Item> ability = heroItems.getItem(SelectableItem.Ability.ZEPHYR_MMT, ItemFlag.AVAILABLE);
+        if (ability.isEmpty())
+            ability = heroItems.getItem(SelectableItem.Ability.HECATE_PLUS_STOCKPILE, ItemFlag.AVAILABLE);
+
+        return radius + ability
                 .map(Item::getQuantity)
                 .orElse(0d) * 5;
```
Pablete1234, line 153:
```suggestion
        return radius + heroItems.getItem(SelectableItem.Ability.ZEPHYR_MMT, ItemFlag.AVAILABLE)
            .or(() -> heroItems.getItem(SelectableItem.Ability.HECATE_PLUS_STOCKPILE, ItemFlag.AVAILABLE))
            .map(Item::getQuantity)
            .orElse(0d) * 5;
```

Java 11 has some nice optional extra operations (added in java 9), in this case, `Optional#or(Supplier<Optional>)` 
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#or(java.util.function.Supplier)

src/main/resources/lang/strings_en.properties
```diff
@@ -46,6 +46,8 @@ config.general.safety.wait_before_revive=Wait before revive (sec)
 config.general.safety.wait_before_revive.desc=Seconds to wait before reviving
 config.general.safety.wait_after_revive=Wait after revive (sec)
 config.general.safety.wait_after_revive.des=Seconds to wait after reviving, lets ship repair
+config.general.safety.instant_repair=Minimum amount of InstantRepairs to use
+config.general.safety.instant_repair.des=Minimum amount of InstantRepairs to use after revive - 0 to disable
```
Pablete1234, line 50:
```suggestion
config.general.safety.instant_repair=Use instant repairs when above
config.general.safety.instant_repair.desc=Use instant repairs to fully heal the ship after revive if there's at least these amount of instant repairs left
```
The comment of 0 to disable shouldn't be needed if it's turned into a checkbox+number style

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/utils/EntityList.java
```diff
@@ -161,22 +177,18 @@ public void clear() {
             obstacles.clear();
             fakeNpc.removed();
 
-            allEntities.forEach(entities -> {
-                entities.forEach(Entity::removed);
+            for (Entity entity : all) {
+                entity.removed();
+                eventBroker.sendEvent(new EntityRemoveEvent(entity));
```
Pablete1234, line 182: Although probably irrelevant, the behavior here is different than the previous, in the earlier EntityRemoveEvent the entity has already been removed from the list (by the iterator), while in this one it hasn't yet been removed.

Would that potentially cause any issue?

Pablete1234 (approved):

src/main/java/com/github/manolo8/darkbot/core/utils/EntityList.java
Punikekk: Entity was removed afterwards too in previous clear() method.
First was entity set to be removed -> send the event -> clear whole list -> repeat

PR #269, opened 2023-04-03
captcha update check

anjeshshrestha:
this update, checks the redirected url for `lostPilot` for captcha solve requirement,
We also set captcha solver to use `command` parameter from captchaSolver. This will be `checkRecaptcha` or `checkHcaptcha` dependent on the captcha page type

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/utils/CaptchaHandler.java
+++ b/src/main/java/com/github/manolo8/darkbot/utils/CaptchaHandler.java
@@ -52,9 +52,9 @@ public class CaptchaHandler {
                             http.setParam("desiredAction", action);
                             String response = http.getContent();
 
-                            if(response.toString().contains("OK")){
+                            if (response.toString().contains("OK")) {
                                 System.out.println("Captcha Solve Accepted");
-                            }else{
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/manolo8/darkbot/utils/CaptchaHandler.java
```diff
@@ -40,17 +41,22 @@ public void solveCaptcha() throws IOException {
...
                             r.forEach(http::setParam);
-                            http.closeInputStream();
+                            http.setParam("desiredAction", action);
+                            String response = http.getContent();
+
+                            if(response.toString().contains("OK")){
+                                System.out.println("Captcha Solve Accepted");
+                            }else{
+                                System.out.println("Captcha Solve Denied: " + response);
+                            }
```
Pablete1234, line 59:
```suggestion
                            if (response.toString().contains("OK")) {
                                System.out.println("Captcha Solve Accepted");
                            } else {
                                System.out.println("Captcha Solve Denied: " + response);
                            }
```

Pablete1234 (approved):

PR #272, opened 2023-05-29
dispatch & assembly mediator

anjeshshrestha:
Adding dispatch & assembly mediator and gui window for retrievers


src/main/java/com/github/manolo8/darkbot/core/objects/facades/DispatchRetrieverMediator.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public void update() {
+        long dispatchRetrieverData = API.readMemoryLong(address + 80) & ByteUtils.ATOM_MASK;
+        availableSlots = API.readMemoryInt(dispatchRetrieverData + 0x40);
+        totalSlots = API.readMemoryInt(dispatchRetrieverData + 0x44);
+
+        availableRetrieverArr.update(API.readMemoryLong(dispatchRetrieverData + 0x58) & ByteUtils.ATOM_MASK);
+        inProgressRetrieverArr.update(API.readMemoryLong(dispatchRetrieverData + 0x60) & ByteUtils.ATOM_MASK);
+        synchronized (UPDATE_LOCKER) {
```
Pablete1234, line 34: as long as these aren't being rendered, no need to do this

src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
```diff
@@ -68,6 +68,10 @@ public class GuiManager implements Manager, GameScreenAPI {
     public final SettingsGui settingsGui;
     public final ChatGui chat;
 
+    public final Gui dispatch;
+    public final Gui icon;
+    public final Gui icon_ok;
```
Pablete1234, line 73:
```suggestion
    public final Gui iconOk;
```

src/main/java/com/github/manolo8/darkbot/backpage/DispatchManager.java
```diff
@@ -50,30 +51,14 @@ public DispatchData getData() {
 
     @Deprecated
     public Boolean update(int expiryTime) {
-        return this.update((long) expiryTime);
+        return false;
```
Pablete1234, line 54: shouldn't this simply return "true" now as it will instantly have successfully updated?

src/main/java/com/github/manolo8/darkbot/core/manager/FacadeManager.java
```diff
@@ -95,6 +98,10 @@ public long getProxyAddressOf(String key) {
         return proxies.getPtr(key);
     }
 
+    public long getMediatorAddressOf(String key) {
```
Pablete1234, line 101: no, we're not exposing mediators outside of this class, and i'd rather actually change the register methods to be private, this WILL break i do not want 3rd parties using it. If there is a good use-case (and there's PLENTY) then register the mediator on the bot directly and expose it as an api

src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
```diff
@@ -124,6 +128,10 @@ public GuiManager(Main main, PluginAPI pluginAPI, RepairManager repairManager) {
         this.chat = register("chat", ChatGui.class);
         this.settingsGui = register("settings", SettingsGui.class);
 
+        this.dispatch = register("dispatch");
+        this.icon = register("popup_generic_icon");
+        this.iconOk = register("popup_generic_icon_ok");
```
Pablete1234, line 133: these should probs be using custom gui sub-classes, which provide some extra methods that allow some level of interaction with the gui

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/backpage/DispatchManager.java
anjeshshrestha: this is not longer part of backpage, is not deprecated and should be deleted

src/main/java/com/github/manolo8/darkbot/core/manager/FacadeManager.java
```diff
@@ -53,29 +56,36 @@ public class FacadeManager implements Manager, eu.darkbot.api.API.Singleton {
...
+        this.chrominEvent              = registerProxy("chrominEvent",              ChrominProxy.class);
+        this.astralGate                = registerProxy("rogue_lite",                AstralGateProxy.class);
+        this.highlight                 = registerProxy("HighlightProxy",            HighlightProxy.class);
+        this.spaceMapWindowProxy       = registerProxy("spacemap",                  SpaceMapWindowProxy.class);
+        this.plutus                    = registerProxy("plutus",                    GauntletPlutusProxy.class);
+        this.npcEventProxy             = registerProxy("npc_event",                 NpcEventProxy.class);
+        this.worldBossOverview         = registerProxy("worldBoss_overview",        WorldBossOverviewProxy.class);
+        this.group                     = registerProxy("GroupProxy",                Updatable.NoOp.class);
+        this.dispatchProxy             = registerProxy("dispatch",                  DispatchProxy.class);
+        this.dispatchRetrieverMediator = registerMediator("dispatch_retriever",     DispatchRetrieverMediator.class);
```
Pablete1234, line 87:
```suggestion
        this.dispatchMediator = registerMediator("dispatch_retriever",     DispatchMediator.class);
```
i'd simplify this name down

src/main/java/com/github/manolo8/darkbot/core/objects/facades/AssemblyMediator.java
```diff
@@ -0,0 +1,276 @@
...
+            if (address == 0) return;
+            isCraftable = API.readBoolean(address + 0x20);
+
+            long itemVoAddress = Main.API.readMemoryPtr(address + 0x58);
+            recipeId = API.readMemoryString(itemVoAddress, 0x48);
+
+            long rewardsArrAddress = Main.API.readMemoryPtr(address + 0x60);
+            rewardsArr.update(rewardsArrAddress);
+            rewards.clear();
+            rewardsArr.forEach(ptr -> rewards.add(API.readMemoryString(ptr, 0x48)));
```
Pablete1234, line 109: is it possible that this only needs to run on `update(long addr)`, that being, if the object is being read from a diff address? It would be nice for performance if stuff that does not ever update didn't need to update

src/main/java/com/github/manolo8/darkbot/core/objects/facades/AssemblyMediator.java
```diff
@@ -0,0 +1,276 @@
...
+        public String resourceId = "";
+        public double amountRequired, amountRequiredBackup = -1.0;
+
+        public void update() {
+            if (address <= 0) return;
+            long itemVoAddress = Main.API.readMemoryPtr(address + 0x28);
+            resourceId = API.readMemoryString(itemVoAddress, 0x48);
+            amountRequired = API.readDouble(address + 0x30);
+            //this also gives back same value, not sure which is correct
+            amountRequiredBackup = API.readDouble(address + 0x38);
```
Pablete1234, line 153: Is it possible that one is the unitary amount required, while the other is the total amount required? so if a recipe needs 10 material, but you're trying to craft 2 of the item, one will be 10, while the other is 20. If so, can we get that reflected?

src/main/java/com/github/manolo8/darkbot/core/objects/facades/AssemblyMediator.java
```diff
@@ -0,0 +1,276 @@
...
+
+    public static class RowFilter extends Auto implements AssemblyAPI.RowFilter {
+        public ItemFilter first;
+        public ItemFilter second;
+
+        public void update() {
+            if (address <= 0) return;
+            long firstAddress = Main.API.readMemoryPtr(address + 0x20);
+            if (firstAddress > 0) {
+                first = new ItemFilter();
```
Pablete1234, line 185: You should be able to just declare these off the bat, and call update(addr) without needing to re-create them

src/main/java/com/github/manolo8/darkbot/core/objects/facades/AssemblyMediator.java
anjeshshrestha: Checked to validate that ingame but doesnt reflect to be the case

src/main/java/com/github/manolo8/darkbot/core/objects/facades/AssemblyMediator.java
anjeshshrestha: If the address doesnt exist, no need to create them. also Im using this for null check to add them in filters list

Pablete1234: issue here is you're going to create new item filters *every single tick*, and nothing is setting them back to null if they were to stop being there either (ie: if they address went to 0, you'd just keep the previous value). So the question is, what is it really you're getting here? i think not much.

The solution is simply to always initialize both, and add them to the bigger list of address is != 0. Also should implement syncAndReport instead of relying on size*2 for updating the real filters

Pablete1234 (approved):

PR #274, opened 2023-06-10
move runClicks to Gui class

anjeshshrestha:
Move runClicks to gui class, used for dispatch and possible eternal, (future galaxy gate gui)


src/main/java/com/github/manolo8/darkbot/core/objects/Gui.java
```diff
@@ -216,6 +219,29 @@ public long getElementsList(int elementsListId) {
         return 0;
     }
 
+    /**
+     * Click Points on Gui, only if visible
+     * @param points lists of points to click
+     * @return if clicks were processed
+     */
+    public boolean runClicks(ClickPoint... points) {
```
Pablete1234, line 227: this used to be private, it should remain protected.

This is INTENTIONALLY not part of the API, and should not be used externally, if you want certain sequences of clicks keep them in the bot and have methods that perform them

Pablete1234 (changes requested):

PR #276, opened 2023-06-17
Move tick, ping, mem & cpu stats to title bar, plus few changes

Punikekk:
Fixed disable 3d render
Show names of portals & CBS on map
Avoid modules instead of CBS alone(workaround about invisible CBS in-game where modules still exists) Fixed dumpMainSWF method - need one change in API otherwise SafetyType will throw CastException to BattleStation#Hull
Update flatlaf version


src/main/java/com/github/manolo8/darkbot/core/entities/BattleStation.java
```diff
@@ -75,8 +76,7 @@ public boolean isRemoved() {
 
     @Override
     public boolean use() {
-        boolean allowEnemy = main.hero.invisible && main.config.GENERAL.ROAMING.ENEMY_CBS_INVISIBLE;
-        return hullId > 0 && hullId < 255 && (info.isEnemy() && !allowEnemy);
+        return false;
```
Pablete1234, line 79: if use is hard-coded to false, why even implement obstacle?

src/main/java/com/github/manolo8/darkbot/core/entities/BattleStation.java
```diff
@@ -16,10 +16,11 @@
 public class BattleStation
         extends Entity
         implements Obstacle, eu.darkbot.api.game.entities.BattleStation {
+    private static final int AVOID_RADIUS = 700;
```
Pablete1234, line 19: are we sure that 700 is enough not to be hit? i think the original 1200 around center is more than the 700 around the module.

Also one big consideration: we need to be certain that the center does not become un-avoided, or else, we may create invalid paths that get the bot stuck (ie: the bot trying to navigate from outside the battle station to inside the middle of the station).
Are we certain that 700 radius overlaps enough?

Ontop of that, i'm pretty certain that at some point in the past this was already a problem, and i believe we switched to a unique region to reduce issues, can we double-check git history and confirm what happened in the past won't be an issue now?

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
```diff
@@ -243,4 +256,63 @@ public double getTotalHonor() {
...
+                onChange.accept(MAX_TWO_PLACES_FORMAT.format(value));
+            last = value;
+
+            if (count == Integer.MAX_VALUE) {
+                count = 0;
+                average = 0;
+            }
+
+            average = average + (value - average) / (++count);
+        }
```
Pablete1234, line 298: I'm not convinced that keeping "permanent" stats is useful. The min/average/max over a long period doesn't seem too relevant, and i think for display purposes it makes more sense to user "rolling averages" or "weighted averages" where the most recent data is the one that matters the most.

That's essentially what tick-time used to do, where it would take the new sample as 1/10th while all past samples were 9/10th. This means that what the data will reflect most is the past few ticks/updates, as they "each" will have about 10%, while older data will become irrelevant over time.

I believe this class needs similar behavior, for all 3 min, avg and max.

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
```diff
@@ -243,4 +256,63 @@ public double getTotalHonor() {
...
+            min = Math.min(min, value);
+
+            if (last != value && onChange != null)
+                onChange.accept(MAX_TWO_PLACES_FORMAT.format(value));
+            last = value;
+
+            if (count == Integer.MAX_VALUE) {
+                count = 0;
+                average = 0;
+            }
```
Pablete1234, line 295: If we're thinking it's even possible that we'll reach int max value, just use a long instead

src/main/java/com/github/manolo8/darkbot/gui/drawables/ConstantEntitiesDrawer.java
```diff
@@ -34,22 +34,32 @@ public void onDraw(MapGraphics mg) {
...
 
             if (bs instanceof BattleStation.Module) {
                 mg.drawRectCentered(bs, 3.0, false);
             } else {
                 mg.drawOvalCentered(bs, 11.0, 9.0, true);
+
+                String name = bs instanceof BattleStation.Hull ? ("[" + bs.getEntityInfo().getClanTag() + "] ") : "";
+                name += bs.getEntityInfo().getUsername();
+
+                mg.drawString(bs, name, bs instanceof BattleStation.Hull ? -14 : -10, MapGraphics.StringAlign.MID);
```
Pablete1234, line 62: I think this one should go behind a display toggle, or be included in the "show usernames" one, as it can be sensitive info (ie: reveal what server you're playing on in the bot ui)

src/main/java/com/github/manolo8/darkbot/gui/drawables/DynamicEntitiesDrawer.java
```diff
@@ -6,19 +6,14 @@
...
-import eu.darkbot.api.game.entities.Box;
-import eu.darkbot.api.game.entities.Entity;
-import eu.darkbot.api.game.entities.Mine;
-import eu.darkbot.api.game.entities.Npc;
-import eu.darkbot.api.game.entities.Pet;
-import eu.darkbot.api.game.entities.Player;
-import eu.darkbot.api.game.entities.Relay;
-import eu.darkbot.api.game.entities.SpaceBall;
-import eu.darkbot.api.game.entities.StaticEntity;
+import eu.darkbot.api.game.entities.*;
```
Pablete1234, line 9: disable wildcard imports in your ide settings

src/main/java/com/github/manolo8/darkbot/gui/drawables/DynamicEntitiesDrawer.java
```diff
@@ -133,21 +135,44 @@ private void drawNpcs(MapGraphics mg) {
     private void drawPlayers(MapGraphics mg) {
         for (Player player : players) {
             boolean isEnemy = player.getEntityInfo().isEnemy();
-            mg.setColor(isEnemy ? "enemies" : "allies");
-            drawEntity(mg, player, 4, false);
 
+            int level = -1;
+            String color = isEnemy ? "enemies" : "allies";
```
Pablete1234, line 140: should probably cache the actual colors to avoid the N lookups when changing from one color to the other.

```
Color enemy = mg.getColor("enemy");
Color allies = mg.getColor("allies");
Color group = mg.getColor("group");

for (Player player : players) {
  // ...
  mg.setColor(isGroup ? group : isEnemy ? enemies : allies);
  // ...
}
```

src/main/java/com/github/manolo8/darkbot/gui/drawables/DynamicEntitiesDrawer.java
```diff
@@ -133,21 +135,44 @@ private void drawNpcs(MapGraphics mg) {
...
             if (mg.hasDisplayFlag(DisplayFlag.USERNAMES))
-                mg.drawString(player, player.getEntityInfo().getUsername(), -5, MapGraphics.StringAlign.MID);
+                mg.drawString(player, player.getEntityInfo().getUsername()
+                        + (level != -1 ? " [" + level + "]" : ""), -6, MapGraphics.StringAlign.MID);
+        }
+
+        if (group.hasGroup()) {
+            mg.setColor("group_member");
+            for (GroupMember member : group.getMembers()) {
+                if (member.getMapId() == hero.getMap().getId() && !isMemberVisible(member)) {
```
Pablete1234, line 159: I'm assuming this draws players who are not within render-distance, based on the coordinates gotten from group system.

If that's the case, i'd consider rendering these in a slightly different color/form. Maybe as a dotted outline, or a bit transparent, or a darker/lighter color, just something to signify there a bit of a "ghost" added-in by the bot. 
It would help know if they're within distance or not, as it would visually change in the ui whenever get got close/far from you.

src/main/java/com/github/manolo8/darkbot/gui/drawables/DynamicEntitiesDrawer.java
```diff
@@ -189,11 +214,20 @@ private void drawHeroTarget(MapGraphics mg) {
         }
     }
 
-    private void drawEntity(MapGraphics mg, Entity entity, double size, boolean fill) {
+    private boolean isMemberVisible(GroupMember member) {
+        for (Player player : players) {
```
Pablete1234, line 218: I don't particularily like the o(n^2) complexity of this (for each group member you're iterating all the visible players). Consider instead creating a set with the ids of all group members, and when rendering players call remove on the set. After that when rendering the extra group members you just gotta check that they're in the set to know they are not-visible.


Roughly:
```
Set<Long> nonVisibleMembers = group.members().map(toIds).collect(toSet());
for (Player player : players) {
  // render the player
  nonVisibleMembers.remove(player.id());
}

for (GroupMember member : group.members()) {
  if (!nonVisibleMembers.contains(member)) continue;
  // render ghostly-figure of the member
}
```

src/main/java/com/github/manolo8/darkbot/gui/titlebar/MainTitleBar.java
```diff
@@ -31,21 +41,100 @@ public MainTitleBar(Main main, MainGui frame) {
...
 
-    }
+        private JComponent createLeftPanel(StatsManager statsManager) {
+            JPanel panel = new JPanel(new GridLayout(2, 1));
+
+            JLabel tick = createLabel("tick_time", false, null,
+                    () -> "Tick time\n\n" + statsManager.getTickStats());
+            JLabel ping = createLabel("ping", false,
+                    color -> {
+                        float hue = 1 - Math.min(500, Math.max(1, main.statsManager.getPing())) / 500f;
```
Pablete1234, line 73: i think the scale here may not be all that great; 500 ping is real bad. given the 500 is purely for color purposes, i think past 300ms getting the worst possible red is fine, with the green-to-red going from 1ms to 300ms.

src/main/java/com/github/manolo8/darkbot/gui/titlebar/MainTitleBar.java
```diff
@@ -31,21 +41,100 @@ public MainTitleBar(Main main, MainGui frame) {
...
+
+            if (alignRight) {
+                label.setHorizontalAlignment(SwingConstants.RIGHT);
+                label.setHorizontalTextPosition(SwingConstants.LEFT);
+            } else label.setHorizontalAlignment(SwingConstants.LEFT);
+
+            Dimension preferredSize = label.getPreferredSize();
+            label.setPreferredSize(new Dimension(40, preferredSize.height));
+
+            label.setText("0");
```
Pablete1234, line 126:
```suggestion
            label.setText("...");
```
i'd rather start out at ... than a plain 0

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/entities/BattleStation.java
Punikekk: to do not break plugins, also in API battlestation implement Obstacle

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -80,7 +83,22 @@ public void mousePressed(MouseEvent e) {
                     main.setRunning(!main.isRunning());
                     repaint();
 
-                } else main.hero.drive.move(mapGraphics.toGameLocation(e));
+                } else {
+                    Locatable loc = mapGraphics.toGameLocation(e);
+                    if (e.getClickCount() == 2) {
```
Pablete1234, line 88: how about right click to open a pop-up dropdown that has an option "jump to X" map? that way there's no possible confusion about it

src/main/java/com/github/manolo8/darkbot/gui/drawables/DynamicEntitiesDrawer.java
```diff
@@ -131,23 +146,61 @@ private void drawNpcs(MapGraphics mg) {
...
-            mg.setColor(isEnemy ? "enemies" : "allies");
-            drawEntity(mg, player, 4, false);
 
+            int level = -1;
+            Color color = isEnemy ? enemy : ally;
+            if (group.hasGroup()) {
+                GroupMember member = group.getMember(player);
+                if (member != null) {
+                    color = groupMember;
+                    level = member.getLevel();
```
Pablete1234, line 167: the level sounds pretty irrelevant info for the bot gui, i don't see the utility in it

src/main/java/com/github/manolo8/darkbot/gui/drawables/DynamicEntitiesDrawer.java
```diff
@@ -131,23 +146,61 @@ private void drawNpcs(MapGraphics mg) {
...
             if (mg.hasDisplayFlag(DisplayFlag.USERNAMES))
-                mg.drawString(player, player.getEntityInfo().getUsername(), -5, MapGraphics.StringAlign.MID);
+                mg.drawString(player, player.getEntityInfo().getUsername()
+                        + (level != -1 ? " [" + level + "]" : ""), -6, MapGraphics.StringAlign.MID);
+        }
+
+        if (group.hasGroup()) {
+            mg.setColor(groupMember.darker());
+
+            Map<TextAttribute, Integer> attrs = Map.of(TextAttribute.UNDERLINE, TextAttribute.UNDERLINE_LOW_DASHED);
```
Pablete1234, line 181: nah, underline on the name itself ain't it, that makes them seem more important

i'd just keep the .darker on the color, and maybe make it so player entities are filled-in but these are not-filled?

src/main/java/com/github/manolo8/darkbot/gui/MapDrawer.java
```diff
@@ -314,6 +354,7 @@ public int getHeightMiddle() {
 
         @Override
         public boolean hasDisplayFlag(eu.darkbot.api.config.types.DisplayFlag displayFlag) {
+            if (true) return false;
```
Pablete1234, line 357: i think you have some leftovers

src/main/java/com/github/manolo8/darkbot/gui/titlebar/DiagnosticBar.java
```diff
@@ -0,0 +1,106 @@
...
+                    color -> UIUtils.getTrafficLight(Main.API.getCpuUsage(), 100), statsManager.getCpuStats());
+            JLabel ram = createLabel("ram", "Ram usage", true,
+                    color -> UIUtils.getTrafficLight(Main.API.getMemoryUsage(), 2500), statsManager.getMemoryStats());
+            add(cpu, "cell 1 0, gapleft 5px");
+            add(ram, "cell 1 1, gapleft 5px");
+        }
+
+        for (Component component : getComponents()) {
+            component.addMouseListener(new RedirectMouseAdapter());
+            component.addMouseMotionListener(new RedirectMouseAdapter());
```
Pablete1234, line 46: try to use a layered ui for the parent to capture all events occurring inside its bounds, instead of attaching redirecting listeners to each child

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/titlebar/RefreshButton.java
```diff
@@ -61,30 +67,24 @@ public void paintIcon(Component c, Graphics g, int x, int y) {
...
-            Color foreground = c.getForeground();
-            return FlatButtonUI.buttonStateColor(c, foreground, null, null,
-                    darker(foreground, 0.75), darker(foreground, 0.6));
-        }
-
         @Override
-        public int getIconWidth() {
-            return 16;
+        public float getValue(Component c) {
+            return ((AbstractButton)c).getModel().isRollover() ? 1 : 0;
```
Pablete1234, line 76: instead of animating when mouse goes over, i'd animate when it's clicked

src/main/java/com/github/manolo8/darkbot/gui/utils/FloatingDialog.java
```diff
@@ -1,22 +1,21 @@
...
+    }
+
+    public static JPopupMenu show(JComponent component, JComponent parent, int x, int y) {
         JPopupMenu popup = new JPopupMenu();
         popup.setBorder(BorderFactory.createEmptyBorder());
-        popup.add(panel);
-        popup.show(parent, 0, parent.getHeight());
+        popup.add(component);
+        popup.show(parent, x, y);
+        return popup;
```
Pablete1234, line 19: i'd revert all changes in this, keep the FloatingDialog class for... well.. arbitrary JPanels that you want to show as floating dialogs, which was the main mission. The whole showing a jmenu is just already supported behavior that's available via the jmenu itself, which is what ended up happening anyways for the use-case you tried to migrate to use this

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #280, opened 2023-07-02
Added an option for the "DisconnectModule" module to be able to close the bot as well.

dm94:


src/main/java/com/github/manolo8/darkbot/modules/DisconnectModule.java
```diff
@@ -83,6 +92,8 @@ public void tickStopped() {
             else {
                 pauseUntil = System.currentTimeMillis() + pauseTime;
                 main.setRunning(false);
+
+                if (closeBot) System.exit(0);
```
Pablete1234, line 96: I see a few main lacking things here:
 - Whenever we call system.exit, we *always* log explaining why the bot closed, makes it easier to know  (just posting bot exited and the reason would do)
  - It'd be nice if the status in the module also clarified that the bot won't just pause, but completely exit when the countdown gets to 0
  - It'd be nice if, instead of instantly closing when reaching 0, it opened a pop-up telling the user the bot will automatically close in 10s (counting down) and closing only after that. That way the user can cancel the shutdown instead of having no control over it shutting down just because they missconfigured something and now the bot is instantly shutting down without them knowing, possibly on loop every time they turn the bot on

dm94: Regarding the status, I think that the one who should make the warning is the one who uses the option by means of the variable "reason".
The option to add the pop-up can be interesting especially if the user has activated the "Confirm exit" option in the bot's configuration.

dm94:
Changes made but not tested

dm94:
Tested, works as it should

src/main/java/com/github/manolo8/darkbot/modules/DisconnectModule.java
```diff
@@ -22,15 +28,30 @@ public class DisconnectModule extends TemporalModule {
...
+     * @param pauseTime null for infinite pause, otherwise pause for that amount of
+     *                  MS.
+     */
+    public DisconnectModule(Long pauseTime, String reason, boolean closeBot) {
+        this(pauseTime, reason);
+        this.closeBot = closeBot;
+
+        if (closeBot) {
+            showClosePopup();
+        }
```
Pablete1234, line 52: This is showing the pop-up when the module is created, not when the module is set/activated to run.
It should be part of the tick function to run this

Pablete1234 (changes requested):
I think the approach here is going to give some trouble, as the pop-up is not actually timed, you may not have any time at all to click it (imagine your ship is already by the portal, something triggers it, instant shutdown).

Consider changing the behavior so that when it reaches stopped state, instead of running system.exit(0), then, it shows the pop-up, and if no action is taken, then 15s after, then the pop-up logic itself is what closes it and calls the stop

src/main/java/com/github/manolo8/darkbot/modules/DisconnectModule.java
dm94: I've put it here so that the shutdown timer itself will time the shutdown to prevent it from closing.

Pablete1234: I don't think you understand the issue:

```java
class MyFeature implements Behavior {

    private final DisconnectModule shutdown = new DisconnectModule(0L, "Shutting down bot", true);

    public void tick() {
        if (someConditionMet) bot.setModule(shutdown);
    }
}

```

This is a valid to create/initialize/setup the disconnect module, however this would show the pop-up instantly on module creation, instead of when "someConditionMet" which is when it actually starts attempting the disconnect.

Even if that itself is fixed, it would still shutdown in 0s if it was already near the portal giving no time to the user to react

dm94: I didn't take that into account, I have just fixed it.

dm94:
> I think the approach here is going to give some trouble, as the pop-up is not actually timed, you may not have any time at all to click it (imagine your ship is already by the portal, something triggers it, instant shutdown).
> 
> Consider changing the behavior so that when it reaches stopped state, instead of running system.exit(0), then, it shows the pop-up, and if no action is taken, then 15s after, then the pop-up logic itself is what closes it and calls the stop

Minimum will be 5 seconds and in most cases 20 seconds.

PR #286, opened 2023-07-22
added combo box to inspector, update child addresses on the fly

Punikekk:


src/main/java/com/github/manolo8/darkbot/gui/utils/inspector/ObjectInspectorUI.java
```diff
@@ -1,54 +1,106 @@
...
             addActionListener(ae -> {
-                try {
-                    long object = Long.parseLong(getText(), 16);
+                Long object;
+                Object selectedItem = getSelectedItem();
+                if (selectedItem instanceof AddressEntry) {
+                    object = ((AddressEntry) selectedItem).address.get();
+                } else {
+                    object = tryParse((String) selectedItem, true);
+                    if (object == null) tryParse((String) selectedItem, false);
```
Pablete1234, line 53: this won't work, if a number looks decimal, eg: "1234" it will successfully parse as hex, but the value will be something else.
If anything you'd have to try to parse as decimal, and it'd fail under situations like "a123", then fall-back to hex.

I'd just make it so that 0x[whatever] is hex, check with startsWith or smthg, and based on that do one or the other

src/main/java/com/github/manolo8/darkbot/gui/utils/inspector/ObjectTreeNode.java
```diff
@@ -36,49 +37,50 @@ public boolean isLeaf() {
...
     public void update(DefaultTreeModel model) {
+        long address = this.address.get();
         if (address == 0) {
             return;
         }
 
         ObjectInspector.Slot slot = (ObjectInspector.Slot) (this.getUserObject());
 
-        long oldValue = value;
+        boolean changed = value != address;
```
Pablete1234, line 51: value and address will not be equal for all nodes where addressIsValue is false. the point of the check was to avoid re-computing string values and updating the tree if the value is the same

Pablete1234 (approved):

PR #289, opened 2023-07-22
update dispatch and assembly with more info

anjeshshrestha:
Updates dispatch to use correct offset for slotId and tier, add retriever cost information
Update assembly to add inProgress and isCompleted info
Update gui to use correct updated offsets for width, height, visible and minimizable and change all offsets to hex


src/main/java/com/github/manolo8/darkbot/core/objects/facades/DispatchMediator.java
```diff
@@ -45,18 +45,39 @@ public static class Retriever extends Auto implements DispatchAPI.Retriever {
...
+    @ToString
+    public static class Cost extends Auto  implements DispatchAPI.Cost{
+        private String lootId = "";
+        private int amount = -1;
+
+        @Override
+        public void update() {
+            if (address <= 0) return;
+            this.amount = API.readMemoryInt(address + 0x20); // 1
+            this.lootId = API.readMemoryString(address, 0x28); // 1
```
Pablete1234, line 80: loot id example value is... 1?
i assume it should probably be something like 200000 and "credits" or something

src/main/java/com/github/manolo8/darkbot/core/objects/facades/AssemblyMediator.java
```diff
@@ -101,6 +104,7 @@ public void update() {
         }
     }
 
+    @Getter
```
Pablete1234, line 107:
```suggestion
```
This is not part of the api, it should be hidden away and never available or used.

Please stop relying on internals from plugins, separate api and implementation

src/main/java/com/github/manolo8/darkbot/core/objects/IconGui.java
```diff
@@ -1,15 +1,26 @@
...
 
 public class IconGui extends Gui implements API.Singleton {
+    @Override
+    public void update() {
+        if (address == 0) return;
+        super.update();
+        width = (int) Main.API.readMemoryDouble(address + 0x1F8);
+        height = (int) Main.API.readMemoryDouble(address + 0x200);
+        visible = Main.API.readMemoryLong(address + 0x180) != 0; // isVisible, sprite object is set
+    }
```
Pablete1234, line 14: the same update method exists in two guis, what's behind this decision?

Do all guis share this? is this a special type of gui only that does this?

Maybe the play here is a "PopupGui" class that extends gui and overrides update method like this, then icon gui & icon ok gui can extend PopupGui.

src/main/java/com/github/manolo8/darkbot/core/objects/Gui.java
```diff
@@ -16,7 +16,7 @@ public class Gui extends SpriteObject implements API, eu.darkbot.api.game.other.
     protected final Point minimized = new Point();
 
     public long addressInfo, featureWindowDefinition;
-    public boolean visible;
+    public boolean visible, isClosing;
```
Pablete1234, line 19: I'm against adding more public fields unless they're needed (as it makes it impossible to get rid of them later on without causing plugin exceptions), if you must, add it as private and add a getter.

We have isTweening which is for when a gui is transitioning (ie: showing the open or close animation), is this different? We have show(true) and show(false) methods which will return false until transition has finished, based on timers, does this superceed that?

src/main/java/com/github/manolo8/darkbot/core/objects/facades/AssemblyMediator.java
```diff
@@ -121,13 +125,15 @@ public static class ItemFilter extends Reporting implements AssemblyAPI.Filter {
         private String filterName = "";
         private boolean isChecked;
         private int row, col;
+        private double x, y;
```
Pablete1234, line 128: it doesn't seem like anything in darkbot is using these, are you using them from the plugin? If so, please stop doing so and add methods in darkbot for the purpose.

This class should not be public, AssemblyAPI.Filter is the public interface and should be the only visible face of it.

Pablete1234 (changes requested):
Ontop of the comments, i've deployed darkbot api v0.7.7, update to it so that you can get a passing build

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DispatchMediator.java
anjeshshrestha: ahh this should be some copy paste forgot to remove comments

Pablete1234 (approved):

PR #292, opened 2023-07-26
Update AstralGateProxy

dm94:
Added methods for accessing the following data: rewardItems, inventoryItems, allowedToEquip, highScore, currentRift, currentScore and cpuCount


src/main/java/com/github/manolo8/darkbot/core/objects/facades/AstralGateProxy.java
```diff
@@ -89,7 +90,11 @@ public void update() {
             this.upgradeLevel = API.readInt(itemData + 0x2C);
 
             this.lootId = API.readString(itemData, 0x48);
-            this.status = API.readString(itemData, 0x60);
+
+            itemStatsArr.update(API.readMemoryLong(address + 0x38));
+            synchronized (UPDATE_LOCKER) {
```
Pablete1234, line 95: is this going to be used in the ui thread? if not you probably do not need synchronization.

Maybe we should synchronize at the facade manager level to avoid having it all around

dm94: It is true that the stats are not necessary because they will not be updated with UI changes.

src/main/java/com/github/manolo8/darkbot/core/objects/facades/AstralGateProxy.java
```diff
@@ -1,21 +1,137 @@
...
+        this.cpuCount = API.readMemoryInt(API.readMemoryLong(data + 96) + 40);
+        this.canEquip = API.readBoolean(API.readMemoryLong(data + 0x0B0) + 0x20);
+
+        rewardItemsArr.update(API.readMemoryLong(data + 0x88));
+        synchronized (UPDATE_LOCKER) {
+            rewardItemsArr.sync(rewardItems, AstralItem::new);
+        }
+
+        inventoryItemsArr.update(API.readMemoryLong(data + 0x0A0));
+        synchronized (UPDATE_LOCKER) {
```
Pablete1234, line 42: those update locks can probably go away too

Pablete1234 (commented):
darkbot api 0.7.7 is available, implement astral api in it and this should be good to merge

src/main/java/com/github/manolo8/darkbot/core/objects/facades/AstralGateProxy.java
Pablete1234: the update locker exists for the **bot** gui not to re-paint while stuff is being updated, NOT the in-game gui.

If the data is not being painted by darkbot's MapDrawer, this serves no purpose

dm94:
I have had to implement isAccessible to fix the build

Pablete1234 (approved):

PR #293, opened 2023-07-27
Quest proxy

dm94:


Pablete1234:
Pending on finding all the values to make a usable API, see https://github.com/darkbot-reloaded/DarkBotAPI/pull/81

dm94:
Waiting for https://github.com/darkbot-reloaded/DarkBotAPI/pull/81

src/main/java/com/github/manolo8/darkbot/core/manager/FacadeManager.java
```diff
@@ -58,30 +59,32 @@ public class FacadeManager implements Manager, eu.darkbot.api.API.Singleton {
...
+        this.chrominEvent = registerProxy("chrominEvent", ChrominProxy.class);
+        this.astralGate = registerProxy("rogue_lite", AstralGateProxy.class);
+        this.highlight = registerProxy("HighlightProxy", HighlightProxy.class);
+        this.spaceMapWindowProxy = registerProxy("spacemap", SpaceMapWindowProxy.class);
+        this.plutus = registerProxy("plutus", GauntletPlutusProxy.class);
+        this.npcEventProxy = registerProxy("npc_event", NpcEventProxy.class);
+        this.worldBossOverview = registerProxy("worldBoss_overview", WorldBossOverviewProxy.class);
+        this.group = registerProxy("GroupProxy", Updatable.NoOp.class);
+        this.quests = registerProxy("QuestProxy", QuestProxy.class);
+        this.groupMediator = registerMediator("GroupSystemMediator", Updatable.NoOp.class);
```
Pablete1234, line 87: avoid re-formats like these, complying to a reformatting tool doesn't always benefit code readability

src/main/java/com/github/manolo8/darkbot/core/manager/FacadeManager.java
```diff
@@ -123,13 +126,15 @@ public void install(BotInstaller botInstaller) {
 
     public void tick() {
         // Currently commands are not used by the bot and they represent
-        // a decently big cpu chunk in ticking. Leaving them out reduces tick time significantly.
-        //commands.update();
+        // a decently big cpu chunk in ticking. Leaving them out reduces tick time
+        // significantly.
+        // commands.update();
```
Pablete1234, line 131: if you're going to reformat it try to do so by moving the a upwards see if it can fit in 2 lines instead of 3 with a single word leftover

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
```diff
@@ -0,0 +1,300 @@
...
+    private boolean visibleQuestGiver;
+    private int tabSelected;
+
+    @Override
+    public void update() {
+        if (address == 0) {
+            return;
+        }
+
+        long questClass = API.readMemoryLong(address + 0x98) & ByteUtils.ATOM_MASK;
```
Pablete1234, line 36:
```suggestion
        long questClass = API.readMemoryPtr(address + 0x98);
```
try to use readMemoryPtr to read pointers instead of adding the & mask at the end

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
```diff
@@ -0,0 +1,300 @@
...
+        } else {
+            if (this.currentQuest == null) {
+                this.currentQuest = new Quest();
+            }
+
+            this.currentQuest.update(currentQuestAddr);
+        }
+
+        questItemsArr.update(API.readMemoryPtr(0x58));
+        questItemsArr.sync(questItems, QuestListItem::new);
```
Pablete1234, line 54: i am concerned about the performance implications of this, refreshing a whole long list of quests and all their details every tick can lead to 1000s of mem reads, when most of the times they're not wanted or cared about

If we have it so this only updates in ticks where the getter(s) is/are invoked it would greatly reduce unnecessary overhead

src/main/java/com/github/manolo8/darkbot/core/manager/FacadeManager.java
dm94: Fixed

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
dm94: Fixed

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
dm94: Fixed

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
```diff
@@ -0,0 +1,336 @@
...
+
+        if (currentQuestAddr == 0) {
+            this.currentQuest = null;
+        } else {
+            if (this.currentQuest == null) {
+                this.currentQuest = new Quest();
+            }
+
+            this.currentQuest.update(currentQuestAddr);
+        }
```
Punikekk, line 48: Instead of checking nulls & address here, you can make `currentQuest, questInfoGiverSelected, questGiverSelected` final, and only in the getters check the address `return currentQuest.address == 0 ? null : currentQuest`.
I see that you already return on address 0 in update() methods of those classess

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
```diff
@@ -0,0 +1,336 @@
...
+
+        this.visibleQuestGiver = API.readMemoryBoolean(address, 0x40);
+
+        if (!visibleQuestGiver) {
+            this.questInfoGiverSelected = null;
+            this.questGiverSelected = null;
+            return;
+        }
+
+        this.tabSelected = API.readMemoryInt(address, 0X50);
```
Punikekk, line 58: 0X50 -> 0x50

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
```diff
@@ -0,0 +1,336 @@
...
+            if (address == 0) {
+                return;
+            }
+
+            this.id = API.readMemoryInt(address + 0x20);
+            this.active = API.readMemoryBoolean(address + 0x24);
+            this.title = API.readMemoryString(address, 0x68);
+            this.description = API.readMemoryString(address, 0x70);
+
+            long requirementAddr = API.readMemoryLong(address + 0x38) & ByteUtils.ATOM_MASK;
```
Punikekk, line 199: Use readMemoryPtr or use `readAtom(0x38` which Updatable class have. Also have many others like `readInt, readString` etc. which uses address variable internally

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
```diff
@@ -0,0 +1,336 @@
...
+                return;
+            }
+
+            this.enabled = API.readMemoryBoolean(address + 0x24);
+            this.completed = API.readMemoryBoolean(address, 0x34);
+            this.description = API.readMemoryString(address, 0x60);
+            this.goalReached = API.readMemoryDouble(address + 0x78);
+            this.goal = API.readMemoryDouble(address + 0x80);
+
+            long definitionAddr = API.readMemoryLong(address + 0x58) & ByteUtils.ATOM_MASK;
```
Punikekk, line 294: Same here

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
dm94: done

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
dm94: done

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
dm94: done

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
dm94: done

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
```diff
@@ -0,0 +1,323 @@
...
+    public void update() {
+        if (address == 0) {
+            return;
+        }
+
+        long questClass = API.readMemoryPtr(address + 0x98);
+
+        long currentQuestAddr = API.readMemoryPtr(questClass + 0x28);
+        if (this.currentQuest == null) {
+            this.currentQuest = new Quest();
```
Pablete1234, line 39: we generally avoid having null objects and needing to initialize them, consider just having the `currentQuest = new Quest()` be the default value, and in getCurrentQuest just do `quest.address == 0 ? null : quest;`, same for questInfoGiverSelected

src/main/java/com/github/manolo8/darkbot/core/objects/facades/QuestProxy.java
```diff
@@ -0,0 +1,323 @@
...
+        return questInfoGiverSelected == null || questInfoGiverSelected.address == 0 ? null : questInfoGiverSelected;
+    }
+
+    @Override
+    public @Nullable List<? extends QuestAPI.QuestListItem> getCurrestQuests() {
+        questItemsArr.update(API.readMemoryPtr(address + 0x58));
+        questItemsArr.sync(questItems, QuestListItem::new);
+
+        return questItems;
+    }
```
Pablete1234, line 87: The implementation of this method could cause a naive plugin to cause tons of re-reading in memory, eg:
```java
for (int i = 0; i < getCurrentQuests().size(); i++) {
  if (getCurrentQuests().get(i).something()) { ... }
}
```
I know it's annoying, but this needs to happen only at most once per tick, you can wrap it arround an:
```java 
if (!questsUpdated) {
  questsUpdated = true;
  questItemsArr.update(API.readMemoryPtr(address + 0x58));
  questItemsArr.sync(questItems, QuestListItem::new);
}
return questItemsArr
```
and just set questsUpdated back to false on the tick method

Pablete1234 (approved):

PR #303, opened 2023-08-13
New StatType condition

dm94:


src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+
+        return null;
+    }
+
+    public enum StatData {
+        CURRENT,
+        EARNED,
+        SPENT;
```
Pablete1234, line 52: Missing a "difference" or something like that, which is earned-spent, its what makes the most sense for graphs and overall "how much did the bot make"

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
```diff
@@ -0,0 +1,94 @@
...
+
+        @Override
+        public String toString() {
+            return name().toLowerCase(Locale.ROOT).replace("_", "-");
+        }
+
+        public static StatData of(String sd) {
+            for (StatData sD : StatData.values()) {
```
Pablete1234, line 60:
```suggestion
            for (StatData sd : StatData.values()) {
```

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
dm94: Wouldn't that be EARNED?

Pablete1234: Currently earned is only positive differences, so if you earn 100.000 uri, but spend 200.000 uri to gain it, stuff like uri/h will show positive, but the graph of overall uri should be going down (hence, difference, or earned - spent)

Pablete1234 (approved):

PR #307, opened 2023-08-27
New TargetType condition

dm94:


src/main/java/com/github/manolo8/darkbot/config/actions/conditions/TargetTypeCondition.java
```diff
@@ -0,0 +1,89 @@
...
+import com.github.manolo8.darkbot.config.actions.Parser;
+import com.github.manolo8.darkbot.config.actions.SyntaxException;
+import com.github.manolo8.darkbot.config.actions.ValueData;
+import com.github.manolo8.darkbot.config.actions.parser.ParseUtil;
+
+import eu.darkbot.api.game.entities.Npc;
+import eu.darkbot.api.game.other.Lockable;
+import eu.darkbot.api.game.other.EntityInfo.Diplomacy;
+
+@ValueData(name = "target-is", description = "Checks the target type", example = "target-is(Enemy)")
```
Pablete1234, line 18: for flexibility purposes, i believe this function should be getting for a specific ship as parameter, maybe similar to has-effect: `is-relationship(target(), NPC)` or similar

src/main/java/com/github/manolo8/darkbot/config/actions/conditions/TargetTypeCondition.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        target = Value.get(ship, main);
+        if (target == null || !target.isValid()) {
+            return type == TargetType.NO_TARGET ? Result.ALLOW : Result.DENY;
+        }
+
+        if (type == TargetType.NPC && target instanceof Npc) {
+            return Result.ALLOW;
+        }
```
Pablete1234, line 41: if the type is npc, it allows, but if it isn't? it doesn't seem to deny, and will probably just abstain.

consider rewriting this into a switch:
```
private boolean matches(Ship target) {
  switch (type) {
    case NO_TARGET: return target == null || !target.isValid();
    case NPC: return target instanceof Npc;
    case ENEMY: return target != null && target.getEntityInfo.isEnemy();
    ---
  }
}
```

then simply return `Result.fromBool(matches(target))`

src/main/java/com/github/manolo8/darkbot/config/actions/conditions/TargetTypeCondition.java
```diff
@@ -0,0 +1,97 @@
...
+        ALLIED,
+        NOT_ATTACK_PACT;
+
+        @Override
+        public String toString() {
+            return name().toLowerCase(Locale.ROOT).replace("_", "-");
+        }
+
+        public static TargetType of(String targetType) {
+            for (TargetType tT : TargetType.values()) {
```
Pablete1234, line 67:
```suggestion
            for (TargetType tt : TargetType.values()) {
```
Avoid capitalizing abreviated variable names

Pablete1234 (approved):

PR #305, opened 2023-08-21
Added booty keys stats and premium

dm94:

PR #308, opened 2023-08-28
Added booty keys stats and premium V2

dm94:
Same as https://github.com/darkbot-reloaded/DarkBot/pull/305 but with refactoring


dm94:
Updated https://github.com/darkbot-reloaded/DarkBotAPI/pull/86

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
```diff
@@ -91,7 +101,10 @@ public void tick() {
             instance = main.settingsManager.readString(0x298);
         }
 
-        novaEnergy.track(readInt(0x108, 0x28));
+        novaEnergy.track(readInt(0x100, 0x28));
+        teleportBonus.track(readInt(0x50));
+        premium = API.readBoolean((API.readMemoryLong(address + 0xF0) & ByteUtils.ATOM_MASK) + 0x20);
```
Pablete1234, line 106:
```suggestion
        premium = readBoolean(0xF0, 0x20);
```

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
```diff
@@ -299,4 +335,47 @@ public String toString() {
...
+        ASTRAL(0x7c),
+        ASTRAL_SUPREME(0x80),
+        EMPYRIAN(0x84),
+        LUCENT(0x88),
+        PERSEUS(0x8c);
+
+        private int offset;
+
+        private BootyKeyType(int offset) {
+            this.offset = offset;
```
Pablete1234, line 359:
```suggestion
        private final int offset;
        private final StatsAPI.Key key;
        private StatImpl stat;

        private BootyKeyType(int offset, StatsAPI.Key statKey) {
            this.offset = offset;
```
If you add the stat key, and simply pass in `Stats.BootyKey.RED` etc in the constructor, you can more directly correlate the stats enum to this enum.

Then you can add a simple:

```java
public void track(StatsManager stats) {
  if (stat == null) stat = stats.get(StatKey.of(key));
  stat.track(stats.readDouble(offset));
}
```
in the enum, and call it for each of the enum constants

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
```diff
@@ -40,9 +42,12 @@ public class StatsManager implements Manager, StatsAPI, NativeUpdatable {
     private final Map<StatKey, StatImpl> statistics = new HashMap<>();
 
     private final StatImpl runtime;
-    private final StatImpl credits, uridium, experience, honor, cargo, maxCargo, novaEnergy;
+    private final StatImpl credits, uridium, experience, honor, cargo, maxCargo, novaEnergy, teleportBonus;
     private final AverageStats cpuStat, pingStat, tickStat, memoryStat;
 
+    private int teleportBonusAmount;
```
Pablete1234, line 48:
```suggestion
```
should be obtained from the stat itself

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
dm94: Updated

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
```diff
@@ -91,7 +98,10 @@ public void tick() {
             instance = main.settingsManager.readString(0x298);
         }
 
-        novaEnergy.track(readInt(0x108, 0x28));
+        novaEnergy.track(readInt(0x100, 0x28));
```
Pablete1234, line 101: Are you sure about this revert from 108 to 100?

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
```diff
@@ -299,4 +322,38 @@ public String toString() {
...
+        EMPYRIAN(0x84, Stats.BootyKey.EMPYRIAN),
+        LUCENT(0x88, Stats.BootyKey.LUCENT),
+        PERSEUS(0x8c, Stats.BootyKey.PERSEUS);
+
+        private int offset;
+        private final StatsAPI.Key key;
+
+        private BootyKeyType(int offset, StatsAPI.Key key) {
+            this.offset = offset;
+            this.key = key;
```
Pablete1234, line 348:
```suggestion
            this.key = StatKey.of(key)
```
If you store a stat key directly, and make the track method use it, you'll avoid the N allocations each tick

Pablete1234 (approved):

PR #317, opened 2023-09-19
add inventory proxy data

anjeshshrestha:


src/main/java/com/github/manolo8/darkbot/core/objects/facades/InventoryProxy.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static com.github.manolo8.darkbot.Main.API;
+
+@Getter
+public class InventoryProxy extends Updatable implements InventoryAPI {
+    @Getter(AccessLevel.NONE)
+    private final ObjArray itemsArr = ObjArray.ofVector(true);
+
+    public List<Item> items = new ArrayList<>();
+    private final Timer updateTimer = Timer.get(10_000);
```
Pablete1234, line 23: maybe instead of just a hard-coded timer to re-fetch all items every 10s, also consider doing it only if some plugins is polling the api (ie: calling getItems()). Basically define getItems and make that method be the one that calls update (if timer.tryactivate) and then returns the result.

src/main/java/com/github/manolo8/darkbot/core/objects/facades/InventoryProxy.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public void update() {
+        if (updateTimer.tryActivate()) {
+            itemsArr.update(Main.API.readMemoryPtr(address, 0x30, 0x58));
+            itemsArr.syncAndReport(items, Item::new);
```
Pablete1234, line 29: Sync and report doesn't seem to be used for anything, you don't care if an update was made or not, so why use it?

src/main/java/com/github/manolo8/darkbot/core/objects/facades/InventoryProxy.java
```diff
@@ -0,0 +1,52 @@
...
+import lombok.ToString;
+
+import java.util.List;
+
+import static com.github.manolo8.darkbot.Main.API;
+
+@Getter
+public class InventoryProxy extends Updatable implements InventoryAPI {
+    @Getter(AccessLevel.NONE)
+    public FlashList<Item> items = FlashList.ofVector(Item::new);
```
Pablete1234, line 19:
```suggestion
    private FlashList<Item> items = FlashList.ofVector(Item::new);
```

PR #318, opened 2023-09-20
added possibility to disable fake dailyLogin.php in KekkaPlayer

Punikekk:


src/main/java/com/github/manolo8/darkbot/core/IDarkBotAPI.java
```diff
@@ -134,6 +134,7 @@ default void setVisible(boolean visible, boolean fullyHideEnabled) {
     }
 
     void handleRefresh();
```
Pablete1234, line 136: this can probably just default to calling handleRefresh with a true, and change GameAPIImpl to have a single method (always with a param)

Punikekk: I'm not sure it won't mess up some plugins

src/main/java/com/github/manolo8/darkbot/core/api/GameAPIImpl.java
```diff
@@ -447,18 +447,26 @@ public void resetCache() {
...
-        handler.reload();
-
+        reload(useFakeDailyLogin);
         extraMemoryReader.resetCache();
     }
 
+    protected void reload(boolean useFakeDailyLogin) {
+        handler.reload();
+    }
+
```
dm94, line 469: Why create a new function that does the same as the old one?
If you do not use the boolean of useFakeDailyLogin you do not need a new function asking for it.

Punikekk: It's default implementation, as you see I further override this method in KekkaPlayerAdapter

src/main/java/com/github/manolo8/darkbot/core/IDarkBotAPI.java
Pablete1234: i don't see how it would mess up plugins, just:
```java
default void handleRefresh() {
  handleRefresh(false);
}
```
it should be the exact same behavior as now

Pablete1234 (approved):

PR #320, opened 2023-09-22
Close windows via method call, minimize guis only if bot is running

Punikekk:
attempt to fix targeted offer missclicks

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/objects/Gui.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/objects/Gui.java
@@ -109,9 +109,14 @@ public class Gui extends SpriteObject implements API, eu.darkbot.api.game.other.
 
     public boolean show(boolean value) {
         if (trySetShowing(value)) {
-            if (!(minimizable && toggleVisibility()) && !(closable && close()))
-                legacyToggle(value);
+            // visibility state != value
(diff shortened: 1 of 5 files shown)


src/main/java/com/github/manolo8/darkbot/core/objects/Gui.java
```diff
@@ -108,8 +109,8 @@ public void moveWindow(int newX, int newY) {
 
     public boolean show(boolean value) {
         if (trySetShowing(value)) {
-            if (minimizable) toggleVisibility();
-            else legacyToggle(value);
+            if (!(minimizable && toggleVisibility()) && !(closable && close()))
+                legacyToggle(value);
```
Pablete1234, line 113:
```suggestion
            if (minimizable ? !toggleVisibility() : closeable ? !close() : false) {
               legacyToggle(value);
             }
```

This logic is kinda confusing and hard to follow, what exactly is the intended flow?

if minimizable -> toggleVisibility (and fallback to legacyToggle)
if closeable -> close (and fallback to legacyToggle)
if neither? do nothing? or legacyToggle?

Shouldn't `value` be used within this method? (ie, what the end goal state is) such that `visible ? toggleVisibility() : close()` ?

Pablete1234 (approved):

PR #323, opened 2023-10-19
add agatus proxy

anjeshshrestha:


src/main/java/com/github/manolo8/darkbot/core/objects/facades/AgatusEventProxy.java
```diff
@@ -0,0 +1,6 @@
+package com.github.manolo8.darkbot.core.objects.facades;
+
+import eu.darkbot.api.managers.AgatusAPI;
+
+public class AgatusEventProxy extends DefaultNpcEventProxy implements AgatusAPI {
```
Pablete1234, line 5: It seems like a pretty bad pattern to be extending a class without any changes at all. Is the issue here simply that this should be a higher-level api?

Instead of a `DefaultNpcEventAPI` maybe it should be a `NpcEventAPI` which returns a list or a map of `NpcEvent` objects.

```java
interface NpcEventAPI {

  NpcEvent getEvent(EventType type);
  
  enum EventType {
    GENERIC, AGATUS, ...;
  }
}
```

The implementation can use the Agatus proxy and the generic npc event proxy, and in the plugin-side you can:
`NpcEventAPI.getEvent(EventType.AGATUS)` to get an `NpcEvent` with the info related to agatus.

Pablete1234:
Added in b902589e

PR #324, opened 2023-10-22
Fake entities & minor changes

m9w:
It allow customize behavior of collector module & pathfinder without overriding methods these modules. Was added handling of loading raw bytecode. Also was suppressed showing instructions windows if bot running.

This changes required API changes https://github.com/darkbot-reloaded/DarkBotAPI/pull/91

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/Bot.java
+++ b/src/main/java/com/github/manolo8/darkbot/Bot.java
@@ -55,7 +55,11 @@ public class Bot {
             UIManager.put("Table.cellFocusColor", new Color(0, 0, 0, 160));
 
             // set icon here to use it in login popup, java check popup etc.
-            Popups.setDefaultIcon(MainGui.ICON);
+            try {
+                Popups.setDefaultIcon(MainGui.ICON);
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
 
             // Not recommended to keep for production
             //FlatInspector.install("ctrl shift alt X");
--- a/src/main/java/com/github/manolo8/darkbot/core/entities/Box.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/entities/Box.java
@@ -3,7 +3,9 @@ package com.github.manolo8.darkbot.core.entities;
 import com.github.manolo8.darkbot.config.BoxInfo;
 import com.github.manolo8.darkbot.config.ConfigEntity;
 import com.github.manolo8.darkbot.core.api.Capability;
+import com.github.manolo8.darkbot.core.manager.HeroManager;
 import com.github.manolo8.darkbot.utils.Offsets;
+import eu.darkbot.api.game.other.Location;
 import eu.darkbot.util.Timer;
 import org.jetbrains.annotations.Nullable;
 
@@ -25,6 +27,10 @@ public class Box extends Entity implements eu.darkbot.api.game.entities.Box {
 
     public BoxInfo boxInfo;
 
+
+    private Box(int id) {
+        super(id);
+    }
 
     public Box(int id, long address) {
         super(id);
@@ -134,5 +140,62 @@ public class Box extends Entity implements eu.darkbot.api.game.entities.Box {
         public Beacon(int id, long address) {
             super(id, address);
         }
+    }
+
+    public static class FakeBox extends Box implements eu.darkbot.api.game.entities.FakeEntity.FakeBox {
+        private static int CURR_ID = Integer.MIN_VALUE;
+        private Timer timeout;
+        private long removeDistance;
+        private boolean isRemoveWhenAttemptSelect;
+
+        public FakeBox(String boxName, Location loc, long removeDistance, long keepAlive, boolean isRemoveWhenAttemptSelect) {
+            super(CURR_ID++);
+            super.locationInfo.updatePosition(loc.x(), loc.y());
+            super.main = HeroManager.instance.main;
+            super.type = boxName;
+            super.boxInfo = ConfigEntity.INSTANCE.getOrCreateBoxInfo(type);
+            super.removed = false;
+            setRemoveDistance(removeDistance);
+            setTimeout(keepAlive);
+            setRemoveWhenAttemptSelect(isRemoveWhenAttemptSelect);
+        }
+
+        public void setTimeout(long keepAlive) {
+            if (keepAlive != -1) {
+                timeout = Timer.get(keepAlive);
+                timeout.activate();
+            }
+            else timeout = null;
+        }
+
+        public void setRemoveDistance(long removeDistance) {
+            this.removeDistance = removeDistance;
+        }
+
+        public void setRemoveWhenAttemptSelect(boolean removeWhenAttemptSelect) {
+            isRemoveWhenAttemptSelect = removeWhenAttemptSelect;
+        }
+
+        public String getHash() {
+            return type + locationInfo.getCurrent();
+        }
+
+        public boolean tryCollect() {
+            return trySelect(false);
+        }
+
+        public boolean isInvalid(long mapAddress) {
+            if (timeout != null && timeout.isInactive()) return false;
+            return HeroManager.instance.distanceTo(this) < removeDistance;
+        }
+
+        public boolean trySelect(boolean tryAttack) {
+            if (isRemoveWhenAttemptSelect) removed();
+            return false;
+        }
+
+        public void update() {}
+
+        public void update(long address) {}
     }
 }
--- a/src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
@@ -141,6 +141,7 @@ public class GuiManager implements Manager, GameScreenAPI {
         this.assembly = register("assembly");
 
         register("ggBuilder", GateSpinnerGui.class);
+        register("refinement_count");
 
         this.commandCenter = register("command_center");
 
--- a/src/main/java/com/github/manolo8/darkbot/core/utils/EntityList.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/utils/EntityList.java
@@ -23,6 +23,7 @@ import com.github.manolo8.darkbot.core.itf.Updatable;
 import com.github.manolo8.darkbot.core.objects.swf.ObjArray;
 import com.github.manolo8.darkbot.core.utils.factory.EntityFactory;
 import com.github.manolo8.darkbot.core.utils.factory.EntityRegistry;
+import eu.darkbot.api.game.entities.FakeEntity;
 import eu.darkbot.api.game.entities.Mist;
 import eu.darkbot.api.game.entities.Station;
 import eu.darkbot.api.managers.EntitiesAPI;
@@ -126,6 +127,24 @@ public class EntityList extends Updatable implements EntitiesAPI {
             this.obstacles.add((Obstacle) entity);
 
         this.eventBroker.sendEvent(new EntityCreateEvent(entity));
+    }
+
+    public FakeEntity.FakeMine createFakeMine(int typeId, eu.darkbot.api.game.other.Location  loc, long removeDistance, long keepAlive) {
+        return register(new Mine.FakeMine(typeId, loc, removeDistance, keepAlive), mines);
+    }
+
+    public FakeEntity.FakeBox createFakeBox(String boxName, eu.darkbot.api.game.other.Location loc, long removeDistance, long keepAlive, boolean removeIfAttemptSelect) {
+        return register(new Box.FakeBox(boxName, loc, removeDistance, keepAlive, removeIfAttemptSelect), boxes);
+    }
+
+    public FakeEntity.FakeShip createFakeNpc(String npcName, eu.darkbot.api.game.other.Location loc, long removeDistance, long keepAlive, boolean removeIfAttemptSelect) {
+        return register(new Npc.FakeNpc(npcName, loc, removeDistance, keepAlive, removeIfAttemptSelect), npcs);
+    }
+
+    private <T extends Entity> T register(T entity, Collection<? super T> collection) {
+        collection.add(entity);
+        onEntityCreate(entity);
+        return entity;
     }
 
     @SuppressWarnings("unchecked")
--- a/src/main/java/com/github/manolo8/darkbot/extensions/features/decorators/InstructionProviderDecorator.java
+++ b/src/main/java/com/github/manolo8/darkbot/extensions/features/decorators/InstructionProviderDecorator.java
@@ -1,5 +1,6 @@
 package com.github.manolo8.darkbot.extensions.features.decorators;
 
+import com.github.manolo8.darkbot.config.ConfigEntity;
 import com.github.manolo8.darkbot.extensions.features.FeatureDefinition;
 import com.github.manolo8.darkbot.gui.utils.Popups;
 import com.github.manolo8.darkbot.utils.I18n;
@@ -13,7 +14,10 @@ public class InstructionProviderDecorator extends FeatureDecorator<InstructionPr
 
     @Override
     protected void load(FeatureDefinition<InstructionProvider> fd, InstructionProvider obj) {
-        if (obj instanceof Module && !(obj instanceof TemporalModule)) showInstructions(obj, fd.getName());
+        if (obj instanceof Module
+                && !(obj instanceof TemporalModule)
+                && ConfigEntity.INSTANCE.getConfig().BOT_SETTINGS.OTHER.SHOW_INSTRUCTIONS
+        ) showInstructions(obj, fd.getName());
     }
 
     @Override
--- a/src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginClassLoader.java
+++ b/src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginClassLoader.java
@@ -36,6 +36,13 @@ public class PluginClassLoader extends URLClassLoader {
         super(urls);
     }
 
+    @SuppressWarnings("unused")
+    public Class<?> defineClass(String name, byte[] bytes) throws ClassNotFoundException {
+        if (PROTECTED.stream().anyMatch(p -> p.test(name)))
+            throw new ClassNotFoundException(name + " is a protected class");
+        return super.defineClass(name, bytes, 0 ,bytes.length);
+    }
+
     @Override
     public Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
         if (PROTECTED.stream().anyMatch(p -> p.test(name)))
--- a/src/main/java/com/github/manolo8/darkbot/config/Config.java
+++ b/src/main/java/com/github/manolo8/darkbot/config/Config.java
@@ -294,6 +294,7 @@ public class Config implements eu.darkbot.api.config.legacy.Config {
 
         public @Option @Visibility(Level.INTERMEDIATE) Other OTHER = new Other();
         public static class Other {
+            public @Option boolean SHOW_INSTRUCTIONS = true;
             public @Option boolean DISABLE_MASTER_PASSWORD = false;
             public @Option boolean ALWAYS_SHOW_CAPTCHA = false;
             public @Option @Number(min = 10, max = 300) int ZONE_RESOLUTION = 30;
--- a/src/main/java/com/github/manolo8/darkbot/core/entities/Mine.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/entities/Mine.java
@@ -1,8 +1,11 @@
 package com.github.manolo8.darkbot.core.entities;
 
 import com.github.manolo8.darkbot.core.itf.Obstacle;
+import com.github.manolo8.darkbot.core.manager.HeroManager;
 import com.github.manolo8.darkbot.core.utils.pathfinder.AreaImpl;
 import com.github.manolo8.darkbot.core.utils.pathfinder.CircleImpl;
+import eu.darkbot.api.game.other.Location;
+import eu.darkbot.util.Timer;
 
 import static com.github.manolo8.darkbot.Main.API;
 
@@ -13,6 +16,10 @@ public class Mine extends Entity implements Obstacle, eu.darkbot.api.game.entiti
     public int typeId;
 
     private final CircleImpl area = new CircleImpl(0, 0, 200);
+
+    private Mine(int id) {
+        super(id);
+    }
 
     public Mine(int id, long address) {
         super(id);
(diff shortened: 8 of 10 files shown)


src/main/java/com/github/manolo8/darkbot/extensions/features/decorators/InstructionProviderDecorator.java
```diff
@@ -13,7 +14,9 @@ public class InstructionProviderDecorator extends FeatureDecorator<InstructionPr
 
     @Override
     protected void load(FeatureDefinition<InstructionProvider> fd, InstructionProvider obj) {
-        if (obj instanceof Module && !(obj instanceof TemporalModule)) showInstructions(obj, fd.getName());
+        if (obj instanceof Module && !(obj instanceof TemporalModule)
+                && !HeroManager.instance.main.isRunning()
+        ) showInstructions(obj, fd.getName());
```
Pablete1234, line 19: I think the pop-up should still appear if the bot is running, but what should change the behavior is as a result of what the change is. If the user changed module i think it's good that they get the pop-up, but i agree if it's an automatic change it shouldn't be shown.
I'm not exactly sure how to properly detect if it was automatic or not, since plugins can be the ones changing the config themselves.

src/main/java/com/github/manolo8/darkbot/core/utils/EntityList.java
```diff
@@ -128,6 +129,21 @@ private void onEntityCreate(Entity entity) {
         this.eventBroker.sendEvent(new EntityCreateEvent(entity));
     }
 
+    //TODO: add to EntitiesAPI.class
```
Pablete1234, line 132: Agreed, needs to be added to the API.

Maybe also add the same treatment for createFakeNpc, and migrate the current FakeNpc (used for npc locator) to be more in-line with everything else here.

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginClassLoader.java
```diff
@@ -36,6 +36,13 @@ private static Predicate<String> toMatcher(String pattern) {
         super(urls);
     }
 
+    @SuppressWarnings("unused")
+    public Class<?> loadClass(String name, byte[] bytes) throws ClassNotFoundException {
+        if (PROTECTED.stream().anyMatch(p -> p.test(name)))
+            throw new ClassNotFoundException(name + " is a protected class");
+        return super.defineClass(name, bytes, 0 ,bytes.length);
```
Pablete1234, line 43:
```suggestion
        return super.loadClass(name, bytes);
```
I'd rather not skip steps, and still call exactly the same super method even if that is an instant redirection to defineClass

src/main/java/eu/darkbot/api/game/entities/FakeEntity.java
```diff
@@ -0,0 +1,51 @@
...
+    static boolean isFakeEntity(Entity entity) {
+        if (entity instanceof FakeEntity.FakeBox) return true;
+        if (entity instanceof FakeEntity.FakeMine) return true;
+        return false;
+    }
+
+    /**
+     * Fake box is a target for colector module if more prefer boxes is absent
+     */
+    interface FakeBox extends Box {
```
Pablete1234, line 20:
```suggestion
    interface FakeBox extends Box, FakeEntity {
```

src/main/java/eu/darkbot/api/game/entities/FakeEntity.java
```diff
@@ -0,0 +1,51 @@
...
+        /**
+         * Is remove fake entity after attempt select it
+         */
+        void setRemoveWhenAttemptSelect(boolean removeWhenAttemptSelect);
+    }
+
+    /**
+     * Fake mine is an obstacle that counted by path builder
+     */
+    interface FakeMine extends Mine {
```
Pablete1234, line 40:
```suggestion
    interface FakeMine extends Mine, FakeEntity {
```

src/main/java/eu/darkbot/api/game/entities/FakeEntity.java
```diff
@@ -0,0 +1,51 @@
...
+ * It should use when need to assist or predict position these entities for active module.
+ * Cases of use: keeping mine locations for choose better path and prevent release it,
+ *               keeping box locations that should be collected after more prefer tasks.
+ */
+public interface FakeEntity extends Entity {
+    static boolean isFakeEntity(Entity entity) {
+        if (entity instanceof FakeEntity.FakeBox) return true;
+        if (entity instanceof FakeEntity.FakeMine) return true;
+        return false;
+    }
```
Pablete1234, line 15:
```suggestion
    static boolean isFakeEntity(Entity entity) {
        return entity instanceof FakeEntity;
    }
```
If the Fake* classes actually inherit the interface, it can be simplified.

Could also actually include generic fake-stuff behavior in here,
i see a good candidate being `setEnabled(true/false)` where if you disable the fake entity it behaves as if it wasn't there at all (ie: starts returning false in isValid()) and you can just arbitrarily set it back to true later to make it relevant again. setTimeout and setRemoveDistance also seem good candidates for the generic FakeEntity

Pablete1234 (changes requested):
Generally looking good, just a few tweaks and i'd like to see the semantics of current FakeNpc be brought close together to the other fakes.

Ontop of that, while this is a great API to have, i think the bot itself should be keeping box/mine fakes without needing a behavior/module to do it.

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginClassLoader.java
m9w: loadClass don't have this signature, so I renamed new method (same to super method)
This change help exclude creating child classloader in plugin code for loading generated in-flight bytecode.

src/main/java/com/github/manolo8/darkbot/extensions/features/decorators/InstructionProviderDecorator.java
m9w: Was added API method for changing config and this call suppressing showing instructions. Also was added default true setting property for suppress showing instructions in all cases.

src/main/java/com/github/manolo8/darkbot/core/utils/EntityList.java
m9w: Moved to API. PR in description.

m9w: Also, added not implemented API for fake ship type.

src/main/java/eu/darkbot/api/game/entities/FakeEntity.java
m9w: implemented in API PR

src/main/java/eu/darkbot/api/game/entities/FakeEntity.java
m9w: implemented in API PR

m9w: I not want allow hiding entities from user code. I mean closed lifecycle where fake entity is required - it created and add on map immediately and when it nor required then it removing without possibility add it again. Only creating new instance.

src/main/java/eu/darkbot/api/game/entities/FakeEntity.java
m9w: implemented in API PR

m9w:
[CAETest.java.txt](https://github.com/darkbot-reloaded/DarkBot/files/13628232/CAETest.java.txt)
module for testing cross-applications events

Pablete1234:
Please rebase the changes instead of merging, it makes reviewing much harder as all the changes already in the bot since the start of the PR now appear as changes, so its hard to tell what's new and what's not

m9w:
@Pablete1234 changes was rebased, chery-picked, squashed and force pushed.

src/main/java/com/github/manolo8/darkbot/config/ConfigHandler.java
```diff
@@ -120,4 +128,13 @@ public void onPluginUnload(ExtensionsAPI.PluginLifetimeEvent e) {
         }
     }
 
+    @Override
+    public Collection<String> options() {
+        return Stream.concat(Stream.of("(none)"), loader.getAvailableConfigs().stream()).collect(Collectors.toList());
```
Pablete1234, line 133: none should not be an option, the option is `null`, and it's `getText` that should return "(none)" for the null case

src/main/java/com/github/manolo8/darkbot/config/tree/handlers/DropdownHandler.java
```diff
@@ -20,7 +20,7 @@ public static DropdownHandler of(Field field, PluginInfo namespace, PluginAPI ap
 
         Class<? extends Dropdown.Options<?>> optionCl = dropdown.options();
         LazyValue<Dropdown.Options<?>> options = optionCl != Dropdown.NullOptions.class ?
-                LazyValue.of(() -> api.requireInstance(optionCl)) :
+                LazyValue.of(() -> optionCl.isInterface() ? api.requireAPI(optionCl) : api.requireInstance(optionCl)) :
```
Pablete1234, line 23: interfaces should really *not* be supported for dropdowns, just have specific *instances* for the dropdowns, not apis

src/main/java/com/github/manolo8/darkbot/core/entities/Box.java
```diff
@@ -135,4 +141,61 @@ public Beacon(int id, long address) {
...
         }
     }
+
+    public static class FakeBox extends Box implements eu.darkbot.api.game.entities.FakeEntity.FakeBox {
+        private Timer timeout;
+        private long removeDistance;
+        private boolean isRemoveWhenAttemptSelect;
+
+        public FakeBox(String boxName, Location loc, long removeDistance, long keepAlive, boolean isRemoveWhenAttemptSelect) {
+            super(Integer.MIN_VALUE);
```
Pablete1234, line 151: should probably use different IDs for these, could probably just hold a global fake entity id counter and keep on using extremely low numbers (counting up from min value)

src/main/java/com/github/manolo8/darkbot/core/manager/CrossApplicationEventManager.java
```diff
@@ -0,0 +1,169 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.Queue;
+import java.util.UUID;
+
+
+@Feature(name = "Cross-application event", description = "Allow send events between applications")
+public class CrossApplicationEventManager implements Task, Installable, CrossApplicationAPI {
```
Pablete1234, line 27: This whole feature *really* feels like it's just a plugin feature that has been copy-pasted into the bot.

I believe the bot should offer a way to setup external communication that the plugins can use, but i don't think this is the right approach, because it takes the *whole* responsability of the system on the bot, which also makes it rather rigid. No option for 2 plugins to use this system at the same time for example.

Consider making this more of a "you can request a socket server be setup by the bot, and the bot takes care of it" but you're still the one who does everything else, meaning config for what addr to connect to, or the general layout of messaging. The bot simply provides the back-end for establishing the connection and handling multithreading if need be, but the messaging logic should be on the plugin side.

src/main/java/com/github/manolo8/darkbot/extensions/features/decorators/InstructionProviderDecorator.java
```diff
@@ -39,4 +45,11 @@ public static void showInstructions(Object featureInstance, String name) {
         }
     }
 
+    private boolean isApiCall() {
+        for (StackTraceElement el : new Exception().getStackTrace()) {
+            if (ConfigHandler.class.getName().equals(el.getClassName()) && "changeConfig".equals(el.getMethodName()))
+                return true;
+        }
```
Pablete1234, line 52: This is NOT a valid way to check, and will probably never be true anyways since it happens async. (ie: you set a lazy, and when the lazy gets consumed in the next tick is when the change actually occurs)

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginClassLoader.java
Pablete1234: Plugins should not be generating in-flight bytecode, that sounds totally out of the scope of what plugins should be doing, and if this is not an override but a new method i'd want it not to be added at all unless you can give me a good reason for this

src/main/java/com/github/manolo8/darkbot/core/manager/CrossApplicationEventManager.java
m9w: Yes it was wrote as plugin and it can be run as plugin, but it required API PR https://github.com/darkbot-reloaded/DarkBotAPI/pull/92 for make possible general entry point from different places. 

It's only transport layer where exist mandatory type_id implement via UUID. It not have limitation for features that communicate over this channel so all plugin features can send messages in parallel and receive this messages known UUID it can be sure in structure before parsing it.

UUID is a marker show who send this message.

>This feature only send base64(UUID+message)+'\n' to router and generate OnMessage events when receive message in this format from another bot. It's all.

This idea is all that need to know developer of creating plugin code that work over this interface.

This code need in repo for has non-blocking connection that configured in single place.

src/main/java/com/github/manolo8/darkbot/config/tree/handlers/DropdownHandler.java
m9w: Ok, I will remove it from PR. But I think that this suppliers can be implemented into managers too. I not understood why API interfaces cannot has dropdown option and you want keep implementation in API repo.
image

src/main/java/com/github/manolo8/darkbot/core/entities/Box.java
m9w: what about random id? `super(ThreadLocalRandom.current().nextInt(Integer.MIN_VALUE, -1000));`
but I found where it can be used and not sure that it's matter

anjeshshrestha: lets not use randomly generated value here, a defined value or add boolean check to tag them as a "fake" to show the entity is as fake

m9w: > lets not use randomly generated value here, a defined value or add boolean check to tag them as a "fake" to show the entity is as fake

it's random negative values. For detecting fake can use https://github.com/darkbot-reloaded/DarkBotAPI/pull/91/files#diff-21bb577e67916994b49cca980705fdb72d7f47329f09bac1596e907da599c906R12

src/main/java/com/github/manolo8/darkbot/config/tree/handlers/DropdownHandler.java
Pablete1234: The shared package is still inside the API repo, not this one, but as an instance not as an interface, see https://github.com/darkbot-reloaded/DarkBotAPI/blob/master/shared/src/main/java/eu/darkbot/shared/config/NpcNames.java

src/main/java/com/github/manolo8/darkbot/core/entities/Box.java
Pablete1234: So there's two issues here: the game itself already uses negative ids for some stuff, starts with -1, then -2, etc, and it likely will reach -1000 if you run the bot for long enough.

Secondly, with random values it's possible to end with a collision, and i'd want to avoid that.

I don't see the issue in simply having `private static int CURR_ID = Integer.MIN_VALUE;` and assigning id as `super(CURR_ID++)`, which should generate unique, sequantial, negative ids for everything.

And yeah, as to what anjesh mentioned, the ID is not a way to determine if they're fake entities, check via instanceOf

src/main/java/com/github/manolo8/darkbot/core/manager/CrossApplicationEventManager.java
Pablete1234: You make it sound like determining message format to be base64 only, and always prefixed by uuid, is not a limiting factor.

Again this is YOUR implementation that works for YOU, but the point of an API is to be able to be used by others too. If a plugin wants to establish a connection with an already-existing service, that communicates with a different scheme (eg: json) then what? This offers too little flexibility and it's just too-tailored to your specific use-case and your specific implementation decisions

src/main/java/com/github/manolo8/darkbot/core/entities/Box.java
m9w: if use `private static int CURR_ID = Integer.MIN_VALUE;` and `super(CURR_ID++)` than we have `2 147 483 648` negative values so when we create more entities then IDs will be positive. I think it not good thing. 

P.S. I made it like it wrote in FakeNPC class
image

src/main/java/com/github/manolo8/darkbot/core/manager/CrossApplicationEventManager.java
m9w: If plugin developer want to establish a connection with an already-existing service using low level messaging than he should know UUIDs of this service and listen responses from this service. This low level api need for establish a connection between feature instance in linked apps (not third-perty service it possible but it will required implementation a protocol of this service inside plugin). 
Message content not specified it can be json, xml, serialized java object - no matter. 
image

Now `this api + simple router` works like UDP broadcast. Where router address is a subnet for broadcast, UUID is a port, message is a datagram content. I not use bytes for message type because string easy parsing. because it easy parsing.
Over this api can be implemented RMI and RPC.
I don't agree that it has little flexibility.

src/main/java/com/github/manolo8/darkbot/core/entities/Box.java
Punikekk: creating over 2 bilions of entities seems unreal for me, even with runtime over a year
year have over 30 milions of seconds, so you have to create ~70 entities every second to fill it up in a year

Pablete1234: Sorry i actually meant CURR_ID-- not ++

src/main/java/com/github/manolo8/darkbot/core/manager/CrossApplicationEventManager.java
Pablete1234: > If plugin developer want to establish a connection with an already-existing service using low level messaging than he should know UUIDs of this service

An already-existing service, will **not** have a uuid-prefixed message nor base64 encoding, nor be running in the same server as the other plugin.

m9w: Why you focus on base64 format?, it no mater for plugin code side. it need only for debugging message sequence on router. And server is a router app in this model. Each bot is a client that ready to communicate. Using static UUID allow to find remote instances of feature like works ARP protocol in networking. There UUID is analog of MAC address. 

In this moment we not have any built-in services for communication and each developer can implement itself. it give freedom in building personal model of communication. Because I implemented only transport layer.

It not have limitation of communication between different features but it not prefer task. And it not have limitation for communicating between single app services but for this action need to create OnMessage event inside feature code.

m9w:
All previous changes was reverted and was created ConfigProfile, BotModule interfacres 
that allow set config profile or Module that selected in config in way calling apply() method. It haven't mixing UI and API classes.

src/main/java/com/github/manolo8/darkbot/config/tree/handlers/DropdownHandler.java
m9w: reverted

src/main/java/com/github/manolo8/darkbot/core/entities/Box.java
m9w: implemented

src/main/java/com/github/manolo8/darkbot/extensions/features/decorators/InstructionProviderDecorator.java
m9w: reverted

src/main/java/com/github/manolo8/darkbot/extensions/features/decorators/InstructionProviderDecorator.java
m9w: implement via checking config value

src/main/java/com/github/manolo8/darkbot/config/ConfigHandler.java
m9w: reverted

src/main/java/com/github/manolo8/darkbot/core/manager/CrossApplicationEventManager.java
m9w: reverted from this PR, will implemented in another place

src/main/java/com/github/manolo8/darkbot/extensions/features/decorators/InstructionProviderDecorator.java
```diff
@@ -13,7 +14,10 @@ public class InstructionProviderDecorator extends FeatureDecorator<InstructionPr
 
     @Override
     protected void load(FeatureDefinition<InstructionProvider> fd, InstructionProvider obj) {
-        if (obj instanceof Module && !(obj instanceof TemporalModule)) showInstructions(obj, fd.getName());
+        if (obj instanceof Module
+                && !(obj instanceof TemporalModule)
+                && ConfigEntity.INSTANCE.getConfig().BOT_SETTINGS.OTHER.SHOW_INSTRUCTIONS
+        ) showInstructions(obj, fd.getName());
```
Pablete1234, line 20:
```suggestion
                && ConfigEntity.INSTANCE.getConfig().BOT_SETTINGS.OTHER.SHOW_INSTRUCTIONS)
                    showInstructions(obj, fd.getName());
```
minor formatting change

src/main/java/com/github/manolo8/darkbot/extensions/plugins/PluginClassLoader.java
Pablete1234: This still needs to be reverted for this to be merged

src/main/java/com/github/manolo8/darkbot/core/utils/EntityList.java
```diff
@@ -128,6 +129,24 @@ private void onEntityCreate(Entity entity) {
         this.eventBroker.sendEvent(new EntityCreateEvent(entity));
     }
 
+    public FakeEntity.FakeMine createFakeMine(int typeId, eu.darkbot.api.game.other.Location  loc, long removeDistance, long keepAlive) {
```
Pablete1234, line 132:
```suggestion
    public FakeEntity.FakeMine createFakeMine(int typeId, eu.darkbot.api.game.other.Location loc, long removeDistance, long keepAlive) {
```
minor formatting change

src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
```diff
@@ -141,6 +141,7 @@ public GuiManager(Main main, PluginAPI pluginAPI, RepairManager repairManager) {
         this.assembly = register("assembly");
 
         register("ggBuilder", GateSpinnerGui.class);
+        register("refinement_count");
```
Pablete1234, line 144: does this not belong to a diff PR?

src/main/java/com/github/manolo8/darkbot/core/entities/Box.java
```diff
@@ -135,4 +141,61 @@ public Beacon(int id, long address) {
             super(id, address);
         }
     }
+
+    public static class FakeBox extends Box implements eu.darkbot.api.game.entities.FakeEntity.FakeBox {
+        private static int CURR_ID = Integer.MIN_VALUE;
```
Pablete1234, line 146: needs to be shared by all fake entities, to avoid duplicate ids. Just have this sit in some util where you can `EntityUtil.allocateFakeId()` and it returns CURR_ID++ or similar

src/main/java/com/github/manolo8/darkbot/Bot.java
```diff
@@ -55,7 +55,11 @@ public static void main(String[] args) throws IOException {
             UIManager.put("Table.cellFocusColor", new Color(0, 0, 0, 160));
 
             // set icon here to use it in login popup, java check popup etc.
-            Popups.setDefaultIcon(MainGui.ICON);
+            try {
+                Popups.setDefaultIcon(MainGui.ICON);
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
```
Pablete1234, line 62: i wonder, what's the motive behind this change? is it throwing an exception under some systems? if so, which exception?

Pablete1234 (changes requested):

Pablete1234:
Given the commit was abandoned i've re-coded it, implemented the updated fake entities api, and also introduced the option for instructions.
If you feel other important changes were left out feel free to open separate PRs for each of the feature changes you want us to consider, avoid making mega-prs which merge different features.

PR #326, opened 2023-11-01
Added FlashMap & FlashList, support collections in ObjectInspector

Punikekk:


src/main/java/com/github/manolo8/darkbot/gui/utils/inspector/InspectorTree.java
```diff
@@ -61,8 +65,12 @@ public void treeWillCollapse(TreeExpansionEvent event) throws ExpandVetoExceptio
             public void ancestorAdded(AncestorEvent event) {
                 if (timer == null) {
                     timer = new Timer(delay, e -> {
-                        ((ObjectTreeNode) model.getRoot()).update(InspectorTree.this);
-                        invalidate();
+                        try {
+                            ((ObjectTreeNode) model.getRoot()).update(InspectorTree.this);
+                            invalidate();
+                        } catch (Throwable t) {
+                            t.printStackTrace();
```
Pablete1234, line 72: why is this throwing exceptions in the first place?

Punikekk: due to memory misreads caused by invalid memory reads via different threads
anyway shouldn't happen anymore i hope

src/main/java/com/github/manolo8/darkbot/gui/utils/inspector/InspectorTree.java
```diff
@@ -116,26 +143,151 @@ private void editValue(boolean primitivesOnly) {
...
+            add(editTypeNameItem);
+        }
+
+        private void generateClass(ActionEvent e) {
+            ObjectTreeNode node = getSelectedNode();
+            if (node != null) {
+                ObjectInspector.Slot slot = node.getSlot();
+                if (slot.slotType == ObjectInspector.Slot.Type.OBJECT || slot.slotType == ObjectInspector.Slot.Type.PLAIN_OBJECT) {
+
+                    StringBuilder clazzBuilder = new StringBuilder("import com.github.manolo8.darkbot.core.itf.Updatable;");
```
Pablete1234, line 232: this thing def needs to go in a diff class, and consider making it a template, as in, a resource text-file with some stuff like {className}, {methods} or w/e that you can search-and-replace instead of having the whole thing written with appends

src/main/java/com/github/manolo8/darkbot/gui/utils/inspector/InspectorTree.java
Pablete1234: if it doesn't happen, can we get rid of it?

Pablete1234 (approved):

PR #330, opened 2023-11-14
Added manual captcha solver using DarkBrowser on first login

Punikekk:


src/main/java/com/github/manolo8/darkbot/utils/login/LoginUtils.java
```diff
@@ -55,6 +55,8 @@ public static LoginData performUserLogin(StartupParams params) {
             System.out.println("Closed login panel, exited without logging in");
             System.exit(0);
         }
+
+        LoginCaptchaTask.firstLogin = false;
```
Pablete1234, line 59: any good reason why only first login may do this? seems more reasonable that the browser itself simply has some time-out if no action is being done, but if the user is there they get the option to solve it at least

Punikekk: simply to do not interfere with maximized games, videos, etc. because captcha window is topmost.
there is actually a exit timeout `"--exit", String.valueOf(30_000)`

Pablete1234: If it's for that reason, maybe just have a config setting for it

Pablete1234 (approved):

PR #331, opened 2023-12-19
Multiple account feature

m9w:
it required https://github.com/darkbot-reloaded/DarkBotAPI/pull/90 API.

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/utils/login/LoginData.java
+++ b/src/main/java/com/github/manolo8/darkbot/utils/login/LoginData.java
@@ -7,12 +7,17 @@ import lombok.Getter;
 
 import java.lang.reflect.Type;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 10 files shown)


src/main/java/com/github/manolo8/darkbot/utils/login/LoginData.java
```diff
@@ -7,19 +7,26 @@
...
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiConsumer;
+import java.util.stream.Collectors;
 
 public class LoginData {
     private static final Type PARAMS_TYPE = new TypeToken<Map<String, String>>() {}.getType();
 
+    @Getter
+    private transient final Map<Integer, LoginData> accounts = new ConcurrentHashMap<>();
```
Pablete1234, line 20: Avoid contaminating the LoginData class with stuff that clearly is not from a single user's login data

Instead try to pass around either a List<LoginData> or another object that is able to support 1 or many login datas and handle the logic for switching around etc.

m9w:
Found solution for diminish quests so bot can be active full time. Closed

PR #334, opened 2024-01-04
Implement NativeBrowserAPI

Punikekk:


src/main/java/com/github/manolo8/darkbot/backpage/NativeBrowserImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private final BackpageManager backpageManager;
+
+    public NativeBrowserImpl(BackpageManager backpageManager) {
+        this.backpageManager = backpageManager;
+    }
+
+    @Override
+    public boolean isSupported() {
+        return BackpageTask.isSupported(new Version("1.3.0"));
```
Pablete1234, line 20: maybe consider having some static constants like:

`OPEN_LINK_VERSION = new Version(1.3.0);`

that way everywhere that you want to check for support, it will be like: `BackpageTask.isSupported(OPEN_LINK_VERSION)` which makes it way clearer what you're checking for, and allows a kinda version-history-like mapping of "name/description" to number

src/main/java/com/github/manolo8/darkbot/backpage/NativeBrowserImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public @Nullable Process openLink(String link) {
+        if (isSupported() && backpageManager.isInstanceValid()) {
+            try {
+                return BackpageTask.createBrowser("--sid " + backpageManager.getSid(),
+                        "--url " + backpageManager.getInstanceURI().toString(), // for sid
+                        "--fullurl " + link);
```
Pablete1234, line 29: This makes no sense, if you expect the SID to be passed i'd use a openGameLink, if you are opening any other link it'd expect it not to have sid passed, am i wrong?

Pablete1234 (approved):

PR #337, opened 2024-01-28
Moved to FlashList & FlashMap

Punikekk:
Added overloaded varargs method for memory reading Commit table changes on focus lost
Added tooltips for spinner's commit fail
Fixed inconsistency in Tables
Upgraded FlatLaf to 3.3
Many more little changes


src/main/java/com/github/manolo8/darkbot/core/MemoryAPI.java
```diff
@@ -0,0 +1,428 @@
...
+    default long readLong(long address, int o1, int o2) {
+        return readLong(readLong(address, o1) + o2);
+    }
+
+    default long readLong(long address, int o1, int o2, int o3) {
+        return readLong(readLong(address, o1, o2) + o3);
+    }
+
+    default long readLong(long address, int o1, int o2, int o3, int o4) {
+        return readLong(readLong(address, o1, o2, o3) + o4);
```
Pablete1234, line 76: shouldn't this be like:
```suggestion
        return readLong(readAtom(address, o1, o2, o3) + o4);
```
?

Pablete1234 (approved):

PR #338, opened 2024-01-29
remove backpage galaxygate spinner

anjeshshrestha:
remove old implementation of GG relied on backpage


Pablete1234:
For the time being this is staying as old legacy plugins may rely on it and it costs us nothing to keep

PR #340, opened 2024-02-21
Fix parsing PET gears

m9w:


src/main/java/com/github/manolo8/darkbot/backpage/hangar/ItemInfo.java
```diff
@@ -68,4 +76,20 @@ public String toString() {
...
+    private static class LevelsDeserializer implements JsonDeserializer<List<Map<String, Object>>> {
+        @Override
+        public List<Map<String, Object>> deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+            if (json.isJsonArray()) return context.deserialize(json, typeOfT);
+            if (json.isJsonObject()) {
+                final JsonObject obj = json.getAsJsonObject();
+                return obj.keySet().stream().map(key -> {
+                    Map<String, Object> map = context.deserialize(obj.get(key), Map.class);
+                    map.put("level", key);
+                    return  map;
```
Pablete1234, line 89:
```suggestion
                return obj.entrySet().stream().map(entry -> {
                    Map<String, Object> map = context.deserialize(entry.getValue(), Map.class);
                    map.put("level", entry.getKey());
                    return map;
```

Pablete1234 (commented):
Looks good, weird how this wasn't catched earlier, maybe the format changed not that long ago?

Left a bit of a cleanup as suggestion

m9w:
I not know once it was changed. I get it when staring making proof-of-concept for limitation bid count for my auction plugin.

Pablete1234 (approved):

PR #342, opened 2024-03-02
Diminish quest mediator

m9w:


src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestFacade.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.github.manolo8.darkbot.Main;
+import com.github.manolo8.darkbot.core.itf.Updatable;
+import com.github.manolo8.darkbot.core.objects.Gui;
+import eu.darkbot.api.API;
+import eu.darkbot.api.managers.QuestAPI;
+
+import static com.github.manolo8.darkbot.Main.API;
+
+public class DiminishQuestFacade extends Updatable implements API.Singleton {
```
Pablete1234, line 11:
```suggestion
public class DiminishQuestMediator extends Updatable implements API.Singleton {
```
We try to keep naming of mediator and proxy in our classes

src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
```diff
@@ -153,6 +154,8 @@ public GuiManager(Main main, PluginAPI pluginAPI, RepairManager repairManager) {
         this.assembly = register("assembly");
 
         register("ggBuilder", GateSpinnerGui.class);
+        register("refinement_count");
```
Pablete1234, line 157: does this have any relationship with this pr?

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestFacade.java
```diff
@@ -0,0 +1,56 @@
...
+        if (address == 0) state = State.NOT_EXIST;
+        else if (!"diminish_quests".equals(API.readString(address, 0x38))) {
+            address = 0;
+            state = State.NOT_EXIST;
+        }
+        else try {
+            gui = Main.INSTANCE.guiManager.getGui("diminish_quests");
+            state = State.values()[API.readInt(address, 0x50, 0x40) + 1];
+            diminishQuest.update(API.readAtom(address, 0x50, 0x60));
+        } catch (Exception ignored) {
```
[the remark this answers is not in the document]
m9w: I using this implementation ~250h and I catch behavior where fields have incorrect values or this mediator was excluded from mediator flashmap and address referenced to unexpected place when quest not active. This code covered this cases but now exist case where all fields say that quest ready to accept like accepting timer counting, gui window showing, it have correct position etc, but in-game it look like quest not present. I'm solving it via accepting attempts timer that lock stitching to temp module on 30 min.

This catch-all should fused sensitivity logic that can interrupt active module in case when we cannot read correctly diminish quest state.

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestFacade.java
```diff
@@ -0,0 +1,56 @@
...
+    
+    @Override
+    public void update() {
+        if (address == 0) state = State.NOT_EXIST;
+        else if (!"diminish_quests".equals(API.readString(address, 0x38))) {
+            address = 0;
+            state = State.NOT_EXIST;
+        }
+        else try {
+            gui = Main.INSTANCE.guiManager.getGui("diminish_quests");
```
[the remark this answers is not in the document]
m9w: As i remained it call circle dependency. I will check it again.

src/main/java/com/github/manolo8/darkbot/core/manager/GuiManager.java
m9w: I added this one-row change of registration GUI window is need for getting window  position for click accept updating generators with promerium. It not have relationship with diminish quest but it's useful in plugin code because provide required information via API.
I mean it allow exclude reflection code, and implement updating logic on not-patched bot.
image

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestFacade.java
m9w: Done

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestFacade.java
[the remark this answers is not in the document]
m9w: Added forwarding catched exception.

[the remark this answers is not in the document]
Punikekk: You can simply check index of the state instead of catching the exceptions.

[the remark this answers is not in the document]
m9w: Done

src/main/java/com/github/manolo8/darkbot/core/objects/gui/SizableGui.java
```diff
@@ -0,0 +1,13 @@
...
+import com.github.manolo8.darkbot.core.objects.Gui;
+
+public class SizableGui extends Gui {
+    public void update() {
+        if (address == 0) return;
+        super.update();
+        width = (int) Main.API.readDouble(address + 0x1F8);
+        height = (int) Main.API.readDouble(address + 0x200);
+    }
+}
```
Pablete1234, line 13: This SizableGui seems to be exactly what PopupGui is supposed to be, a generic superclass for guis that have the width and height in those addresses

Maybe the issue is PopupGui should not have the API.Singleton extension, that looks like an issue

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestMediator.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public DiminishQuestMediator(DiminishQuestGui diminishQuestGui) {
+        this.diminishQuestGui = diminishQuestGui;
+    }
+
+    @Override
+    public void update() {
+        if (address == 0) {
+            state = State.NOT_EXIST;
+        } else if (!"diminish_quests".equals(API.readString(address, 0x38))) {
```
Pablete1234, line 23: given the string caching that we do, and how otherwise string reading is an expensive operation, i'd say you can probably avoid doing this check at all.

Look into implementing separate update(addr) and update() methods, where the former invalidates some stuff and checks for this string, and this method checks if the address of the string it points to changes, if that is even required.

Is this an actual requirement or just a memory-bug fixing strategy?

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestMediator.java
```diff
@@ -0,0 +1,59 @@
...
+            address = 0;
+            state = State.NOT_EXIST;
+        } else {
+            state = State.NOT_EXIST;
+            int stateIndex = API.readInt(address, 0x50, 0x40) + 1;
+            if (stateIndex >= 0 && stateIndex < 5) {
+                state = State.values()[stateIndex];
+                diminishQuest.update(API.readAtom(address, 0x50, 0x60));
+            }
+        }
```
Pablete1234, line 33:
```suggestion
        if (!isValid()) {
            state = State.NOT_EXIST;
            return;
        }
        int stateIndex = API.readInt(address, 0x50, 0x40) + 1;
        if (stateIndex >= 0 && stateIndex < STATES.length) {
            state = STATES[stateIndex];
            diminishQuest.update(API.readAtom(address, 0x50, 0x60));
        } else {
            state = State.NOT_EXIST;
        }
```
i'd simplify the method in this logic, move validation check (addr != 0, and the other string check) into its own method, and make sure to check against the actual size of states instead of hard-coded `5`
also just add a `private static final State[] STATES = State.values()` to avoid the allocations every tick.

Also, what's the +1? is there an extra state 0 in-game we're ignoring here?

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestFacade.java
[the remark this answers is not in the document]
Pablete1234: seems like you figured this one out too

Pablete1234:
Code looks much better now, just a few more simplifications and it's good to merge

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestMediator.java
m9w: It's solution that allow be sure that flash object was not removed by GC.
It's issue of flashmap updating where key removing from collection, bot foget about this instance and it caused that the existing memory pointer (address) is not actual can be reused by another object.

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestMediator.java
m9w: Possible state indexes: 

- -1 when quest not active (timer left etc), 
- 0 when waiting accepting, 
- 2 when quest accepted and active. 
- 3 when quest done.

'+1' is mapping in-game index to enum index.

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestMediator.java
Pablete1234: that's something that the bot's dictionary for guis should already be handling, if the object stops being in the map it should be reset to address 0
i believe it was a bug that the older implementations for dictionary weren't doing it, but the new flashmap should be doing that just fine

src/main/java/com/github/manolo8/darkbot/core/objects/gui/SizableGui.java
m9w: I not used PopupGui because it using incorrect way for getting isVisible for this case.

src/main/java/com/github/manolo8/darkbot/core/objects/facades/DiminishQuestMediator.java
m9w: It's great. So I'll remove this condition.

m9w: Done

Pablete1234 (approved):

PR #347, opened 2024-03-10
feat: QuestGiverMediator

dm94:


Pablete1234:
This looks like a placeholder instead of a proper API. Please consider supplying an actual api that makes more sense as a whole

dm94:
> This looks like a placeholder instead of a proper API. Please consider supplying an actual api that makes more sense as a whole

In order to make progress while testing I made a lot of logic in my plugin and now I will move it here.

Pablete1234 (approved):

PR #353, opened 2024-03-19
pass language to browser

anjeshshrestha:


fabio1999ita:
this will works only for int account right?

dm94 (approved):

Pablete1234 (approved):

PR #354, opened 2024-03-19
feat: seasson pass mediator

dm94:
Linked to https://github.com/darkbot-reloaded/DarkBotAPI/pull/97


src/main/java/com/github/manolo8/darkbot/core/objects/facades/SeassonPassMediator.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public @Nullable List<? extends SeassonPassQuest> getSeassonQuests() {
+        if (!seassonQuestsUpdated && questDataAddr != 0) {
+            seassonQuestsUpdated = true;
+            seassonQuests.update(API.readAtom(questDataAddr + 0x70));
+        }
+
+        return this.seassonQuests;
+    }
```
Pablete1234, line 86: how big are these lists of quests? is it 100 quests or like 10? if it's a small list (like 5 or 10) there's no need to do this optimization of updating once required

dm94: From 5 to 20 each list. Then there is a list that is all there that I have not implemented so I could implement that list with this system and you are without this optimisation.

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SeassonPassMediator.java
```diff
@@ -0,0 +1,135 @@
...
+    @Getter
+    private boolean seassonPassAvailable = false;
+
+    private final FlashList<SeassonPassQuest> dailyQuests = FlashList.ofVector(SeassonPassQuest.class);
+    private final FlashList<SeassonPassQuest> weeklyQuests = FlashList.ofVector(SeassonPassQuest.class);
+    private final FlashList<SeassonPassQuest> seassonQuests = FlashList.ofVector(SeassonPassQuest.class);
+    private boolean allQuestsUpdated = false;
+    private final FlashList<SeassonPassQuest> allQuests = FlashList.ofVector(SeassonPassQuest.class);
+
+    private long questDataAddr = 0;
```
Pablete1234, line 36: prefer ordering fields such that final fields are first, then nonfinal fields

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SeassonPassMediator.java
```diff
@@ -0,0 +1,135 @@
...
+            if (address == 0) {
+                return;
+            }
+
+            this.maxProgress = API.readInt(address + 0x24);
+            this.currentProgress = API.readInt(address + 0x28);
+        }
+
+        public double getProgressPercentage() {
+            return this.currentProgress / this.maxProgress;
```
Pablete1234, line 104: If max progress wasn't ever read, you're going to divide by 0
Also you're doing integer division and returning it as a double, this will always be 0.

Consider using an IDE which has warnings for things like these like intellij
image


```suggestion
            return maxProgress == 0 ? 1d : (double) this.currentProgress / this.maxProgress;
```

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SeassonPassMediator.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    @Getter
+    @ToString
+    public class SeassonPassQuest extends Updatable implements SeassonPassAPI.SeassonPassQuest {
```
Pablete1234, line 110:
```suggestion
    public static class SeassonPassQuest extends Updatable implements SeassonPassAPI.SeassonPassQuest {
```

This inner class should be static, idea would've warned about this too:
image

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SeassonPassMediator.java
```diff
@@ -0,0 +1,135 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+public class SeassonPassMediator extends Updatable implements SeassonPassAPI {
+
+    @Getter
+    private CurrentLevelProgress currentLevelProgress = new CurrentLevelProgress();
+
+    @Getter
+    private CurrentLevelProgress currentSeassonLevelInfo = new CurrentLevelProgress();
```
Pablete1234, line 25: These two can be final, idea would've warned about this too:
image

dm94: done

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SeassonPassMediator.java
dm94: done

src/main/java/com/github/manolo8/darkbot/core/objects/facades/SeassonPassMediator.java
dm94: done

PR #357, opened 2024-05-09
feat: Added the option to use all the stats

dm94:


src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
```diff
@@ -35,10 +35,22 @@ public class StatTypeValue implements Value<Number>, Parser {
...
+
+        for (Stats.Bot statBot : Stats.Bot.values()) {
+            if (statBot.name().equalsIgnoreCase(key))
+                return statBot;
+        }
+
+        for (Stats.BootyKey statKey : Stats.BootyKey.values()) {
+            if (statKey.name().equalsIgnoreCase(key))
+                return statKey;
+        }
```
Pablete1234, line 52: instead of hard-coding the bot's stats, you should be using the stat API to grab any of the registered stats. The point here is that 3rd party plugins can register their own stats and with a custom tracker update the values, and they should be possible to use here too!

Consider having a syntax for namespaced keys, so you can have:
`experience`: namespace -> null (the bot's namespace), category -> any, name -> experience
`dmplugin:experience`: namespace -> dmplugin, category -> any, name -> experience
`dmplugin:general:experience`: namespace -> dmplugin, category -> general, name -> experience

To implement this matching logic you can probably expose in `StatsManager` and `StatsAPI` a:
```java
public Collection<? extends StatsAPI.Key> getStatKeys() {
  return Collections.unmodifiableSet(statistics.keySet());
}
```
and use that to search for the category for the namespace/key given, if none is found defer the operation to later as the stat could've not been registered yet.

Pablete1234 (approved):

Pablete1234 (changes requested):
Misclicked previous review as approved, this needs the changes mentioned

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
dm94: Implementation done, I'm not very proud of this implementation but I can't think of any other way right now

dm94: Implementation done, I'm not very proud of this implementation but I can't think of any other way right now.

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
```diff
@@ -35,11 +39,18 @@ public class StatTypeValue implements Value<Number>, Parser {
...
-        for (Stats.General stat : Stats.General.values()) {
-            if (stat.name().equalsIgnoreCase(key)) return stat;
-        }
-        return null;
+    private StatsAPI.Key getKeyFromString(String token) {
+        String[] tokenParts = token.split(":");
+
+        String statNamespace = tokenParts.length == 3 ? tokenParts[0] : null;
+        String statCategory = tokenParts.length >= 2 ? tokenParts[tokenParts.length - 2] : null;
+        String statKey = tokenParts[tokenParts.length - 1];
```
Pablete1234, line 47:
```suggestion
        String statNamespace = tokenParts.length >= 2 ? tokenParts[0] : null;
        String statCategory = tokenParts.length == 3 ? tokenParts[1] : null;
        String statKey = tokenParts[tokenParts.length - 1];
```

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
```diff
@@ -35,11 +39,18 @@ public class StatTypeValue implements Value<Number>, Parser {
         }
     }
 
-    private Stats.General getKeyFromString(String key) {
-        for (Stats.General stat : Stats.General.values()) {
-            if (stat.name().equalsIgnoreCase(key)) return stat;
-        }
-        return null;
+    private StatsAPI.Key getKeyFromString(String token) {
+        String[] tokenParts = token.split(":");
```
Pablete1234, line 43:
```suggestion
        String[] tokenParts = token.split(":", 3);
```

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
```diff
@@ -35,11 +39,18 @@ public class StatTypeValue implements Value<Number>, Parser {
...
+
+        String statNamespace = tokenParts.length == 3 ? tokenParts[0] : null;
+        String statCategory = tokenParts.length >= 2 ? tokenParts[tokenParts.length - 2] : null;
+        String statKey = tokenParts[tokenParts.length - 1];
+
+        return StatsManager.getStatKeys().stream()
+                .filter(s -> statNamespace == null || s.namespace().equals(statNamespace))
+                .filter(s -> statCategory == null || s.category().equals(statCategory))
+                .filter(s -> statKey == null || s.name().equals(statKey))
```
Pablete1234, line 52: nah, the key is not allowed to be null

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
```diff
@@ -35,11 +39,18 @@ public class StatTypeValue implements Value<Number>, Parser {
...
-        return null;
+    private StatsAPI.Key getKeyFromString(String token) {
+        String[] tokenParts = token.split(":");
+
+        String statNamespace = tokenParts.length == 3 ? tokenParts[0] : null;
+        String statCategory = tokenParts.length >= 2 ? tokenParts[tokenParts.length - 2] : null;
+        String statKey = tokenParts[tokenParts.length - 1];
+
+        return StatsManager.getStatKeys().stream()
+                .filter(s -> statNamespace == null || s.namespace().equals(statNamespace))
```
Pablete1234, line 50: a null namespace is just, the null namespace. that's the bot's one, no need to resolve it

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
```diff
@@ -35,11 +39,18 @@ public class StatTypeValue implements Value<Number>, Parser {
...
 
-    private Stats.General getKeyFromString(String key) {
-        for (Stats.General stat : Stats.General.values()) {
-            if (stat.name().equalsIgnoreCase(key)) return stat;
-        }
-        return null;
+    private StatsAPI.Key getKeyFromString(String token) {
+        String[] tokenParts = token.split(":");
+
+        String statNamespace = tokenParts.length == 3 ? tokenParts[0] : null;
```
Pablete1234, line 45: if this turns out to be an empty string, it needs to be treated as null, eg: `:general:uridium` would reference namespace = null, category = general, key = uridium

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
```diff
@@ -123,6 +128,14 @@ private void registerImpl(Key key, StatImpl stat) {
         statistics.put(StatKey.of(key), stat);
     }
 
+    public static Collection<? extends StatsAPI.Key> getStatKeys() {
+        return Collections.unmodifiableSet(getInstance().statistics.keySet());
+    }
+
+    public static StatsManager getInstance() {
```
Pablete1234, line 135: preferrably get this thru a reference to `Main.INSTANCE` instead of creating more static stuff

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/manager/StatsManager.java
dm94: Done

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
dm94: As it stands right now it is backwards compatible and as far as I can see it works correctly.

dm94:
I think the way it is now is the most correct way as it allows backwards compatibility to be maintained.

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
Pablete1234: In case of `:general:uridium` it will try to match against namespace=`""`, category=`"general"` and stat=`"uridium"`, which will find nothing in that "" namespace.
Just make it so if it's an empty string it's also treated as null

src/main/java/com/github/manolo8/darkbot/config/actions/values/StatTypeValue.java
Pablete1234: If you don't cap it at 3, you may have a string like:
`plugin:something:general:uridium`
which would be parsed as
namespace=plugin, category=general, name=uridium, leaving the "something" being completely ignored. This is faulty behavior, so you either limit it to 3 splits or you throw an error if there's more than 3, but it needs to be handled

PR #363, opened 2024-06-20
Fix broken offsets after game update

Punikekk:


cristophermr (approved):
When generating the jar the code does not work, the change only works in the IDE
image

edited

dm94 (approved):

anjeshshrestha (approved):

PR #365, opened 2024-06-20
fix: updated seasson pass mediator to be less resource consuming

dm94:


Pablete1234:
should be taken care of by #367

PR #366, opened 2024-06-20
refactor: updated astral gate proxy to improve performance

dm94:


Pablete1234:
Astral gate mem reading is negligible in terms of cpu usage, the optimization is not worth the effort of making sure nothing broke

PR #371, opened 2024-07-14
performance

mgsorigo:


src/main/java/eu/darkbot/api/utils/NativeAction.java
```diff
@@ -32,17 +29,16 @@ public enum Mouse {
         }
 
         public long of(int x, int y) {
-            return toActionLong(message, (short) 1, (short) x, (short) y, false);
+            return toActionLong(message, (short) 1, x, false);
```
dm94, line 32: Why do you remove the use of "y"?

mgsorigo: sorry my mistake i add:
 public long after(int x, int y) {
            return toActionLong(message, (short) 1, x, y, true);
        }

TeYroXOfficial:
Kinda don't understand what is the point of this PR. All you do was remove some comments and change tabs.. With what it will help according to "performance"?

PR #372, opened 2024-08-05
Add plugin config to ConfigAPI tree

m9w:


m9w:
+ https://github.com/darkbot-reloaded/DarkBotAPI/pull/100

Pablete1234:
What's the purpose of this PR?

The configs are considered different config roots, and it's  just a visual-side thing that they show up within the same UI tree. API-wise they're accessible via extensions api (to get all loaded features) plus config API where you can get config for a root other than the bot's main config tree.

m9w:
> What's the purpose of this PR?

Add access to plugin configs from ConfigAPI. I not see way to get root object of CUSTOM_CONFIGS without dependencies to main repo.

Pablete1234:
`FeatureInfo#getConfig` returns you the config root of a feature, after that you can use `ConfigApi#getChildren(root, "")` to find all the children programatically etc

You can get `FeatureInfo`s from the `ExtensionsAPI`, although from what i can see you can only get either by class or by id, so there's no programatic way to "get them all", you can still, without any dependency, access the configs of any given feature as long as you know the id tho.

Would be nice to add a way to get all loaded `FeatureInfo`s to the api to cover the missing gap, but that's about it

Pablete1234:
The "missing gap" of Being able to get the features for a plugin has been covered in https://github.com/darkbot-reloaded/DarkBotAPI/commit/9fc40fd0a717ab0b42dcc15d5d37766b5b6ca01d#diff-36255cf57bd7f91b529ec55b8a1b1b26363e472c26a5c29dc328f8a4df3a584cR66

I consider the searched functionality to be covered

PR #377, opened 2024-10-15
Change how locator npcs are handled & add API to select locator pick

Punikekk:


src/main/java/com/github/manolo8/darkbot/config/NpcInfo.java
```diff
@@ -153,6 +154,10 @@ public static String getId(Enum<?> flag) {
         return lastSearchedId = className.concat(name);
     }
 
+    public String getFuzzyName(String originalName) {
```
Pablete1234, line 157: Needs to be exposed in the API so that LocatorPick's getFuzzyName can be compared to this without relying on internals. Also consider just having name be a field in NpcInfo so that the key (originalName) doesn't need to be passed in as param.

src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
```diff
@@ -736,6 +736,21 @@ public void setModule(long gearsSprite) {
...
+
+        @Override
+        public String getName() {
+            return name;
+        }
+
+        @Override
+        public String getFuzzyName() {
+            return fuzzyName;
+        }
```
Pablete1234, line 748: For performance reasons, we may want to just make gear have a getNpcInfo, that gets mapped and updated by the bot (when updating gears, if name changes.

That way we avoid N^2 where each NpcInfo iterates thru all LocatorPick to find the matching one being ran every time, and in several places.

src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
```diff
@@ -767,5 +787,19 @@ public PetGear get() {
         public @Nullable Boolean enablePet() {
             return pet.isEnabled();
         }
+
+        @Override
+        public @Nullable LocatorPick getNpcLocatorPick(@UnmodifiableView Collection<? extends LocatorPick> picks) {
```
Pablete1234, line 792: Instead of providing a collection and expecting each impl to filter thru it, would it make more sense to provide this in a similar fashion to `compareTo` at the API level?

Maybe even have both methods, and have defaults (pseudo-code-kinda):
```java
default LocatorPick getNpcLocatorPick(Collection<LocatorPick> ...) {
  LocatorPick pick = picks.stream().max(Comparator.nullsLast().comparingInt(this::getPriority)).orElse(null);
  return pick != null && getPriority(pick) != null ?  pick : null;
}

default Integer getPriority(LocatorPick pick) {
  NpcInfo npc = pick.getNpcInfo();
  return npc != null && npc.kill() && npc.has(PET_LOCATOR) ? npc.priority (): null;
}
```

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
Punikekk: I thought about it but not always gear can find the corresponding NpcInfo, so have to think of a way to 'fix' it, when the bot see the given npc for the first time

Pablete1234: You can make the result nullable, at the end of the day the issue isn't going to be much different to what we have now, if no npc info matches the gear then it just can't really be selected

src/main/java/com/github/manolo8/darkbot/core/manager/PetManager.java
```diff
@@ -450,11 +441,8 @@ public Integer getGearOverride() {
     public @NotNull Collection<? extends NpcInfo> getLocatorNpcs() {
         if (locatorList.isEmpty()) return Collections.emptyList();
 
-        return main.config.LOOT.NPC_INFOS.entrySet()
-                .stream()
-                .map(entry -> new NpcPick(entry.getKey(), entry.getValue()))
-                .filter(p -> p.gear != null)
-                .map(p -> p.npc)
+        return locatorList.stream()
+                .map(gear -> gear.npcInfo)
```
Pablete1234, line 445: filter out nulls

Pablete1234 (approved):

PR #378, opened 2024-10-19
add returnee calendar manager to claim rewards

anjeshshrestha:
API Changes: https://github.com/darkbot-reloaded/DarkBotAPI/pull/105


src/main/java/com/github/manolo8/darkbot/core/objects/facades/ReturneeCalendarProxy.java
```diff
@@ -0,0 +1,42 @@
...
+import eu.darkbot.api.API;
+import eu.darkbot.api.managers.ReturneeAPI;
+import lombok.Getter;
+import lombok.ToString;
+
+@Getter
+public class ReturneeCalendarProxy extends Updatable implements API.Singleton {
+    private int daysClaimed = -1;
+    private boolean claimable = false;
+    private final FlashList<CalendarList> calendarList = FlashList.ofVector(CalendarList::new);
```
Pablete1234, line 14: List of lists? seems like the class probably needs to be renamed
```suggestion
    private final FlashList<CalendarItem> calendarList = FlashList.ofVector(CalendarItem::new);
```
something along those lines, and in the API too

PR #383, opened 2024-12-11
update change hangar call

anjeshshrestha:


src/main/java/com/github/manolo8/darkbot/backpage/LegacyHangarManager.java
```diff
@@ -60,22 +60,22 @@ public String getDataInventory(String params) {
 
     public boolean changeHangar(String hangarId) {
         if (lastHangarChange > System.currentTimeMillis() || !backpageManager.sidStatus().contains("OK")) return false;
-
         try {
-            String token = backpageManager.getConnection("indexInternal.es", Method.GET)
-                    .setRawParam("action", "internalDock")
-                    .consumeInputStream(backpageManager::getReloadToken);
+            JsonObject paramObj = new JsonObject();
```
Pablete1234, line 64: these changes are being done to LegacyHangarManager, what about the non-legacy version of this?

anjeshshrestha: this change was already implemented in the new hangarManager

https://github.com/darkbot-reloaded/DarkBot/blob/e98adf12c347b4ba065e4553bcfae75bfa1ca9e6/src/main/java/com/github/manolo8/darkbot/backpage/HangarManager.java#L127 

updating the old one to keep it working still

Pablete1234 (approved):

PR #389, opened 2025-03-14
add ship formation selector

anjeshshrestha:
Add formation selector for config, reduce key dependency to free up slotbar

https://github.com/user-attachments/assets/3ce9b8fa-c64f-4804-99e0-08191b2a3d61


src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -76,44 +78,71 @@ public class Config implements eu.darkbot.api.config.legacy.Config {
     public transient long changedAt;
 
     public @Option General GENERAL = new General();
+
     public static class General {
```
Pablete1234, line 82: avoid reformatting the whole class, it makes reading the diff impossible

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -196,13 +196,19 @@ public boolean attackMode() {
 
     public boolean attackMode(Npc target) {
         if (target == null) return attackMode();
-        Config.ShipConfig config = this.main.config.GENERAL.OFFENSIVE;
+        Config.ConfigFormation config = this.main.config.GENERAL.OFFENSIVE;
```
Pablete1234, line 199: This datatype change being needed points to a problem; ideally we do not want plugins to need to change how they do any of their work for this, the Config.ShipConfig type ideally should work (ie: just don't make it a new datatype, update the existing one)

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -196,13 +196,19 @@ public boolean attackMode() {
...
-                otherConfig ? ((config.CONFIG % 2) + 1) : config.CONFIG,
-                target.npcInfo.attackFormation != null ?
-                        target.npcInfo.attackFormation : config.FORMATION);
+
+        HeroAPI.Configuration newConfig = HeroAPI.Configuration.of(
+                otherConfig ? (config.getConfiguration().ordinal() % 2) + 1 : config.getConfiguration().ordinal()
+        );
+
+        SelectableItem.Formation formation = target.npcInfo.attackFormationNew != null ?
+                target.npcInfo.attackFormationNew :
```
Pablete1234, line 208: this probably needs accessors, so that it can support updating

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/NewShipModeEditor.java
```diff
@@ -0,0 +1,96 @@
...
+public class NewShipModeEditor extends JPanel implements OptionEditor<ShipMode> {
+    private HeroAPI.Configuration config;
+
+    private final List<ConfigButton> configButtons = Arrays.stream(HeroAPI.Configuration.values())
+            .filter(c -> c != HeroAPI.Configuration.UNKNOWN)
+            .map(config -> new ConfigButton(this, config)).collect(Collectors.toList());
+
+    private Formation formation;
+
+    private final JComboBox<Formation> comboBox = new JComboBox<>(Formation.values());
```
Pablete1234, line 33: Ideally instead of a dropdown with enum names, could use a set of icons for the formations and show the name as the tooltip

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/NewShipModeEditor.java
```diff
@@ -0,0 +1,96 @@
...
+import javax.swing.JPanel;
+import java.awt.Component;
+import java.awt.Dimension;
+import java.awt.FlowLayout;
+import java.awt.event.ItemEvent;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class NewShipModeEditor extends JPanel implements OptionEditor<ShipMode> {
```
Pablete1234, line 24: ideally this should be merged into the priorly existing ship config editor, there's no need for them to be separate

src/main/java/com/github/manolo8/darkbot/gui/utils/table/FormationNpcInfoEditor.java
```diff
@@ -0,0 +1,99 @@
...
+import javax.swing.JTable;
+import javax.swing.SwingUtilities;
+import javax.swing.ToolTipManager;
+import javax.swing.event.PopupMenuEvent;
+
+import java.awt.Rectangle;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+public class FormationNpcInfoEditor extends TableDelegateEditor<JLabel> {
```
Pablete1234, line 18: similar as pointed-out above, maybe we need icons and not text to show these, that would make them as compact as prior editor, where it was just a char plus the "key" icon

Pablete1234 (changes requested):

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/NewShipModeEditor.java
anjeshshrestha: that would be cleaner, maybe can be a follow up PR if someone can make the icons

src/main/java/com/github/manolo8/darkbot/config/Config.java
```diff
@@ -312,33 +312,61 @@ public static class ShipConfig implements LegacyShipMode {
...
             this.CONFIG = CONFIG;
             this.FORMATION = FORMATION;
         }
 
         @Override
         public HeroAPI.Configuration getConfiguration() {
-            return HeroAPI.Configuration.of(CONFIG);
+            if (CONFIG != -1) {
+                newConfiguration = HeroAPI.Configuration.of(CONFIG);
+                CONFIG = -1;
```
Pablete1234, line 334: for ideal backwards compatibility, you'd want to keep CONFIG and FORMATION updated too, just use the new ones if available preferrably

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -199,10 +199,16 @@ public boolean attackMode(Npc target) {
         Config.ShipConfig config = this.main.config.GENERAL.OFFENSIVE;
 
         boolean otherConfig = target.npcInfo.extra.has(NpcExtra.OPPOSITE_CONFIG);
-        return setMode(
-                otherConfig ? ((config.CONFIG % 2) + 1) : config.CONFIG,
-                target.npcInfo.attackFormation != null ?
-                        target.npcInfo.attackFormation : config.FORMATION);
+
+        HeroAPI.Configuration newConfig = HeroAPI.Configuration.of(
+                otherConfig ? (config.getConfiguration().ordinal() % 2) + 1 : config.getConfiguration().ordinal()
```
Pablete1234, line 204: configuration should probably just have a "other" method that returns the next one, would simplify this

src/main/java/com/github/manolo8/darkbot/core/manager/HeroManager.java
```diff
@@ -199,10 +199,16 @@ public boolean attackMode(Npc target) {
...
+        );
+
+        SelectableItem.Formation formation = target.npcInfo.attackFormationNew != null ?
+                target.npcInfo.attackFormationNew :
+                config.getFormation();
```
Pablete1234, line 209: Ideally this should spell out as:

`target.npcInfo.getFormation().orElse(config.getFormation());`
which would work if getFormation returns an Optional<Formation> within NpcInfo

src/main/java/com/github/manolo8/darkbot/config/Config.java
anjeshshrestha: i think the issue will be, since the standard formation is being set by default, it will likely use the "NEW" standard and not the one set by the key or update?

anjeshshrestha: wait not sure where to keep the old values updated from, i think from ShipModeEditor?

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ShipModeEditor.java
```diff
@@ -2,72 +2,102 @@
...
+        items[0] = "None";
+        System.arraycopy(SelectableItem.Formation.values(), 0, items, 1, SelectableItem.Formation.values().length);
+
+        comboBox = new JComboBox<>(items);
+        comboBox.setRenderer(new FormationRenderer());
+
+        // FIX: Add focus listener to show popup immediately
+        comboBox.addFocusListener(new FocusAdapter() {
+            @Override
+            public void focusGained(FocusEvent e) {
```
dm94, line 62: [AI] **Remove debug comment:** The // FIX: Add focus listener to show popup immediately comment should be removed before merging. If this is intentional behavior, consider a more descriptive comment explaining why.

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/ShipModeEditor.java
```diff
@@ -2,72 +2,102 @@
...
     private void setConfig(HeroAPI.Configuration config) {
-        this.config = config;
+        this.newConfig = config;
         configButtons.forEach(ConfigButton::repaint);
     }
 
+    private void setFormation(SelectableItem.Formation newFormation) {
+        this.newFormation = newFormation;
+        this.comboBox.setSelectedItem(newFormation == null ? "None" : newFormation);
+    }
```
dm94, line 96: [AI] **Focus behavior issue:** The comboBox triggers showPopup() on focus gained, which may interfere with keyboard navigation. Users expecting to tab through fields may find this disruptive.

src/main/java/com/github/manolo8/darkbot/config/utils/ItemUtils.java
```diff
@@ -34,4 +35,10 @@ public static Optional<Item> findAssociatedItem(@Nullable ItemCategory category,
                 .findAny();
     }
 
+    public static Character findAssociatedItem(@Nullable ItemCategory category, SelectableItem selectableItem) {
```
dm94, line 38: [AI] **Duplicate method name:** There's already a indAssociatedItem method above with similar functionality. This creates ambiguity. Consider renaming this one to something more specific like indKeyBindForItem or remove it if not used.

src/main/java/com/github/manolo8/darkbot/config/NpcInfo.java
```diff
@@ -109,7 +113,21 @@ public Optional<SelectableItem.Laser> getAmmo() {
 
     @Override
```
dm94, line 114: [AI] **Side effects in getFormation():** This method mutates ttackFormation and ttackFormationNew state during the getter call. This is unexpected behavior and could cause issues with config serialization or unexpected state changes.

dm94 (commented):
[AI] **Code Review Summary**

**Critical Issues:**
1. **State mutation in getter methods** - getConfiguration() and getFormation() in ShipConfig.java mutate instance state during getter calls. This is unexpected behavior and can cause side effects.

2. **Same issue in NpcInfo.java** - getFormation() also mutates ttackFormation and ttackFormationNew state.

**Minor Issues:**
3. **Debug comment** - Remove // FIX: Add focus listener... comment before merging in ShipModeEditor.java:62

4. **Focus behavior** - The comboBox shows popup on every focus gain, which may interfere with keyboard navigation.

5. **Duplicate method name** - ItemUtils.findAssociatedItem() method at line 38 has the same name as existing method above with similar functionality.

6. **File issues:**
   - x2.svg appears identical to mosquito.svg - verify if intentional
   - Remove .DS_Store file from commit

Please address these issues before merging.

src/main/java/com/github/manolo8/darkbot/gui/tree/editors/NewShipModeEditor.java
anjeshshrestha: added formation svg's

PR #396, opened 2025-04-13
auto close assembly

anjeshshrestha:
auto close assembly if not used

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/objects/gui/AssemblyGui.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/objects/gui/AssemblyGui.java
@@ -1,4 +1,22 @@
 package com.github.manolo8.darkbot.core.objects.gui;
 
-public class AssemblyGui {
+import com.github.manolo8.darkbot.core.objects.Gui;
+import eu.darkbot.api.API;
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/manolo8/darkbot/core/objects/gui/AssemblyGui.java
```diff
@@ -1,4 +1,22 @@
...
+import com.github.manolo8.darkbot.core.objects.Gui;
+import eu.darkbot.api.API;
+import eu.darkbot.api.managers.BotAPI;
+import eu.darkbot.util.Timer;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class AssemblyGui extends Gui implements API.Singleton {
+    private final BotAPI bot;
+    private final Timer guiUsed = Timer.getRandom(19_000, 1000);
```
Pablete1234, line 12: Is anmything at all updating this? feels like this pr is missing adding an actual API ontop of assembly that would refresh this timer

anjeshshrestha: added the manager to interact with the gui

Pablete1234:
Added in 6a4d0c05

PR #397, opened 2025-04-22
Ignore PET while jumping

do-gamer:
Future request: https://discord.com/channels/523159099870019584/564190622492262402/1360725629191852293

Name of feature: General -> PET:  Add checkbox "Ignore PET while jumping"

Description: If I use PET in "Guard mode", then every time when escaping and jumping, or when traveling to another map and jumping to the gate, the bot switching the PET to "Guard mode" . 
This causes an issue where the PET window can get stuck.


Pablete1234 (commented):
While i appreciate the pr, i believe this shouldn't be a setting, it should just be considered a bugfix. The bot going crazy with the pet while jumping is a bug and it should just automatically wait during jumps until its done loading in before taking any action

PR #419, opened 2025-08-13
Fix toggleVisibility crashes with updated Flash signatures

NeewJax:
- Replace broken toggleVisibility signature with working
- try-catch to prevent crashes when signatures become invalid


NeewJax:
I did run into some issues with things like enabling the pet / GG spinner :/ closed now

anjeshshrestha:
maybe I can help, but i dont understand the signature checks though 😔

PR #429, opened 2025-09-18
Feat/portals weight

dm94:


src/main/java/com/github/manolo8/darkbot/core/manager/StarManager.java
```diff
@@ -57,7 +57,7 @@ private StarManager() {
                 .addMap(21, "2-5").addPortal(2000, 11500, "4-4").addPortal(18500, 11500, "4-5").addPortal(2000, 2000, "2-6").addPortal(18500, 2000, "2-7").addPortal(10500, 6750, "Experiment Zone 2-2")
                 .addMap(22, "2-6").addPortal(2000, 11500, "2-5").addPortal(18500, 2000, "2-8")
                 .addMap(23, "2-7").addPortal(2000, 11500, "2-5").addPortal(18500, 2000, "2-8")
-                .addMap(24, "2-8").addPortal(2000, 11500, "2-6").addPortal(18500, 11500, "2-7").addPortal(11084, 11084, "2BL");
+                .addMap(24, "2-8").addPortal(2000, 11500, "2-6").addPortal(18500, 11500, "2-7").addPortal(10500, 12500, "2BL", 10.0);
```
Pablete1234: This just makes it so the void will avoid using BL portals even if they may be significantly shorter paths.
I don't understand the rationale behind this, unless its intended to be for accounts of a level lower than minimum requirement, but if that's the aim then it shouldn't be just 10 more weight, it should be straight up disabled until you have the minimum level, and should change based on your level.

Pablete1234 (changes requested):

PR #444, opened 2026-02-08
feat: Added more conditions info

dm94:


Pablete1234 (commented):
Avoid reformatting lines you haven't modified, makes reviewing and checking the git history later on harder

dm94:
> Avoid reformatting lines you haven't modified, makes reviewing and checking the git history later on harder

You're right, my editor does it automatically, and if I don't notice, formatting changes get uploaded

PR #447, opened 2026-02-13
Signature resolver to maintain compatibility with the Tanos API

do-gamer:
This fix is related to targeting the Relay in the LoW gate.

The signature `23(set target)(2626)1016221500` check called from the `MapManager`
https://github.com/darkbot-reloaded/DarkBot/blob/f9969bef4b6f26fb643a1f818ec506f2044b22d2/src/main/java/com/github/manolo8/darkbot/core/manager/MapManager.java#L434-L440

And this signature works fine for the Kekka Player API.
However, for the Tanos API, it always returns an error: "Invalid flash method signature!"
The correct signature for Tanos API is `23(target)(2626)1016221500`

I'm not sure whether this is the correct place or method to specify a different signature for Tanos API only.
Please review and let me know.

### DIFF
--- a/src/main/java/com/github/manolo8/darkbot/core/api/adapters/TanosAdapter.java
+++ b/src/main/java/com/github/manolo8/darkbot/core/api/adapters/TanosAdapter.java
@@ -15,6 +15,7 @@ import eu.darkbot.api.game.other.Locatable;
 import eu.darkbot.api.managers.OreAPI;
 
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


Pablete1234:
If tanos is producing a different signature than kekka, that is a bug on tanos' side

do-gamer:
@Pablete1234 Thank you for pointing me in correct way.
I have fixed the `libdo_lib.so` to resolve correct getters/setters names, just like the Kekka Player API.
```
Signature: 23(set target)(2626)1016221500 == 23(set target)(2626)1016221500
```

